PR #12, opened 2023-08-23
write counters into JFR

richardstartin:
This writes the counters out into the JFR file, for debugging issues remotely. Unfortunately it has to make some compromises on counters updated:
* transient dictionaries get whatever count they're at when the point where the count is serialized - it might be worth adding an attribute to these to hide them if it gets confusing
* the class map gets the count from the previous reporting

I decided it wasn't worth the risk or cost (my engineering time, buffering at runtime) reworking the writing of the constant pool to allow precise reporting of these counts, since the former is pointless and the latter is convergent.

This does not compile the counters into release builds yet - this will be a follow up.


ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -322,6 +323,14 @@ off_t Recording::finishChunk(bool end_recording) {
         writeRecordingInfo(_buf);
     }
 
+    // this will not include counters for any counters updated during writing the constant pool
```
jbachorik, line 326: I think there is a typo here

```suggestion
    // this will not include counts for any counters updated during writing the constant pool
```

jbachorik (approved):

PR #25, opened 2023-09-18
do not unwind stack if last Java stack-pointer is NULL

richardstartin:
**What does this PR do?**:
The last Java SP is set to NULL is used to indicate that the stack is currently unparseable, so we don't attempt to get a stack trace from AsyncGetCallTrace.

**Motivation**:

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:

### DIFF
--- a/ddprof-lib/src/main/cpp/profiler.cpp
+++ b/ddprof-lib/src/main/cpp/profiler.cpp
@@ -355,6 +355,12 @@ int Profiler::getJavaTraceAsync(void* ucontext, ASGCT_CallFrame* frames, int max
         return 0;
     }
 
+    if (VM::isHotspot() && vm_thread->lastJavaSP() == 0 && vm_thread->lastJavaPC() != 0) {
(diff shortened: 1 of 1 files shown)


ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -355,6 +355,12 @@ int Profiler::getJavaTraceAsync(void* ucontext, ASGCT_CallFrame* frames, int max
         return 0;
     }
 
+    if (VM::isHotspot() && vm_thread->lastJavaSP() == 0 && vm_thread->lastJavaPC() != 0) {
```
jbachorik, line 358: This is just a nit - would it make sense to do `vm_thread->lastJavaSP() == NULL` since the comment is talking about NULL value (I know they are the same thing, that's why this is just a nit)

jbachorik (approved):

PR #41, opened 2023-10-31
Improve the robustness of jmethodid checks

jbachorik:
**What does this PR do?**:
It is using a public function `Method::is_valid_method(method)` to avoid trying to resolve a jmethodid pointing to an invalid method.

**Motivation**:
Reduce (and possibly eliminate) the likelihood of JVM crashes due to jmethodid pointing to a partially cleaned (invalid) `Method` object.

**Additional notes**:
I had to modify the code for retrieving the heap usage as JDK 21 with generational ZGC is very unhappy about using the internal function which is working well for pre-JDK21 ZGC implementations.

**How to test the change?**:
It is not easy to trigger the condition with a partially cleaned method and I used an internal project's test suite which is using an extraordinary amount of mocked classes triggering a lot of class unloads for lambda forms.

**For Datadog employees**:
- [x] JIRA: [PROF-8526](https://datadoghq.atlassian.net/browse/PROF-8526)

[PROF-8526]: https://datadoghq.atlassian.net/browse/PROF-8526?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/vmStructs.h
+++ b/ddprof-lib/src/main/cpp/vmStructs.h
@@ -22,6 +22,7 @@
 #include <stdint.h>
 #include <string.h>
 #include "codeCache.h"
+#include "jniHelper.h"
 #include "jvmHeap.h"
 #include "vmEntry.h"
 #include "threadState.h"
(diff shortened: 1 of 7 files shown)


ddprof-lib/src/main/cpp/vmStructs.h
```diff
@@ -446,6 +449,30 @@ class HeapUsage : VMStructs {
...
+        }
+
+        if (!(_max_memory = env->GetMethodID(_rt, "maxMemory", "()J"))) {
+            env->ExceptionDescribe();
+            return -1;
+        }
+
+        jobject rt = (jobject)env->CallStaticObjectMethod(_rt, _get_rt);
+        return (jlong)env->CallLongMethod(rt, _max_memory);
+    }
```
richardstartin, line 474: if this calls into `jniExceptionCheck` it will also clear the exception status. Could you run the tests with `-Xjni:check` to be sure the changes don't introduce new warnings?

richardstartin (approved):

PR #42, opened 2023-11-08
use safeaccess to avoid dereferencing unreadable pointers in program headers

richardstartin:
**What does this PR do?**:

We have a crash report with the following stack:

```
Stack: [0x00007f45072cc000,0x00007f45073cd000],  sp=0x00007f45073cafb8,  free space=1019k
Native frames: (J=compiled Java code, j=interpreted, Vv=VM code, C=native code)
C  [libjavaProfiler6901221762100336125.so+0x149f8]  ElfParser::parseDynamicSection()+0x18
C  [libjavaProfiler6901221762100336125.so+0x15fb9]  Symbols::parseLibraries(CodeCacheArray*, bool)+0x509
C  [libjavaProfiler6901221762100336125.so+0x4508d]  Profiler::start(Arguments&, bool)+0x3cd
C  [libjavaProfiler6901221762100336125.so+0x467d5]  Profiler::runInternal(Arguments&, std::ostream&)+0x375
C  [libjavaProfiler6901221762100336125.so+0xf5fe]  Java_com_datadoghq_profiler_JavaProfiler_execute0+0x32e
```

and siginfo: 

```
siginfo: si_signo: 11 (SIGSEGV), si_code: 2 (SEGV_ACCERR), si_addr: 0x00007f4505cf1000
```

and the following memory mappings:

```
7f4505a31000-7f4505cb9000 r-xp 00000000 fd:00 16919904                   /opt/rh/jbcs-httpd24/root/usr/lib64/libcrypto.so.1.1.1k
7f4505cb9000-7f4505eb9000 ---p 00288000 fd:00 16919904                   /opt/rh/jbcs-httpd24/root/usr/lib64/libcrypto.so.1.1.1k
```

The second memory mapping is the address range containing 0x00007f4505cf1000 so we must have attempted to read from the address range 7f4505cb9000-7f4505eb9000. However, this mapping should have been filtered out, because it is not readable. Right now, I don't know any more than this but this may be because RHEL7 is built with RELRO support.

As for where the sigsegv was raised, from objtool I have this for `parseDynamicSection`: 
```
00000000000148a0 <_ZN9ElfParser19parseDynamicSectionEv>:
   148a0:	4c 8b 57 18          	mov    0x18(%rdi),%r10
   148a4:	41 0f b7 72 38       	movzwl 0x38(%r10),%esi
   148a9:	4c 89 d0             	mov    %r10,%rax
   148ac:	49 03 42 20          	add    0x20(%r10),%rax
   148b0:	85 f6                	test   %esi,%esi
   148b2:	0f 84 56 01 00 00    	je     14a0e <_ZN9ElfParser19parseDynamicSectionEv+0x16e>
   148b8:	83 38 02             	cmpl   $0x2,(%rax)  <------------- offset +0x18 inlined 
   148bb:	41 0f b7 4a 36       	movzwl 0x36(%r10),%ecx
   148c0:	74 24                	je     148e6 <_ZN9ElfParser19parseDynamicSectionEv+0x46>
   148c2:	49 89 c8             	mov    %rcx,%r8
   148c5:	48 01 c8             	add    %rcx,%rax
   148c8:	31 d2                	xor    %edx,%edx
```

Which appears to be code inlined from the call to `findProgramHeader(PT_DYNAMIC)`. I think the sigsegv is raised when dereferencing `pheader` to check the type, so the validity of this pointer is now checked before checking the type.

**Motivation**:

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:

### DIFF
--- a/ddprof-lib/src/main/cpp/symbols_linux.cpp
+++ b/ddprof-lib/src/main/cpp/symbols_linux.cpp
@@ -31,6 +31,7 @@
 #include "symbols.h"
 #include "dwarf.h"
 #include "log.h"
+#include "safeAccess.h"
 
 
 class SymbolDesc {
@@ -212,9 +213,16 @@ ElfProgramHeader* ElfParser::findProgramHeader(uint32_t type) {
     const char* pheaders = (const char*)_header + _header->e_phoff;
 
     for (int i = 0; i < _header->e_phnum; i++) {
-        ElfProgramHeader* pheader = (ElfProgramHeader*)(pheaders + i * _header->e_phentsize);
-        if (pheader->p_type == type) {
-            return pheader;
+        const char* unvalidated_pheader = pheaders + i * _header->e_phentsize;
+        // check we can load the pointer
+        void* checked = SafeAccess::load((void**) unvalidated_pheader);
+        if (checked == NULL) {
+            return NULL;
+        } else {
+            ElfProgramHeader* pheader = (ElfProgramHeader*) unvalidated_pheader;
+            if (pheader->p_type == type) {
+                return pheader;
+            }
         }
     }
 


jbachorik (approved):

sanchda (commented):
Historical review, but just wanted to say that this looks OK.  I have every reason to believe that checking the program header here would be sufficient for the rest of the mappings, based on the presentation of the defect.

I'm not 100% sure how SafeAccess works, but I don't think I need to.  I can see from its implementation that it provides machinery for skipping over faulting instructions in an architecture-aware way, which is good enough for a speculative check.

PR #53, opened 2023-12-13
Sanchda/rust demangling

sanchda:
**What does this PR do?**:
Adds legacy Rust demangling to native code

**Motivation**:
I saw a flamegraph with tokio in the native parts and thought this might be useful.

**Additional Notes**:
Rust demangling isn't free, but I haven't really spent much time trying to speed it up.  This was taken from a similar implementation in the native profiler, but adapted to C++11

**How to test the change?**:
I added tests.  I believe this is fairly comprehensive, as it pulls from both llvm and libiberty.

**For Datadog employees**:
- [X] If this PR touches code that signs or publishes builds or packages, or handles

### DIFF
--- a/ddprof-lib/src/main/cpp/rustDemangler.cpp
+++ b/ddprof-lib/src/main/cpp/rustDemangler.cpp
@@ -1,0 +1,165 @@
+/*
+ * Copyright 2021, 2023 Datadog, Inc
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+#include "rustDemangler.h"
+
+#include <array>
+
+using namespace RustDemangler;
+
+// With some exceptions we don't handle here, v0 Rust symbols can end in a
+// prefix followed by a 16-hexdigit hash, which must be removed
+const std::string hash_pre = "::h";
+const std::string hash_eg = "0123456789abcdef";
+
+const std::array<std::pair<std::string, std::string>, 9> patterns = {{
+    {"..", "::"},
+    {"$C$", ","},
+    {"$BP$", "*"},
+    {"$GT$", ">"},
+    {"$LT$", "<"},
+    {"$LP$", "("},
+    {"$RP$", ")"},
+    {"$RF$", "&"},
+    {"$SP$", "@"},
+}};
+
+inline bool is_hexdig(const char c) {
+  return (c >= 'a' && c <= 'f') || (c >= '0' && c <= '9');
+}
+
+// Simple conversion from hex digit to integer
+inline int hex_to_int(char dig) {
+  constexpr int k_a_hex_value = 0xa;
+  if (dig >= '0' && dig <= '9') {
+    return dig - '0';
+  }
+  if (dig >= 'a' && dig <= 'f') {
+    return dig - 'a' + k_a_hex_value;
+  }
+  if (dig >= 'A' && dig <= 'F') {
+    return dig - 'A' + k_a_hex_value;
+  }
+  return -1;
+}
+
+// Minimal check that a string can end, and does end, in a hashlike substring
+inline bool has_hash(const std::string &str) {
+  // If the size can't conform, then the string is invalid
+  if (str.size() <= hash_pre.size() + hash_eg.size()) {
+    return false;
+  }
+
+  // Check that the string contains the hash prefix in the right position
+  if (str.compare(str.size() - hash_eg.size() - hash_pre.size(),
+                  hash_pre.size(), hash_pre)) {
+    return false;
+  }
+
+  // Check that the string ends in lowercase hex digits
+  for (size_t i = str.size() - hash_eg.size(); i < str.size(); ++i) {
+    if (!is_hexdig(str[i])) {
+      return false;
+    }
+  }
+  return true;
+}
+
+bool is_probably_rust_legacy(const std::string &str) {
+  // Is the string too short to have a hash part in thefirst place?
+  if (!has_hash(str)) {
+    return false;
+  }
+
+  // Throw out `$$` and `$????$`, but not in-between
+  const char *ptr = str.data();
+  const char *end = ptr + str.size() - hash_pre.size() - hash_eg.size();
+  for (; ptr <= end; ++ptr) {
+    if (*ptr == '$') {
+      if (ptr[1] == '$') {
+        return false;
+      }
+      return ptr[2] == '$' || ptr[3] == '$' || ptr[4] == '$';
+    }
+    if (*ptr == '.') {
+      return '.' != ptr[1] ||
+          '.' != ptr[2]; // '.' and '..' are fine, '...' is not
+    }
+  }
+  return true;
+}
+
+// Demangles a Rust string by building a copy piece-by-piece
+std::string demangle(const std::string &str) {
+  std::string ret;
+  ret.reserve(str.size() - hash_eg.size() - hash_pre.size());
+
+  size_t i = 0;
+
+  // Special-case for repairing C++ demangling defect for Rust
+  if (str[0] == '_' && str[1] == '$') {
+    ++i;
+  }
+
+  for (; i < str.size() - hash_pre.size() - hash_eg.size(); ++i) {
+
+    // Fast sieve for pattern-matching, since we know first chars
+    if (str[i] == '.' || str[i] == '$') {
+      bool replaced = false;
+
+      // Try to replace one of the patterns
+      for (const auto &pair : patterns) {
+        const std::string &pattern = pair.first;
+        const std::string &replacement = pair.second;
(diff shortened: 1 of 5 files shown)


ddprof-lib/src/main/cpp/rustDemangler.h
```diff
@@ -0,0 +1,25 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+#include <string>
+
+namespace RustDemangler {
```
jbachorik, line 20: Oh, namespace :)
First time I see it here. It's been a long time, I almost forgot you could do this in C++ 🤣

ddprof-lib/src/main/cpp/rustDemangler.cpp
```diff
@@ -0,0 +1,166 @@
...
+    {"$GT$", ">"},
+    {"$LT$", "<"},
+    {"$LP$", "("},
+    {"$RP$", ")"},
+    {"$RF$", "&"},
+    {"$SP$", "@"},
+}};
+
+inline bool is_hexdig(const char c) {
+  return (c >= 'a' && c <= 'f') || (c >= '0' && c <= '9');
```
jbachorik, line 42: Should this also accept `A`-`F` ? (`hex_to_int` seems to be ok with capital letters ...)

ddprof-lib/src/main/cpp/rustDemangler.cpp
```diff
@@ -0,0 +1,166 @@
...
+  if (dig >= 'A' && dig <= 'F') {
+    return dig - 'A' + k_a_hex_value;
+  }
+  return -1;
+}
+
+// Minimal check that a string can end, and does end, in a hashlike substring
+inline bool has_hash(const std::string &str) {
+  // If the size can't conform, then the string is invalid
+  if (str.size() <= hash_pre.size() + hash_eg.size()) {
```
jbachorik, line 63: Will the calls to `size()` be somehow constant folded when the string is `const`? 
I am genuinely curious here - I guess it should be a pretty simple optimization to be done by compiler, but I have no clue.

ddprof-lib/src/main/cpp/rustDemangler.cpp
jbachorik: Ok. Sratch it :)
I found the usage and it says in the comment that this should be looking for lower case letters only ...

ddprof-lib/src/main/cpp/rustDemangler.cpp
```diff
@@ -0,0 +1,166 @@
...
+  if (!has_hash(str)) {
+    return false;
+  }
+
+  // Throw out `$$` and `$????$`, but not in-between
+  const char *ptr = str.data();
+  const char *end = ptr + str.size() - hash_pre.size() - hash_eg.size();
+  for (; ptr <= end; ++ptr) {
+    if (*ptr == '$') {
+      if (ptr[1] == '$') {
```
jbachorik, line 93: I guess this is safe because the string has the tailing hash and there never will be `$` as the last char here.

ddprof-lib/src/main/cpp/rustDemangler.cpp
```diff
@@ -0,0 +1,166 @@
...
+        return false;
+      }
+      return ptr[2] == '$' || ptr[3] == '$' || ptr[4] == '$';
```
jbachorik, line 96: Does this mean that there will be at most 5 characters in the string?

ddprof-lib/src/main/cpp/rustDemangler.cpp
```diff
@@ -0,0 +1,166 @@
...
+        }
+      }
+
+      // If we failed to replace, try a few failovers.  Notably, we recognize
+      // that Rust may insert Unicode code points in the function name (other
+      // implementations treat many individual points as patterns to search on)
+      if (!replaced && str[i] == '.') {
+        // Special-case for '.'
+        ret += '-';
+      } else if (!replaced && !str.compare(i, 2, "$u") && str[i + 4] == '$') {
```
jbachorik, line 142: I guess the Unicode code point is represented something like `abcd$` ?
What would be the case of substring starting with`$u` ? Is is it a special case?

ddprof-lib/src/main/cpp/rustDemangler.cpp
```diff
@@ -0,0 +1,166 @@
...
+        const int hexa_base = 16;
+        const int hi = hex_to_int(str[i + 2]);
+        const int lo = hex_to_int(str[i + 3]);
+        if (hi != -1 && lo != -1) {
+          ret += static_cast<unsigned char>(lo + hexa_base * hi);
+          i += k_nb_read_chars - 1; // - 1 because iterator inc
+        } else {
+          // We didn't have valid unicode values, but we should still skip
+          // the $u??$ sequence
+          ret += str.substr(i, k_nb_read_chars);
```
jbachorik, line 153: Is this skipping the `$u??$` sequence or copying it to the ret string?

jbachorik (commented):
Hi, I have a bunch of noob questions. Please, be gentle :)

ddprof-lib/src/main/cpp/rustDemangler.h
sanchda: Heh, mostly here I was afraid of conflicting with something else in the project (I'm not familiar with the codebase).  Happy to remove if it would align with the design of the repo.  :pray:

ddprof-lib/src/main/cpp/rustDemangler.cpp
sanchda: Ah, right--it's worth thinking about this a bit.  I adapted this code from a system which was already using `std::string` everywhere, so there might be some unexpected surprises.

Roughly speaking,
* `str.size()` is a struct getter against the string's `size` parameter, which was computed at time of creation
* `hash_pre.size()` and `hash_eg.size()` are the same, except the scan was a one-time thing.  In later C++ standards, we could have used `constexpr std::string_view` to move even that scan to compile-time.

In this case, the size of the string is computed at time of creation.  Ultimately, `str` is created from a C string, and I create it the lazy way (line 89 of flightRecorder.cpp), so there's an implicit `strlen()` that happens (as well as a copy).  Constructing the string with the length as a parameter avoids the `strlen()`, but the `memcpy()` is still linear (albeit "fast", probably).

In later C++ standards, there'a `std::string_view` which can be used to create a non-owning "view" into a string (be it a C++ std::string or a C-style char * equivalent).  This is one way of avoiding any kinds of linear penalties, but it isn't compatible with the project (we'd need to build in C++17 mode, I think).



Anyway, my point is that your observation is indeed true for the consts, but we should be wary of the fact that `str` did require a `strlen()` and a `memcpy()` in order to be created.

ddprof-lib/src/main/cpp/rustDemangler.cpp
sanchda: Indeed, and also because `end` excludes the hash part, so we're guaranteed to have several characters of "string" left after any `ptr`.

ddprof-lib/src/main/cpp/rustDemangler.cpp
sanchda: Sorry, I couldn't think of a way of implementing this section cleanly without resorting to patterns/lookups that would have bloated the code even more.

Overall, this function is trying to check whether a given string _might_ be rust-legacy-mangled.  For that to be the case, it needs to end in a "hash" (checked separately above), and if it has a `$`, then the first observation thereof has to be in a valid pattern (`$?$` or `$??$` or `$???$`, but not `$$` or `$` (alone) or `$????$` etc).  So rather than there just being 5 characters, this is actually a fast sieve for checking that if there is a `$` then the first instance thereof is within a valid pattern.

It could be the case that the string is something like `hello$GT$wo$$::h0123456789abcdef`, which would pass as _probably_ rust legacy (even though it isn't) and get partially "demangled."  This implementation uses the sieve as a good-enough heuristic, since it'd be reeeeaally weird otherwise.

sanchda: Just to be clear, we could absolutely check that all patterns are valid.  In fact, we'd probably just drop the check and then implement the replacement in a way that it returns error if the conversion hit an invalid parse.

The problem is it'd just be more expensive.

ddprof-lib/src/main/cpp/rustDemangler.cpp
sanchda: Good question.  It "skips" the `$u???$` pattern in the sense that it doesn't perform additional processing on it, but it actually inserts the unprocessed `$u???$` sequence into the destination string.  In other words, tokens are only consumed when they are successfully matched, otherwise they are inserted literally into the result.

ddprof-lib/src/main/cpp/rustDemangler.h
jbachorik: I don't think is necessary to remove it - I assume it is a purely syntactic construct and won't result in linking more of libstdc++ to the final library ...

ddprof-lib/src/main/cpp/rustDemangler.cpp
jbachorik: Can you extend the comment with 'inserted, not processed' part? Thanks!

jbachorik (approved):

ddprof-lib/src/main/cpp/rustDemangler.cpp
sanchda: Fixed in 645fe69

sanchda:
Looks like none of the code quality nags were related to this PR.  Merged

PR #69, opened 2024-02-09
Fix the partial kernel/userspace stackwalking

jbachorik:
**What does this PR do?**:
It fixes the mixed kernel/userspace stackwalking

**Motivation**:
The current logic seems flawed when using perf_event_open scheduler and it fails to get the non-kernel part of the stacktrace.
The fix is trivial and it does fix the issue.

**How to test the change?**:
One must have a system with perf_event_open available. Collected stacktraces should not contain a significant number of `UNKNOWN_PACKAGE` frames.

**For Datadog employees**:
- [x] JIRA ticket: [PROF-9142] 

[PROF-9142]: https://datadoghq.atlassian.net/browse/PROF-9142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


richardstartin (approved):
Good catch, let's upstream this one

PR #81, opened 2024-03-25
Add startTime to datadog.ProfilerCounter event

manuel-alvarez-alvarez:
**What does this PR do?**:
Fixes the generation of JFR dumps so they can be read by the `jfr` tool by adding `startTime` to `datadog.ProfilerCounter` event

**Motivation**:
The `jfr` tool could not open the dumps generated by the profiler, failing with:

```
java.lang.ClassCastException: java.lang.String cannot be cast to java.lang.Long
        at jdk.jfr.consumer.EventParser.parse(EventParser.java:62)
```

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:

### DIFF
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -1163,6 +1163,7 @@ void Recording::writeCounters(Buffer* buf) {
         for (int i = 0; i < names.size(); i++) {
             int start = buf->skip(1);
             buf->putVar64(T_DATADOG_COUNTER);
+            buf->putVar64(_start_ticks);
             buf->putUtf8(names[i]);
             buf->putVar64(counters[Counters::address(i)]);
             writeEventSizePrefix(buf, start);
--- a/ddprof-lib/src/main/cpp/jfrMetadata.cpp
+++ b/ddprof-lib/src/main/cpp/jfrMetadata.cpp
@@ -254,6 +254,7 @@ void JfrMetadata::initialize(const std::vector<std::string>& contextAttributes) 
 
             << (type("datadog.ProfilerCounter", T_DATADOG_COUNTER, "Datadog Profiler Internal Counter")
                 << category("Datadog")
+                << field("startTime", T_LONG, "Start Time", F_TIME_TICKS)
                 << field("name", T_COUNTER_NAME, "Name")
                 << field("count", T_LONG, "Count"))
 


ddprof-lib/src/main/cpp/jfrMetadata.cpp
```diff
@@ -254,6 +254,7 @@ void JfrMetadata::initialize(const std::vector<std::string>& contextAttributes)
 
             << (type("datadog.ProfilerCounter", T_DATADOG_COUNTER, "Datadog Profiler Internal Counter")
                 << category("Datadog")
+                << field("startTime", T_LONG, "Start Time", F_TIME_TICKS)
```
manuel-alvarez-alvarez, line 257: It's an instance of `jdk.jfr.Event` so the `startTime` is required

jbachorik: The type is described only by its metadata, so, technically, it is not really required, as long as the parser does not make any assumptions and just follows the metadata.
Unfortunately, the built-in JDK JFR parser seems to make the assumption that events always have start time.

Since this event is emitted once per chunk, the cost of adding the field is almost nothing anyway.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -907,8 +907,12 @@ void Recording::writeCpool(Buffer* buf) {
     buf->put8(0);
     buf->put8(1);
     // constant pool count - bump each time a new pool is added
-    buf->put8(12);
-
+    char cpCount = 12;
+    Dictionary* stringLabelMap = Profiler::instance()->stringLabelMap();
+    if (stringLabelMap->empty()) { cpCount--; }
```
jbachorik, line 912: Sorry for the nitpicks, but could you change the formatting to have new line after `{` ?
(also in other places)
```suggestion
    if (stringLabelMap->empty()) {
        cpCount--;
    }
```

jbachorik (dismissed):
Please, update the formatting before merge

richardstartin (approved):

richardstartin:
@manuel-alvarez-alvarez we need to get to the bottom of why the tests fail on class metadata with this change, which might make this change more complex than it first seems. We've never aimed for compatibility with the JFR tool, only with JMC, but if you need this from us we can ensure support (and testing) in a reasonably short time frame. If you have spare cycles to debug this now, don't be discouraged.

ddprof-lib/src/main/cpp/flightRecorder.cpp
manuel-alvarez-alvarez: Yep, done!, should have done it like this in the first place :wink:

ddprof-lib/src/main/cpp/jfrMetadata.cpp
manuel-alvarez-alvarez: That is true (forgot that JFR is self descriptive via the metadata)

manuel-alvarez-alvarez:
> @manuel-alvarez-alvarez we need to get to the bottom of why the tests fail on class metadata with this change, which might make this change more complex than it first seems. We've never aimed for compatibility with the JFR tool, only with JMC, but if you need this from us we can ensure support (and testing) in a reasonably short time frame. If you have spare cycles to debug this now, don't be discouraged.

There's no real business need from having support for JFR (just me trying to open the dumps from IntelliJ). I took the opportunity as a weekend lecture to learn about the format of the dumps and profiling in general :smile:. Don't waste your time on this, if I find some spare time I'll take it for another ride and see why the tests are failing, but if nothing comes from it we can close the PR.

richardstartin:
@jbachorik I'm ok with this one when you are. @manuel-alvarez-alvarez thanks for doing this!

PR #89, opened 2024-04-29
drop bad queue time events

richardstartin:
**What does this PR do?**:
Queue time events which are committed during a profiler dump may end up with null scheduler and task attributes, if they are committed just as the class map is being cleared, because the lock can't be acquired when looking up the class name. These events aren't very useful so we'll drop them for now.

**Motivation**:

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:


jbachorik (approved):
Do we want to track the number of dropped events as a counter or we mostly don't care?

richardstartin:
Dropping an event isn't too significant except at the trace level - it doesn't change sampling statistics, the events are latency thresholded and tend to occur in batches, so if we drop some during clearing the class map, we'll usually get some just before or after. We can revisit this if we want to change the way buffering works during dumping the recording.

PR #93, opened 2024-05-09
Introduce 'light' liveheap tracking for GC generations

jbachorik:
**What does this PR do?**:
This change introduces a lightweight mode of the liveness tracker which is not collecting stacktraces and is tracking only the allocated type and the instance age.

**Motivation**:
For the Java memory leak page it is beneficial to be able to use the GC generations metric without the threat of crashing the app due to JVMTI jmethodID issue when symbolizing the stacktraces.
This mode is not enabled by default and needs to be turned on by supplying `generations=true` argument to the agent.

**How to test the change?**:
A new sanity test is added which is making sure setting the profiler argument `generations=true` will indeed turn on the liveness tracker.

**For Datadog employees**:
- [x] JIRA: [PROF-9743] 

[PROF-9743]: https://datadoghq.atlassian.net/browse/PROF-9743?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


richardstartin (approved):

ddprof-lib/src/main/cpp/objectSampler.cpp
```diff
@@ -64,25 +61,33 @@ void ObjectSampler::recordAllocation(jvmtiEnv* jvmti, JNIEnv* jni, jthread threa
...
     if (_record_allocations || _record_liveness) {
-        std::set<jclass> classes;
-        jclass method_class;
-        for (int i = 0; i < frames_size; i++) {
-            if (jvmti->GetMethodDeclaringClass(frames[i].method, &method_class) == 0) {
-                classes.insert(method_class);
+        event._size = size;
+        event._weight =  (float)((size == 0 || _interval == 0) ? 1 : 1 / (1 - exp(-size / (double)_interval)));
+
+        frames = new jvmtiFrameInfo[_max_stack_depth];
```
richardstartin, line 72: Can this use a linear allocator instead to avoid fragmentation?

jbachorik: Linear allocator requires a clearly defined lifecycle. This is, unfortunately, not the case here.
I could try something like `_calltrace_buffer` but that would require locking. I did a quick check whether removing this allocation would help with the internal fragmentation (by having only one pre-allocated array and single lock, so reducing parallelism to 1) but the RSS growth I can observe is unaffected :(

PR #101, opened 2024-05-15
Sanitizers

r1viollet:
**What does this PR do?**:

Sanitizers
- Add a configuration for sanitizers
- Run TSan, UBsan, ASan
- Ensure unit tests run with aligned configurations (and remove the CMake files)

Various TSan fixes
- Adjust atomic operations to clarify some of the sequencing

Elf parser fixes
- Fix to ignore some out of bound elf symbols
- Addition of a test for the elf parser

CI
- Daily report of sanitizer issues

**Motivation**:

Better static analysis

**Additional Notes**:

You can now run various build configurations through
```
./gradlew testasan
./gradlew testubsan
./gradlew testdebug
./gradlew testrelease
```

Additionally the library can be used to run on other types of workloads.
```
UBSAN_OPTIONS="halt_on_error=1 abort_on_error=1 print_stacktrace=1" LSAN_OPTIONS="detect_leaks=0" ASAN_OPTIONS="detect_stack_use_after_return=1" LD_PRELOAD=/usr/lib/gcc/x86_64-linux-gnu/11/libasan.so DD_SERVICE=test_r1 java -javaagent:./temp/dd-java-agent.jar -Ddd.profiling.enabled=true -Ddd.profiling.ddprof.enabled=true -Ddd.profiling.upload.period=10 -Ddd.profiling.start-force-first=true -Ddd.profiling.ddprof.debug.lib=/home/r1viollet/dd/java-profiler/ddprof-lib/build/lib/main/asan/linux/x64/libjavaProfiler.so -jar ./temp/renaissance-gpl-0.15.0.jar akka-uct -r 5
```

To avoid deadly signal issues on ubuntu 22, I had to run 
```
sudo sysctl vm.mmap_rnd_bits=28
```
because of https://github.com/actions/runner-images/issues/9491

**How to test the change?**:

This is run within our CI daily.

**For Datadog employees**:


ddprof-lib/src/main/cpp/stackFrame_x64.cpp
```diff
@@ -95,7 +96,14 @@ bool StackFrame::unwindStub(instruction_t* entry, const char* name, uintptr_t& p
         pc = ((uintptr_t*)sp)[0] - 1;
         sp += 8;
         return true;
-    } else if (entry != NULL && *(unsigned int*)entry == 0xec8b4855) {
+    }
+    if(entry == nullptr) {
+        return false;
+    }
+    if (reinterpret_cast<uintptr_t>(entry) % alignof(unsigned int) != 0) {
```
r1viollet, line 103: I am not 100% sure this is true. To be verified.

ddprof-lib/src/test/run_tests.sh
```diff
@@ -0,0 +1,38 @@
...
+
+TMPDIR="${TMPDIR:-$(dirname $(mktemp))}"
+
+function build_test() {
+  NAME=$1
+  echo "Building test ${NAME}..."
+  # shellcheck disable=SC2038
+  files=$(find ${PROJECT_DIR}/build/obj/main/${CONFIG_NAME} -name "*.o" -print | tr '\n' ' ')
+  libdir=${GTEST_DIR}/configs/${CONFIG_NAME}/lib
+  g++ -std=c++11 -o ${TMPDIR}/${NAME}_${CONFIG_NAME} ${PROJECT_DIR}/src/test/cpp/${NAME}.cpp \
```
r1viollet, line 22: I would not go this route ;-)

gradle/configurations.gradle
```diff
@@ -0,0 +1,257 @@
...
+  }
+  return null
+}
+
+def locateLibasan() {
+  return locateLibrary('libasan')
+}
+
+def locateLibtsan() {
+  def libtsan_preinit = locate('libtsan_preinit.o')
```
r1viollet, line 61: I do not require this to build locally. Is this required in CI ?

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -1143,7 +1143,7 @@ Error Profiler::start(Arguments& args, bool reset) {
 
         _state = RUNNING;
         _start_time = time(NULL);
-        _epoch++;
+        __atomic_add_fetch(&_epoch, 1, __ATOMIC_SEQ_CST);
```
r1viollet, line 1146: I am not sure how important it is to put a SEQ_CST here. What is the use of epoch ?

richardstartin: it's only used to check the recording hasn't rolled since a wallclock sample was taken

r1viollet: Ok, that means if we do not retrieve the correct value, we are fine with it. I will go back to no ordering constraints:
 `__ATOMIC_RELAXED`

ddprof-lib/src/main/cpp/callTraceStorage.cpp
```diff
@@ -50,7 +50,9 @@ class LongHashTable {
         if (table != NULL) {
             table->_prev = prev;
             table->_capacity = capacity;
-            table->_size = 0;
+            // The reset is not useful with the anon mmap setting the memory is zeroed.
+            // However this silences a false positive and should not have a performance impact.
+            table->clear();
```
r1viollet, line 55: I could make this a conditional change. I do not think it matters in terms of performance, as it is a memset at startup and on growth.

ddprof-lib/src/main/cpp/stackFrame_x64.cpp
r1viollet: I think this is a reasonable assumption

.github/workflows/nightly.yml
```diff
@@ -1,11 +1,33 @@
...
+        with:
+          name: failures
+          path: ./artifacts
+      - name: Report failures
+        run: |
+          find ./artifacts -name 'failures_*' -exec cat {} \; > ./artifacts/failures.txt
+          scenarios=$(cat ./artifacts/failures.txt | tr '\n' ',')
+          echo "Failed scenarios: $scenarios"
+
+#          curl -X POST "${{ secrets.SLACK_WEBHOOK }}" \
```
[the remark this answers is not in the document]
jbachorik: Yes. Once we are ready to merge we should uncomment this block

ddprof-lib/src/main/cpp/callTraceStorage.cpp
jbachorik: Considering that the next sub-table is 2x size of the previous one the number of memset calls should not get high.

jbachorik (approved):
LGTM

PR #107, opened 2024-06-12
fix async-sample mutex

richardstartin:
**What does this PR do?**:
Fixes a bug ensuring we don't re-enter AsyncGetCallTrace

**Motivation**:

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:

### DIFF
--- a/ddprof-lib/src/main/cpp/asyncSampleMutex.h
+++ b/ddprof-lib/src/main/cpp/asyncSampleMutex.h
@@ -1,0 +1,38 @@
+#ifndef ASYNCSAMPLEMUTEX_H
+#define ASYNCSAMPLEMUTEX_H
+
+#include "threadLocalData.h"
+
+// controls access to AGCT
+class AsyncSampleMutex {
+private:
+    ThreadLocalData* _threadLocalData;
+    bool _acquired;
+
+    bool try_acquire() {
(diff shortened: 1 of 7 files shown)


MattAlp (approved):

ddprof-lib/src/main/cpp/asyncSampleMutex.h
```diff
@@ -0,0 +1,38 @@
...
+        }
+        return false;
+    }
+
+public:
+    AsyncSampleMutex(ThreadLocalData* threadLocalData) : _threadLocalData(threadLocalData) {
+        _acquired = try_acquire();
+    }
+
+    AsyncSampleMutex(AsyncSampleMutex& other) = delete;
```
r1viollet, line 25: NIT: if we delete this one, we can adjust the other constructors (rule of 5) to ensure we can not copy/move or use = operators.

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -683,7 +684,7 @@ void Profiler::recordSample(void* ucontext, u64 counter, int tid, jint event_typ
             int java_frames = 0;
             {
                 // Async events
-                AsyncSampleMutex mutex;
+                AsyncSampleMutex mutex(ProfiledThread::current());
```
r1viollet, line 687: Injecting the dependency is a good idea!

PR #109, opened 2024-06-24
Clean up the sanitizer runs

jbachorik:
**What does this PR do?**:
This is trying to calm down the rather noisy nightly sanitized runs.

**Motivation**:
Prevent 'warning' fatigue

**Additional Notes**:
I had to disable TSAN completely for nightlies. After adding a bunch of suppressions for various JVM related libraries most of the tests started either failing assertions or never finishing.
TSAN can still be run locally via `./gradlew testTsan` if you are on a linux box and tsan is fully installed (I could not get it working on the workspace box, though - it's still missing some tsan related *.so file which I have no idea where to get) - most likely it will make the unit tests failed because of messed up timing, or plainly livelock them, but you can try.

**How to test the change?**:

**For Datadog employees**:
- [x] JIRA: [PROF-10035]

[PROF-10035]: https://datadoghq.atlassian.net/browse/PROF-10035?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/livenessTracker.cpp
+++ b/ddprof-lib/src/main/cpp/livenessTracker.cpp
@@ -44,44 +44,47 @@ void LivenessTracker::cleanup_table(bool forced) {
     JNIEnv* env = VM::jni();
     jvmtiEnv* jvmti = VM::jvmti();
 
-    u64 start = OS::nanotime(), end;
+    int epoch_diff = (int) (target_gc_epoch - current);
 
     _table_lock.lock();
-    int epoch_diff = (int) (target_gc_epoch - current);
-    u32 sz, newsz = 0;
-    std::set<jclass> kept_classes;
-    for (u32 i = 0; i < (sz = _table_size); i++) {
-        if (_table[i].ref != nullptr && !env->IsSameObject(_table[i].ref, nullptr)) {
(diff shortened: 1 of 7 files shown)


ddprof-lib/src/main/cpp/livenessTracker.cpp
```diff
@@ -44,44 +44,47 @@ void LivenessTracker::cleanup_table(bool forced) {
     JNIEnv* env = VM::jni();
     jvmtiEnv* jvmti = VM::jvmti();
 
-    u64 start = OS::nanotime(), end;
+    int epoch_diff = (int) (target_gc_epoch - current);
 
     _table_lock.lock();
-    int epoch_diff = (int) (target_gc_epoch - current);
```
richardstartin, line 50: is this change necessary?

jbachorik: The whole block change was triggered by division by zero in the debug statement. While that is unlikely, getting the table of size 0 is still possible and moving the check for the non-zero check before doing anything more expensive seemed like an ok idea.

richardstartin (approved):

PR #110, opened 2024-06-25
Set 22.0.1-tem as SDKMAN env Java

MattAlp:
**What does this PR do?**:
This configures `sdkman` to use Temurin 22 as the Java version when working in this repo.

**Motivation**:
We need the correct JDK registered for using JVMTI symbols when building, and `dd-trace-java` requires `JAVA_HOME` to point to a JDK 8 build. This is a convenient way of mitigating forgetting to run `sdk use` or some other env var switching prior to building.

**Additional Notes**:
I'm considering adding a shell hook here and in dd-trace-java + relevant private repos that will automatically set the appropriate environment variables for these multi-JDK-dep projects. I've used https://github.com/Shopify/shadowenv in the past and like it quite a bit, any of https://direnv.net/#related-projects would work though.

**How to test the change?**:

**For Datadog employees**:


.sdkmanrc
```diff
@@ -0,0 +1,3 @@
+# Enable auto-env through the sdkman_auto_env config
+# Add key=value pairs of SDKs to use below
+java=22.0.1-tem
```
r1viollet, line 3: should we have a matching README requirement section if we use sdkman ?

jbachorik:
Nice, using the rc file :)

I wonder if we need JDK 22 for anything - we still need to keep the bytecode of the Java part compatible with Java 8. 
What symbols we need from JDK 22?

jbachorik:
@MattAlp Yes, the baseline is JDK 11 for us - because of the JVMTI sampler.

We can be progressive and use the tip (JDK 22) - we are setting the compatibility in ddprof-lib to Java 8 so we are fine at least for JDK 22

jbachorik (approved):

.sdkmanrc
r1viollet: There are currently pre-requisites to building the project.
Using sdkman simplifies this. We can mention it as an alternative or we can just put this as the standard path to building the project.
This is not a blocking comment to merge.

jbachorik:
Is this still relevant? If not, please close it - it's been dangling for long enough time.

PR #111, opened 2024-06-25
Fix array decay truncating native thread names

MattAlp:
**What does this PR do?**:
This addresses PROF-9879 by fixing an array decay issue causing truncation of thread names down to pointer size, rather than available buffer size

**Motivation**:

**Additional Notes**:

We could look at changing how we do the ownership / allocation here as well, cc @r1viollet - but this will resolve the issue safely

**How to test the change?**:
I haven't added a test for this- it's a classic C/C++ bug which thankfully couldn't cause any fatal issues. I have reviewed that this change won't cause issues in the MacOS profiler, which uses `pthread_getname_np` rather than `/proc/self/stat`.

**For Datadog employees**:

### DIFF
--- a/ddprof-lib/src/main/cpp/profiler.cpp
+++ b/ddprof-lib/src/main/cpp/profiler.cpp
@@ -932,8 +932,9 @@ void Profiler::updateNativeThreadNames() {
     ThreadList* thread_list = OS::listThreads();
     for (int tid; (tid = thread_list->next()) != -1; ) {
         _thread_info.updateThreadName(tid, [](int tid) -> std::unique_ptr<char[]> {
-            char *name_buf = new char[64];
-            if (OS::threadName(tid, name_buf, sizeof(name_buf))) {
+            const size_t buffer_size = 64;
+            char *name_buf = new char[buffer_size];
+            if (OS::threadName(tid, name_buf, buffer_size)) {
                 return std::unique_ptr<char[]>(name_buf);
             }
             delete[] name_buf;


ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -932,8 +932,9 @@ void Profiler::updateNativeThreadNames() {
     ThreadList* thread_list = OS::listThreads();
     for (int tid; (tid = thread_list->next()) != -1; ) {
         _thread_info.updateThreadName(tid, [](int tid) -> std::unique_ptr<char[]> {
-            char *name_buf = new char[64];
-            if (OS::threadName(tid, name_buf, sizeof(name_buf))) {
```
r1viollet, line 936: ah yes, sizeof operator on a pointer
Good catch

richardstartin (approved):
LGTM

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -932,8 +932,9 @@ void Profiler::updateNativeThreadNames() {
...
+            const size_t buffer_size = 64;
+            char *name_buf = new char[buffer_size];
```
r1viollet, line 936: So this is fixed size. The sizeof would have worked if we had created the buffer without a dynamic allocation.
The better practice would have been to use a standard container: std::array and size operators

jbachorik (approved):

ddprof-lib/src/main/cpp/profiler.cpp
r1viollet: Apologies for being unclear, the suggestions are as follows:
- use a static C style buffer: this is consistent with the coding style of the project, then sizeof would work (or continue using a const or constexpr buffer_size to avoid possible issues)
- use a static C++ style container (std::array): this is not consistent with the coding style (and possibly C++17). This allows you to just start using size operators which would be constexpr.
Feel free to chose either options.

PR #114, opened 2024-07-10
J9 integration cleanup

jbachorik:
**What does this PR do?**:
It adjusts the activation modes for wallclock and cpu profiler considering the availability and stability of ASGCT since the most recent JDK J9 updated.
It completely removes the allocation profiler because it is based on JVMTI allocation callbacks and as such the overhead is terrible. 

**Motivation**:
I want to get the J9 support into a final state where we can clearly claim what is supported and what is not and why some features can not be supported.

**Additional Notes**:
The JVMTI based j9 cpu/wallclock (non-ASGCT) is still randomly crashy on anything but JDK 8. I kept the possibility to 'force-enable' it in case a customer is ok with the risk. It will be up to the dd-trace-java to use safe defaults (eg. enabling J9 cpu/wallclock only for JDK versions with stable ASGCT **AND** fixed jmethodID bug.

**How to test the change?**:

**For Datadog employees**:
- [x] JIRA: [PROF-10050] 

[PROF-10050]: https://datadoghq.atlassian.net/browse/PROF-10050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


richardstartin (approved):

ddprof-lib/src/main/cpp/j9Ext.cpp
```diff
@@ -54,27 +54,6 @@ bool J9Ext::initialize(jvmtiEnv* jvmti, const void* j9thread_self) {
...
-                break;
-            }
-        }
-       jvmti->Deallocate((unsigned char*)ext_events);
-    }
-
     return _GetOSThreadID != NULL && _GetStackTraceExtended != NULL && _GetAllStackTracesExtended != NULL;
 }
 
-void J9Ext::tryForceKeepJMethodIDs() {
```
richardstartin, line 74: did this not work as expected?

jbachorik: No :(
In one of the followup PRs in J9 they changed the approach to only mark classloaders of classes in frames handed out by ASGCT to retain jmethodids. And since the jmethodid crash is random, I got lucky for a few runs to be fooled into thinking it did :(

PR #118, opened 2024-07-17
PROF-10079: jvmti based wallclock sampler

richardstartin:
**What does this PR do?**:
Introduces a new wallclock sampler not based on signals, which uses JVMTI APIs instead.

**Motivation**:

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:

### DIFF
--- a/ddprof-lib/src/main/cpp/wallClock_jvmti.cpp
+++ b/ddprof-lib/src/main/cpp/wallClock_jvmti.cpp
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2018 Andrei Pangin
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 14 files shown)


ddprof-lib/src/main/cpp/wallClock_jvmti.cpp
```diff
@@ -0,0 +1,112 @@
...
+    javaThreads.reserve(_reservoir_size);
+    ReservoirSampler<int> reservoir(_reservoir_size);
+    int self = OS::threadId();
+    u64 counter = 0;
+
+
+    while (_running.load(std::memory_order_relaxed)) {
+        int numFailures = 0;
+        int threadsAlreadyExited = 0;
+        std::vector<int> sample = reservoir.sample(tids);
```
jbachorik, line 85: Where is the `tids` vector filled with the TIDs?
I might be blind but I can't seem to find it :(

jbachorik: In the ASGCT version I see the thread_filter taking care of that - https://github.com/DataDog/java-profiler/pull/118/files#diff-a368edb3b62aef3c92f4e699fc69e1a6fabed5d158b808dabda9af2018073ddaR158-R170

richardstartin: good catch, the test wasn't testing what I thought it was. I've added some extra tests to verify that the profiler under test is the one in use.

r1viollet:
I will convert to draft to ensure we can keep this around for documentation, though we are not planning to integrate this as of now.

jbachorik:
@r1viollet I hope we will get this finished. The ASGCT and I/O interrupted by signals issues would be great to avoid...

PR #121, opened 2024-08-06
Crash fix - Concurrent unloaded mappings

r1viollet:
**What does this PR do?**:

If a mapping is unloaded while we are parsing the mappings we should not be accessing the memory.
This commit adds a function to check that the memory is readable before trying to access the content of the mapping.

**Motivation**:

Customer crash was reported here:
https://github.com/DataDog/dd-trace-java/issues/7372
The full information is only available in the associated Zendesk ticket.

I disassembled to ensure we were crashing in these conditions and was able to reproduce a crash when forcing an unmapping of the mapping being parsed.

**Additional Notes**:

NA

**How to test the change?**:

I added a test that tries to reproduce the race condition.
It does not reproduce this 100% of the time.

**For Datadog employees**:

### DIFF
--- a/ddprof-lib/src/main/cpp/profiler.cpp
+++ b/ddprof-lib/src/main/cpp/profiler.cpp
@@ -183,7 +183,9 @@ inline u32 Profiler::getLockIndex(int tid) {
 }
 
 void Profiler::updateSymbols(bool kernel_symbols) {
+    Symbols::setupSignalHandlers();
     Symbols::parseLibraries(&_native_libs, kernel_symbols);
+    Symbols::restoreSignalHandlers();
 }
 
 void Profiler::mangle(const char* name, char* buf, size_t size) {
--- a/ddprof-lib/src/main/cpp/safeAccess.h
+++ b/ddprof-lib/src/main/cpp/safeAccess.h
@@ -18,6 +18,7 @@
 #define _SAFEACCESS_H
 
 #include <stdint.h>
+
 #include "arch.h"
 
 
@@ -46,7 +47,14 @@ class SafeAccess {
     }
 
     static uintptr_t skipLoad(uintptr_t pc) {
+#ifndef __SANITIZE_ADDRESS__
         if (pc - (uintptr_t)load < 16) {
+#else
+        // asan significantly increases the size of the load function
+        // checking disassembled code can help adjust the value
+        // gdb --batch -ex 'disas _ZN10SafeAccess4loadEPPv' ./elfparser_ut
+        if (pc - (uintptr_t)load < 132) {
+#endif
 #if defined(__x86_64__)
             return *(u16*)pc == 0x8b48 ? 3 : 0;  // mov rax, [reg]
 #elif defined(__i386__)
--- a/ddprof-lib/src/main/cpp/symbols.h
+++ b/ddprof-lib/src/main/cpp/symbols.h
@@ -27,6 +27,10 @@ class Symbols {
     static bool _have_kernel_symbols;
 
   public:
+    // Ensures we can parse mappings even with concurrent unloading
+    static void setupSignalHandlers();
+    static void restoreSignalHandlers();
+
     static void parseKernelSymbols(CodeCache* cc);
     static void parseLibraries(CodeCacheArray* array, bool kernel_symbols);
     // The clear function is mainly for test purposes
--- a/ddprof-lib/src/main/cpp/symbols_linux.cpp
+++ b/ddprof-lib/src/main/cpp/symbols_linux.cpp
@@ -28,14 +28,16 @@
 #include <unistd.h>
 #include <fcntl.h>
 #include <linux/limits.h>
-#include <signal.h>
 #include <setjmp.h>
+#include <signal.h>
 
 #include "symbols.h"
 #include "symbols_linux.h"
 #include "dwarf.h"
 #include "log.h"
+#include "os.h"
 #include "safeAccess.h"
+#include "stackFrame.h"
 
 
 ElfSection* ElfParser::findSection(uint32_t type, const char* name) {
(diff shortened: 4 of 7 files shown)


ddprof-lib/src/main/cpp/symbols.h
```diff
@@ -29,7 +29,7 @@ class Symbols {
   public:
     static void parseKernelSymbols(CodeCache* cc);
     static void parseLibraries(CodeCacheArray* array, bool kernel_symbols);
-
+    static void clear_parsed_caches();
```
r1viollet, line 32: add a comment that this is for test purposes

ddprof-lib/src/main/cpp/symbols_linux.cpp
```diff
@@ -418,8 +427,58 @@ void Symbols::parseKernelSymbols(CodeCache* cc) {
...
+
+#include "symbols_linux.h"
+
+// Function to check if memory is accessible
+static bool is_memory_accessible(const char* addr) {
+
+    // Try to access the memory
+    if (sigsetjmp(jump_buffer, 1) == 0) {
+        char access_start = addr[0]; // Attempt to read the memory
+        char access_end = addr[sizeof(ElfHeader)]; // Attempt to read the memory
```
r1viollet, line 470: todo: figure out how to remove the print and avoid optimization

ddprof-lib/src/main/cpp/symbols_linux.cpp
```diff
@@ -484,9 +559,15 @@ void Symbols::parseLibraries(CodeCacheArray* array, bool kernel_symbols) {
                 if (_parsed_inodes.insert(inode).second) {
                     if (inode == last_inode) {
                         // If last_inode is set, image_base is known to be valid and readable
-                        ElfParser::parseFile(cc, image_base, map.file(), true);
-                        // Parse program headers after the file to ensure debug symbols are parsed first
-                        ElfParser::parseProgramHeaders(cc, image_base, map_end, MUSL);
+                        bool res = ElfParser::parseFile(cc, image_base, map.file(), true);
```
r1viollet, line 562: minor res is unused

ddprof-lib/src/main/cpp/symbols_linux.cpp
```diff
@@ -418,8 +427,74 @@ void Symbols::parseKernelSymbols(CodeCache* cc) {
     fclose(f);
 }
 
+namespace {
```
r1viollet, line 430: Talking with @nsavoire about the approach:
- It would be better to rely on an equivalent of `process_vm_readv`.
That would avoid the problems associated with segv handlers.
- If we go for the segvhandler solution
there are helpers available within async-profiler to do this in a safer way.
- The current solution only reduces the risk of a race condition (as we can still have a mapping unloaded between time of check and time of use)

Sources that relate to this:
- https://github.com/abseil/abseil-cpp/blob/3848ed7f1b13259a8ddd35dc56bf216ecd36d968/absl/debugging/internal/address_is_readable.cc#L52
- https://github.com/async-profiler/async-profiler/blob/179b0b1285c656bb7afae3f144d540256547b522/src/profiler.cpp#L835

jbachorik: > The current solution only reduces the risk of a race condition (as we can still have a mapping unloaded between time of check and time of use)

Indeed. We would need to use `SafeAccess:load*` calls for all memory loads in the elf parser if the mapping can just disappear in the middle of processing :(

r1viollet: I am assuming the handlers are not setup at that time (when we are doing the parse library function).
I'll see if I can make a version using SafeAcess, though I am not too familiar with profiler init sequences as well as the whole SafeAccess infrastructure.
The important part to protect is the access to the section header, which is the first read into the mapping.
```
_header->e_shoff
```

r1viollet:
New version available. I went for the SafeAccess function call.
I have not yet setup SIGBUS.

ddprof-lib/src/main/cpp/symbols_linux.cpp
```diff
@@ -418,6 +456,64 @@ void Symbols::parseKernelSymbols(CodeCache* cc) {
...
+    }
+}
+}
+
+void Symbols::setupSignalHandlers() {
+    // Replacing the segvhandler only while parsing the libraries
+    // It can be unsafe to touch signal handlers in some Java distributions
+    // The assumption is that doing so for a short time is safe
+    // TODO: verify this assumption
+    orig_segvHandler = OS::replaceSigsegvHandler(segvHandler);
```
r1viollet, line 507: here I removed some of the checks that were available within profiler.cpp on signal handlers.
Is my assumption LGTM ?

r1viollet, line 507: TODO: SIGBUS

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -183,7 +183,9 @@ inline u32 Profiler::getLockIndex(int tid) {
 }
 
 void Profiler::updateSymbols(bool kernel_symbols) {
+    Symbols::setupSignalHandlers();
```
r1viollet, line 186: It would be nice to unify some of the signal handler code and add some tests around these tricky mechanisms.

ddprof-lib/src/main/cpp/safeAccess.h
```diff
@@ -46,7 +47,14 @@ class SafeAccess {
     }
 
     static uintptr_t skipLoad(uintptr_t pc) {
+#ifndef __SANITIZE_ADDRESS__
         if (pc - (uintptr_t)load < 16) {
+#else
+        // asan significantly increases the size of the load function
+        // checking disassembled code can help adjust the value
+        // gdb --batch -ex 'disas _ZN10SafeAccess4loadEPPv' ./elfparser_ut
+        if (pc - (uintptr_t)load < 132) {
```
r1viollet, line 56: This probably depends on compiler versions.
However for test purposes, I would assume this is enough. WDYT ?

r1viollet: I guess I can assert on consecutive function addresses.

r1viollet:
Closing in favour of https://github.com/DataDog/java-profiler/pull/124

PR #124, opened 2024-08-12
Properly handle concurrent unloaded mappings

r1viollet:
**What does this PR do?**:

If a mapping is unloaded while we are parsing the mappings we should not be accessing the memory.
@apangin had the idea to use dl_iterate_phdr to take the dl_open lock.

**Motivation**:

Customer crash was reported here:
https://github.com/DataDog/dd-trace-java/issues/7372
The full information is only available in the associated Zendesk ticket.

I disassembled to ensure we were crashing in these conditions and was able to reproduce a crash when forcing an unmapping of the mapping being parsed.

**Additional Notes**:

NA

**How to test the change?**:

I added a test that tries to reproduce the race condition.
It does not reproduce this 100% of the time.

**For Datadog employees**:

### DIFF
--- a/ddprof-lib/src/test/resources/small-lib/Makefile
+++ b/ddprof-lib/src/test/resources/small-lib/Makefile
@@ -1,0 +1,3 @@
+TARGET_DIR = ../build/test/resources/small-lib
+all:
+	g++ -fPIC -shared -o libsmall-lib.so small_lib.cpp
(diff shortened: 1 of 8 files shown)


r1viollet:
This seems to be crashing in CI with the parametric test.

ddprof-lib/src/test/resources/small-lib/Makefile
```diff
@@ -0,0 +1,3 @@
+TARGET_DIR = ../build/test/resources/small-lib
+all:
+	g++ -fPIC -shared -o libsmall-lib.so small_lib.cpp
```
jbachorik, line 3: This is not necessary any more, right? The library is built in the gradle task.

jbachorik: Ok, let's keep this in for the time being.

jbachorik (approved):

PR #127, opened 2024-09-06
Deallocate jclass reference promptly

jbachorik:
**What does this PR do?**:
It adds defensive deallocation of jclass references to reduce the chance of crashing due to mismanaged jmethodID in JVM.

**Motivation**:
See https://github.com/async-profiler/async-profiler/pull/981 and https://bugs.openjdk.org/browse/JDK-8268364

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:
- [x] If this PR touches code that signs or publishes builds or packages, or handles
- [x] JIRA: [PROF-10461]

[PROF-10461]: https://datadoghq.atlassian.net/browse/PROF-10461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -133,7 +133,7 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
   jvmtiEnv *jvmti = VM::jvmti();
 
   jvmtiPhase phase;
-  jclass method_class;
+  jclass method_class = nullptr;
   // invariant: these strings must remain null, or be assigned by JVMTI
   char *class_name = nullptr;
   char *method_name = nullptr;
@@ -243,6 +243,10 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
       class_name_id = _classes->lookup("");
       method_name_id = _symbols.lookup("jvmtiError");
       method_sig_id = _symbols.lookup("()L;");
+    }
+
+    if (method_class) {
+      _jni->DeleteLocalRef(method_class);
     }
 
     mi->_class = class_name_id;
--- a/ddprof-lib/src/main/cpp/flightRecorder.h
+++ b/ddprof-lib/src/main/cpp/flightRecorder.h
@@ -268,6 +268,7 @@ public:
   Dictionary _symbols;
 
 private:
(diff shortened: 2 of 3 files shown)


r1viollet:
I am not sure I understand why we have differences with upstream PR.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -123,16 +123,15 @@ void Lookup::cutArguments(char* func) {
     }
 }
 
-void Lookup::fillJavaMethodInfo(MethodInfo* mi, jmethodID method, bool first_time) {
-    JNIEnv* jni = VM::jni();
+void Lookup::fillJavaMethodInfo(JNIEnv* jni, MethodInfo* mi, jmethodID method, bool first_time) {
     if (jni->PushLocalFrame(64) != 0) {
```
richardstartin, line 127: Do we really need this patch? We theoretically have the protection of this local frame which will get popped very quickly. Do we fail with the reproducer program from upstream?

jbachorik: The reproducer has nothing to do with the profiler. It is only JVMTI code so I expect it will fail if I run it on my machine.
And the patch is lowering the likelihood of the crash with [JDK-8268364 ](https://bugs.openjdk.org/browse/JDK-8268364). Without that fix this patch is mostly a no-op, just being a nicer citizen and releasing local refs promptly (they will get released when the JNI call returns to JVM - but in this case it means once the whole recording is dumped).

jbachorik:
> I am not sure I understand why we have differences with upstream PR.

Because I started working on this before the final merge in the upstream. I have updated the diffs to be in-sync with what has been done upstream.

ddprof-lib/src/main/cpp/flightRecorder.cpp
richardstartin: But because there's a local frame here, the reference will be will get released when the frame is [popped](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#PopLocalFrame) on method exit anyway. I.e. I think we're already working around this.

jbachorik:
Not needed. We are using push/pop frame to achieve the same effect already.

PR #131, opened 2024-09-17
Fix trivial leak in ThreadFilter

jbachorik:
**What does this PR do?**:
It fixes a trivial leak when the destructor of `ThreadFilter` class did not release the memory allocated for `_bitmaps`.

**Motivation**:
Mostly just a code cleanup - with our current usage the `ThreadFilter` instance will remain active during the whole lifespan of the profiled application so the memory would not really 'leak'. But it is better to have a correct destructor if the usage of `ThreadFilter` would change in the future.

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:
- [] If this PR touches code that signs or publishes builds or packages, or handles

### DIFF
--- a/ddprof-lib/src/main/cpp/threadFilter.cpp
+++ b/ddprof-lib/src/main/cpp/threadFilter.cpp
@@ -43,6 +43,9 @@ ThreadFilter::~ThreadFilter() {
       OS::safeFree(_bitmap[i], BITMAP_SIZE);
     }
   }
+  if (_bitmap) {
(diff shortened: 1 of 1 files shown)


richardstartin (approved):
no functional impact but keeps tsan happy?

jbachorik:
This was actually not pointed out by tsan - and we are not running it by default anyway.
I came to this while trying to rule out a sneaky memory leak in profiler before blaming malloc fragmentation again,

PR #134, opened 2024-09-24
Improve robustness of the crash signal handler

jbachorik:
**What does this PR do?**:
This is an attempt to increase the robustness of the crash signal handler by
- preventing rerunning the signal setup routine (quite unlikely to happen, but just in case)
- using separate stack space for the crash signal handlers to avoid triggering stack overflow
- using TLS guard to prevent recursive crash handling in case the handler itself would cause sigseg
- checking the PC for not being the cause of the original fault - if it is, we can not meaningfully recover and just call the JVM crash handler

**Motivation**:

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:
- [x] JIRA: [PROF-10605]

[PROF-10605]: https://datadoghq.atlassian.net/browse/PROF-10605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/os_macos.cpp
+++ b/ddprof-lib/src/main/cpp/os_macos.cpp
@@ -216,6 +216,7 @@ SigAction OS::replaceSigsegvHandler(SigAction action) {
   sigaction(SIGSEGV, NULL, &sa);
   SigAction old_action = sa.sa_sigaction;
   sa.sa_sigaction = action;
+  sa.sa_flags = SA_SIGINFO | SA_ONSTACK;
   sigaction(SIGSEGV, &sa, NULL);
   return old_action;
 }
--- a/ddprof-lib/src/main/cpp/os_macos.cpp
+++ b/ddprof-lib/src/main/cpp/os_macos.cpp
@@ -216,7 +216,7 @@ SigAction OS::replaceSigsegvHandler(SigAction action) {
   sigaction(SIGSEGV, NULL, &sa);
   SigAction old_action = sa.sa_sigaction;
   sa.sa_sigaction = action;
-  sa.sa_flags = SA_SIGINFO | SA_ONSTACK;
+  sa.sa_flags = SA_SIGINFO;
   sigaction(SIGSEGV, &sa, NULL);
   return old_action;
 }
--- a/ddprof-lib/src/main/cpp/profiler.cpp
+++ b/ddprof-lib/src/main/cpp/profiler.cpp
@@ -967,23 +967,6 @@ bool Profiler::crashHandler(int signo, siginfo_t *siginfo, void *ucontext) {
 void Profiler::setupSignalHandlers() {
   // do not re-run the signal setup (run only when VM has not been loaded yet)
   if (VM::java_version() > 0 && !VM::loaded()) {
-    // register alternative stack for sighandlers
-    int ALT_STACK_SIZE = SIGSTKSZ * 4;
-    stack_t ss;
-    memset(&ss, 0, sizeof(ss));
-    ss.ss_sp = malloc(ALT_STACK_SIZE);
-    if (ss.ss_sp == NULL) {
-        perror("malloc");
-        exit(EXIT_FAILURE);
-    }
-    ss.ss_size = ALT_STACK_SIZE;
-    ss.ss_flags = 0;
-
-    if (sigaltstack(&ss, NULL) == -1) {
-        perror("sigaltstack");
-        exit(EXIT_FAILURE);
-    }
-
     // HotSpot and J9 tolerate interposed SIGSEGV/SIGBUS handler; other JVMs
     // probably not
     orig_segvHandler = OS::replaceSigsegvHandler(segvHandler);
--- a/ddprof-lib/src/main/cpp/profiler.cpp
+++ b/ddprof-lib/src/main/cpp/profiler.cpp
@@ -38,7 +38,6 @@
 #include <algorithm>
 #include <dlfcn.h>
 #include <fstream>
-#include <functional>
 #include <memory>
 #include <set>
 #include <signal.h>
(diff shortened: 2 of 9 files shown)


ddprof-lib/src/main/cpp/os_macos.cpp
```diff
@@ -216,6 +216,7 @@ SigAction OS::replaceSigsegvHandler(SigAction action) {
   sigaction(SIGSEGV, NULL, &sa);
   SigAction old_action = sa.sa_sigaction;
   sa.sa_sigaction = action;
+  sa.sa_flags = SA_SIGINFO | SA_ONSTACK;
```
richardstartin, line 219: What is the memory impact of doing this with the extensive usage of `SafeAccess` as well as segfaults originating in the JVM? How do signal handlers with different flags compose?

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -922,35 +930,61 @@ bool Profiler::crashHandler(int signo, siginfo_t *siginfo, void *ucontext) {
...
+    }
     return true;
   }
 
   if (VM::isHotspot()) {
     // the following checks require vmstructs and therefore HotSpot
-    StackWalker::checkFault();
+
+    // this check can longjmp to a completely different location - need to call exitCrashHandler() before
+    StackWalker::checkFault(std::bind(&ProfiledThread::exitCrashHandler, thrd));
```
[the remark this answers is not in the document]
jbachorik: I guess so. You are right, this is a terrible choice for signal handler.

ddprof-lib/src/main/cpp/os_macos.cpp
jbachorik: Memory impact is about 1KB per invocation required on stack.
The signal handlers are processed one by one so the flags are not composed/merged and for each handler the information to be delivered is prepared according to the flags.

I would say that the extra overhead should be minimal - even with safe access we are trying to avoid accessing known invalid locations so getting sigseg and entering this handler should be rather rare.

richardstartin (approved):

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -906,14 +897,30 @@ void Profiler::busHandler(int signo, siginfo_t *siginfo, void *ucontext) {
 }
 
 bool Profiler::crashHandler(int signo, siginfo_t *siginfo, void *ucontext) {
+  ProfiledThread* thrd = ProfiledThread::current();
+  if (thrd != nullptr && !thrd->enterCrashHandler()) {
```
r1viollet, line 901: NIT: you would usually have a RAII object here, to just call exit crashtracker on object destruction.

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -906,14 +897,30 @@ void Profiler::busHandler(int signo, siginfo_t *siginfo, void *ucontext) {
...
+    // we are already in a crash handler; don't recurse!
+    return false;
+  }
+  uintptr_t fault_address = (uintptr_t)siginfo->si_addr;
   StackFrame frame(ucontext);
   uintptr_t pc = frame.pc();
+  if (pc == fault_address) {
```
r1viollet, line 908: This did not compute initially for me, though you are correct that it can happen :+1:

ddprof-lib/src/main/cpp/profiler.cpp
r1viollet: the `longjmp` makes this hard, got it

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -922,35 +929,44 @@ bool Profiler::crashHandler(int signo, siginfo_t *siginfo, void *ucontext) {
...
 void Profiler::setupSignalHandlers() {
-  orig_trapHandler =
-      OS::installSignalHandler(SIGTRAP, Profiler::trapHandlerEntry);
-  if (orig_trapHandler == (void *)SIG_DFL ||
-      orig_trapHandler == (void *)SIG_IGN) {
-    orig_trapHandler = NULL;
-  }
-  if (VM::java_version() > 0) {
+  // do not re-run the signal setup (run only when VM has not been loaded yet)
+  if (VM::java_version() > 0 && !VM::loaded()) {
```
r1viollet, line 969: I am not sure I understand this.
Would you not expect to chain signal handlers once the VM is loaded (ensuring you do not get overridden following the load of the VM ?).

jbachorik: We want to run this on a vm-init callback only. The async-profiler can do 'late initialization' and, theoretically, run this multiple times. While it really does not change anything for us, having the logic in-sync with async-profiler would make it easier to reason about any changes we might cherry-pick

ddprof-lib/src/main/cpp/profiler.cpp
r1viollet, line 901: this puts some weight on the thread object to be always valid or null

ddprof-lib/src/main/cpp/stackWalker.cpp
```diff
@@ -454,9 +454,12 @@ int StackWalker::walkVM(void *ucontext, ASGCT_CallFrame *frames, int max_depth,
...
 }
 
-void StackWalker::checkFault() {
+void StackWalker::checkFault(ProfiledThread* thrd) {
   VMThread *vm_thread = VMThread::current();
   if (vm_thread != NULL && sameStack(vm_thread->exception(), &vm_thread)) {
+    if (thrd) {
+      thrd->exitCrashHandler();
+    }
     longjmp(*(jmp_buf *)vm_thread->exception(), 1);
```
r1viollet, line 463: should we `std::atomic_thread_fence` ?

ddprof-lib/src/main/cpp/stackWalker.cpp
```diff
@@ -454,9 +454,12 @@ int StackWalker::walkVM(void *ucontext, ASGCT_CallFrame *frames, int max_depth,
   return depth;
 }
 
-void StackWalker::checkFault() {
+void StackWalker::checkFault(ProfiledThread* thrd) {
   VMThread *vm_thread = VMThread::current();
   if (vm_thread != NULL && sameStack(vm_thread->exception(), &vm_thread)) {
+    if (thrd) {
+      thrd->exitCrashHandler();
```
r1viollet, line 461: This does not work, we are doing a longjmp, so we should reset before exiting. Here the increments will build up.

PR #139, opened 2024-10-16
Crash handler recursion protection - Fix

r1viollet:
**What does this PR do?**:

When exiting through the longjmp path, we should reset the depth instead of decrementing.

There were a lot of warnings in the CI. I fixed the most obvious ones.
The PR can be reviewed with separate commits.

**Motivation**:

Jaroslav witnessed a crash in the safe access function (meaning we were not preventing the safe access crashes).

**Additional Notes**:

**How to test the change?**:

I have not been able to reproduce as of now.

**For Datadog employees**:
- [X] JIRA: [JIRA-XXXX]

### DIFF
--- a/ddprof-lib/src/main/cpp/stackWalker.cpp
+++ b/ddprof-lib/src/main/cpp/stackWalker.cpp
@@ -458,7 +458,7 @@ void StackWalker::checkFault(ProfiledThread* thrd) {
   VMThread *vm_thread = VMThread::current();
   if (vm_thread != NULL && sameStack(vm_thread->exception(), &vm_thread)) {
     if (thrd) {
-      thrd->exitCrashHandler();
+      thrd->resetCrashHandler();
     }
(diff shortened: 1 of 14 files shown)


ddprof-lib/src/main/cpp/stackWalker.cpp
```diff
@@ -458,7 +458,7 @@ void StackWalker::checkFault(ProfiledThread* thrd) {
   VMThread *vm_thread = VMThread::current();
   if (vm_thread != NULL && sameStack(vm_thread->exception(), &vm_thread)) {
     if (thrd) {
-      thrd->exitCrashHandler();
+      thrd->resetCrashHandler();
```
r1viollet, line 461: this is the main change

jbachorik: Could you add a comment here about why the reset needs to be done here?

jbachorik (approved):

Issue #136, opened 2024-10-01
Agent incorrectly assumes that JVM version == JDK version


asm0dey:
We at BellSoft have JDK 8 with underlying JVM 17. If I run my Java8 application on my JDK with DataDog agent, I get a crash like [this one](https://github.com/user-attachments/files/17211655/he_err_pid1.log)

In reality, it's possible (and correct) to check the JVM version, not JDK. For example, the property `java.vm.version` contains `17.0.9+10-LTS` in our case.

It seems to be related to these lines: https://github.com/DataDog/java-profiler/blob/11f0ba72528ff254b1089f41c3be17f687963136/ddprof-lib/src/main/cpp/vmStructs.cpp#L592

jbachorik:
Aargh ... 
Would it possible to find something distinctive? Java 8 will put `25.*` into that system property. While we could hard-code '25 == 8', once we get proper JDK 25, this will get incredibly messy.

I just want to make sure what would be the right cause of action - currently, we are reading `java.vm.version` to retrieve the JVM version. But it does return value 17, since you are running JVM 17 there. But the version returned by `java.runtime.version` would be '8.*' I assume, correct? And the code that is crashing is actually using the JDK functionality and as such we would have to have two versions - vm and jdk, right? And then check all occasions where we guard some functionality based on the Java version and decide whether it is JVM or JDK related.

jbachorik:
> If it will help I can privately send you a build of our JDK for you to check everything you need

yes, this would be very helpful

dpochepk:
> Aargh ... Would it possible to find something distinctive? Java 8 will put `25.*` into that system property. While we could hard-code '25 == 8', once we get proper JDK 25, this will get incredibly messy.
> 
> I just want to make sure what would be the right cause of action - currently, we are reading `java.vm.version` to retrieve the JVM version. But it does return value 17, since you are running JVM 17 there. But the version returned by `java.runtime.version` would be '8.*' I assume, correct? And the code that is crashing is actually using the JDK functionality and as such we would have to have two versions - vm and jdk, right? And then check all occasions where we guard some functionality based on the Java version and decide whether it is JVM or JDK related.

Yes. You're correct. java.runtime.version still has something like "1.8.0_*".

asm0dey:
@jbachorik I've sent a mail to your address I found here on GH, could you please answer when you have a minute?

jbachorik:
Fixed in #142

PR #141, opened 2024-10-23
VMStructs: Undefined behaviour fix

r1viollet:
**What does this PR do?**:

Avoid shifting a negative value
This was raised by asan nightly runs

**Motivation**:

```
> Task :ddprof-test:testClasses
ddprof-lib/src/main/cpp/vmStructs.cpp:816:34: runtime error: left shift of negative value -1
    #0 0x7f6f6bd20631 in ScopeDesc::readInt()
    #1 0x7f6f6bce54a6 in ScopeDesc::decode(int)
    #2 0x7f6f6bce7fb0 in StackWalker::walkVM(void*, ASGCT_CallFrame*, int, void const*, void const*)
```

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:
- [ ] JIRA: [JIRA-XXXX]

### DIFF
--- a/ddprof-lib/src/main/cpp/vmStructs.cpp
+++ b/ddprof-lib/src/main/cpp/vmStructs.cpp
@@ -808,17 +808,22 @@ int NMethod::findScopeOffset(const void *pc) {
 }
 
 int ScopeDesc::readInt() {
-  unsigned char c = *_stream++;
-  unsigned int n = c - _unsigned5_base;
-  if (c >= 192) {
-    for (int shift = 6;; shift += 6) {
-      c = *_stream++;
-      n += (c - _unsigned5_base) << shift;
-      if (c < 192 || shift >= 24)
-        break;
+    unsigned char c = *_stream++;
+    if (c < _unsigned5_base) {
+        return 0;
     }
-  }
-  return n;
+    unsigned int n = c - _unsigned5_base;
+    if (c >= 192) {
+        for (int shift = 6; ; shift += 6) {
+            c = *_stream++;
+            if (c < _unsigned5_base) {
+                break;
+            }
+            n += (c - _unsigned5_base) << shift;
+            if (c < 192 || shift >= 24) break;
+        }
+    }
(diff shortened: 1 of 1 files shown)


ddprof-lib/src/main/cpp/vmStructs.cpp
```diff
@@ -808,17 +808,22 @@ int NMethod::findScopeOffset(const void *pc) {
...
+    if (c < _unsigned5_base) {
+        return 0;
     }
-  }
-  return n;
+    unsigned int n = c - _unsigned5_base;
+    if (c >= 192) {
+        for (int shift = 6; ; shift += 6) {
+            c = *_stream++;
+            if (c < _unsigned5_base) {
```
r1viollet, line 819: is this not an indication of a race condition and we should retry the read ?
The previous byte indicated we were expecting something, so having 0 here is unexpected.

richardstartin: reads of `NMethod`s are generally racy - we're reading them from a process-wide codecache which can change even when the sampled thread is halted in a signal handler.

r1viollet: The more valid fix would then be to bail out.

jbachorik: It seems that the problem is rather insidious - the `_unsigned5_base` is set up depending on the JVM version but that is subtly broken for Java 8 - see https://github.com/DataDog/java-profiler/issues/136

I am going to fix that problem first and if that does not stop the asan errors we can revisit this later.

jbachorik: It seems that after #142 we are not getting to shifting a negative value.
I would keep this PR open for ~1 week to be sure the problem is fixed and then close it.

jbachorik:
#142 fixed the underlying cause. Closing this PR.

PR #145, opened 2024-10-29
Downport stack adjustment mode from async-profiler

jbachorik:
**What does this PR do?**:
Downport stack adjustment mode from async-profiler

**Motivation**:
Allow an experimental mode for recovery of 'unknown_Java` stacks. This mode is not enabled by default because it does 'blind' probing, incrementing SP by pointer size at a time, hoping ASGCT would be able to unwind from that location - which is not really safe and may induce crashes.

**For Datadog employees**:


jbachorik:
Closing in favour of #147

PR #156, opened 2024-12-05
Github action fix - Caching

r1viollet:
Test caching the download of zing

**What does this PR do?**:

**Motivation**:

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:
- [ ] JIRA: [JIRA-XXXX]


.github/workflows/test_workflow.yml
```diff
@@ -565,27 +565,35 @@ jobs:
         with:
           distribution: 'temurin'
           java-version: "11"
+      - name: Cache JDK Downloads
+        uses: actions/cache@v3
+        with:
+          path: /home/runner/jvm
+          key: ${{ runner.os }}-zing-jdk-${{ matrix.java_version }}-v0
```
r1viollet, line 572: @jbachorik any reason not to add caching to the JDK downloads ?
I would need to adjust this to all the jobs

jbachorik: Caching the jdk binaries would be useful. We would key on the distro/version/arch but that should speed up builds significantly.

jbachorik:
It is a dup of #159 and #160 - I think you can safely close it now.

PR #164, opened 2025-01-07
Add safeguard for J9-specific jmethodid unloading behaviour (version-dependent)

MattAlp:
**What does this PR do?**:
Adds a safeguard to prevent crashes on certain versions of J9 for which we don't have certain API compliance guarantees assumed to be present by async-profiler / java-profiler.

**Motivation**:
Incident 32141

**Additional Notes**:
See incident channel and private Slack for further details, (private) customer info and correspondence informed this change.

**How to test the change?**:

**For Datadog employees**:
- [x] JIRA: [PROF-11094]

[PROF-11094]: https://datadoghq.atlassian.net/browse/PROF-11094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -150,7 +150,7 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
     bool entry = false;
     if (VMMethod::check_jmethodID(method) &&
         jvmti->GetMethodDeclaringClass(method, &method_class) == 0 &&
-        jvmti->GetClassSignature(method_class, &class_name, NULL) == 0 &&
+        ((!VM::isOpenJ9() || method_class != reinterpret_cast<jclass>(-1)) && jvmti->GetClassSignature(method_class, &class_name, NULL) == 0) &&
         jvmti->GetMethodName(method, &method_name, &method_sig, NULL) == 0) {
 
       if (first_time) {
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -150,6 +150,13 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
     bool entry = false;
     if (VMMethod::check_jmethodID(method) &&
         jvmti->GetMethodDeclaringClass(method, &method_class) == 0 &&
+        // On some older versions of J9, the JVMTI call to GetMethodDeclaringClass will return OK = 0, but when a
+        // classloader is unloaded they free all JNIIDs. This means that anyone holding on to a jmethodID is
+        // pointing to corrupt data and the behaviour is undefined.
+        // The behaviour is adjusted so that when asgct() is used or if `-XX:+KeepJNIIDs` is specified,
+        // when a classloader is unloaded, the jmethodIDs are not freed, but instead marked as -1.
+        // The nested check below is to mitigate these crashes.
+        // In more recent versions, the condition above will short-circuit safely.
         ((!VM::isOpenJ9() || method_class != reinterpret_cast<jclass>(-1)) && jvmti->GetClassSignature(method_class, &class_name, NULL) == 0) &&
         jvmti->GetMethodName(method, &method_name, &method_sig, NULL) == 0) {
 


r1viollet:
Thanks for looking into this.
Is it possible to add a test for this behaviour ?

jbachorik:
> Is it possible to add a test for this behaviour ?

This would probably be almost impossible. We are getting the wrong method ID from a JVMTI call. Hence, we would need to stub the JNI call with some complex patching just for the duration of the test case (patching via LD_PRELOAD, if at all possible, would be too wide and we might fail way before we actually get to the code we want to test).

One thing that is fishy is that `jvmti->GetMethodDeclaringClass(method, &method_class)` is returning -1 in method_class but not returning an error code. By the spec if this function is not able to properly handle the provided methodid it should return `[JVMTI_ERROR_INVALID_METHODID](https://docs.oracle.com/javase/7/docs/platform/jvmti/jvmti.html#JVMTI_ERROR_INVALID_METHODID)`

jbachorik (approved):
Just one thing - could you add a comment about the J9 using -1 for unloaded methods so it is visible that this is actually a bug in J9 JVMTI implementation where they fail to return error code for unloaded method but rather segfault?

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -150,7 +150,14 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
...
         jvmti->GetMethodDeclaringClass(method, &method_class) == 0 &&
-        jvmti->GetClassSignature(method_class, &class_name, NULL) == 0 &&
+        // On some older versions of J9, the JVMTI call to GetMethodDeclaringClass will return OK = 0, but when a
+        // classloader is unloaded they free all JNIIDs. This means that anyone holding on to a jmethodID is
+        // pointing to corrupt data and the behaviour is undefined.
+        // The behaviour is adjusted so that when asgct() is used or if `-XX:+KeepJNIIDs` is specified,
+        // when a classloader is unloaded, the jmethodIDs are not freed, but instead marked as -1.
+        // The nested check below is to mitigate these crashes.
+        // In more recent versions, the condition above will short-circuit safely.
+        ((!VM::isOpenJ9() || method_class != reinterpret_cast<jclass>(-1)) && jvmti->GetClassSignature(method_class, &class_name, NULL) == 0) &&
```
r1viollet, line 160: minor / style: we could have declared the -1 value as a constexpr value (`k_invalid_j9_method_class` ?)
Is checking the -1 value not enough (to avoid the isOpenJ9 call) ?

r1viollet:
> > This would probably be almost impossible. We are getting the wrong method ID from a JVMTI call. Hence, we would need to stub the JNI call with some complex patching just for the duration of the test case (patching via LD_PRELOAD, if at all possible, would be too wide and we might fail way before we actually get to the code we want to test).
> 
> Agreed. The surface area of stubbing and testing something like this would quickly require finicky setup and/or stubbing out _a lot_ of JVMTI and/or core JVM functionality.

Capturing JVM invariants per distributions seems like something very interesting.
I was not particularly precise on what type of test I was referring to. Do you think a unit test mocking the interactions would be difficult to write ?

ddprof-lib/src/main/cpp/flightRecorder.cpp
jbachorik: Hm, now when rereading the comment, this does not make much sense.
The info from J9 engineers is about invalid jmethodid becoming `-1` - but that's not what we were seeing, right? It was the `jclass` returned from `GetMethodDeclaringClass` on such invalid jmethodid  that was `-1`. 
I mean, the workaround would work but the root cause seems to be J9 not handling the invalid jmethodids consistently in the JVMTI calls itself (eg. `GetMethodDeclaringClass` when called on an invalid jmethodid should return an error code instead of putting -1 to a jclass variable 🤷

jbachorik:
> Do you think a unit test mocking the interactions would be difficult to write ?

The problem is that it is the JVMTI function itself which behaves not according to the spec. I mean, theoretically we could create tests for spec conformance of the JVMTI implementation per JDK distribution but that would be no small task (not considering the cost of running and maintaining the suite)

ddprof-lib/src/main/cpp/flightRecorder.cpp
r1viollet: 1. As a general guidance we would want to avoid reinterpret casts when possible. The constexpr part is just a good habit to take (though sometimes hard to apply if we use old C++ versions)
2. The j9 call is more a question of taste and obviously very minor.

PR #167, opened 2025-01-09
Resolve native thread ID through SafeAccess

jbachorik:
**What does this PR do?**:
It changes the way the native thread id is resolved in the JVMTI based wallclock sampler.

**Motivation**:
The thread may disappear between having been selected and sampled, or even during preparation to sample.
Since we need to access internal VM structure to retrieve the native thread ID, it may lead to segfault.
Hence, wrapping in SafeAccess will make this, well, safe.

**For Datadog employees**:
- [x] JIRA: [PROF-11103]


jbachorik:
Closed in favour of #168

PR #168, opened 2025-01-14
Properly manage JNI resources in JVMTI wallclock sampler

jbachorik:
**What does this PR do?**:
It fixes the issue with the JVMTI wallclock sampler keep all threads in the system alive indefinitely

**Motivation**:
Resolve the corresponding memory leak. Make JVMTI wallclock sampler usable.

**Additional Notes**:
Due to stricter checks for thread state before collecting stacktrace and proper resource management, the #167 is not necessary any more.

**How to test the change?**:

**For Datadog employees**:
- [x] JIRA: [PROF-11103]

### DIFF
--- a/ddprof-lib/src/main/cpp/wallClock.cpp
+++ b/ddprof-lib/src/main/cpp/wallClock.cpp
@@ -161,48 +161,82 @@ void WallClockJVMTI::timerLoop() {
   // Attach to JVM as the first step
     VM::attachThread("Datadog Profiler Wallclock Sampler");
     auto collectThreads = [&](std::vector<ThreadEntry>& threads) {
-        jvmtiEnv* jvmti = VM::jvmti();
-        if (jvmti == nullptr) {
-            return;
-        }
-        JNIEnv* jni = VM::jni();
+      jvmtiEnv* jvmti = VM::jvmti();
+      if (jvmti == nullptr) {
+          return;
+      }
+      JNIEnv* jni = VM::jni();
 
-        jint threads_count = 0;
-        jthread* threads_ptr = nullptr;
-        jvmti->GetAllThreads(&threads_count, &threads_ptr);
+      jint threads_count = 0;
+      jthread* threads_ptr = nullptr;
+      jvmti->GetAllThreads(&threads_count, &threads_ptr);
 
-        bool do_filter = Profiler::instance()->threadFilter()->enabled();
-        int self = OS::threadId();
-
-        for (int i = 0; i < threads_count; i++) {
-          jthread thread = threads_ptr[i];
-          if (thread != nullptr) {
-            VMThread* nThread = VMThread::fromJavaThread(jni, thread);
-            if (nThread == nullptr) {
-              continue;
-            }
-            int tid = nThread->osThreadId();
-            if (tid != self && (!do_filter || Profiler::instance()->threadFilter()->accept(tid))) {
-              threads.push_back({nThread, thread});
-            }
+      bool do_filter = Profiler::instance()->threadFilter()->enabled();
+      int self = OS::threadId();
+      for (int i = 0; i < threads_count; i++) {
+        jthread thread = threads_ptr[i];
+        if (thread != nullptr) {
+          VMThread* nThread = VMThread::fromJavaThread(jni, thread);
+          if (nThread == nullptr) {
+            jni->DeleteLocalRef(thread);
+            continue;
           }
+          jint thread_state;
+          if (jvmti->GetThreadState(thread, &thread_state) == JVMTI_ERROR_NONE &&
+              (thread_state & JVMTI_THREAD_STATE_TERMINATED) == 0) {
+              int tid = VMThread::nativeThreadId(jni, thread);
+              if (tid != -1 && tid != self && (!do_filter || Profiler::instance()->threadFilter()->accept(tid))) {
+                  threads.push_back({nThread, jni->NewWeakGlobalRef(thread)});
+              }
+          }
+          jni->DeleteLocalRef(thread);
         }
-        jvmti->Deallocate((unsigned char*)threads_ptr);
+      }
+      jvmti->Deallocate((unsigned char*)threads_ptr);
     };
 
   auto sampleThreads = [&](ThreadEntry& thread_entry, int& num_failures, int& threads_already_exited, int& permission_denied) {
     static jint max_stack_depth = (jint)Profiler::instance()->max_stack_depth();
     static jvmtiFrameInfo* frame_buffer = new jvmtiFrameInfo[max_stack_depth];
     static jvmtiEnv* jvmti = VM::jvmti();
+    static JNIEnv* jni = VM::jni();
 
     int num_frames = 0;
-    jvmtiError err = jvmti->GetStackTrace(thread_entry.java, 0, max_stack_depth, frame_buffer, &num_frames);
+    if (thread_entry.java_ref == nullptr) {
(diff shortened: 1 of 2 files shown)


r1viollet:
Thanks for figuring things out! I'll try to run things to see how it behaves.
Is there a way we could generate a smoke test that would help identify this ? Or perhaps in a benchmarking way ?

ddprof-lib/src/main/cpp/wallClock.cpp
```diff
@@ -161,49 +161,83 @@ void WallClockJVMTI::timerLoop() {
...
+      if (jvmti == nullptr) {
+          return;
+      }
+      JNIEnv* jni = VM::jni();
 
-        bool do_filter = Profiler::instance()->threadFilter()->enabled();
-        int self = OS::threadId();
+      jint threads_count = 0;
+      jthread* threads_ptr = nullptr;
+      jvmti->GetAllThreads(&threads_count, &threads_ptr);
```
r1viollet, line 172: can you comment a little on this part.
So we are getting jthread references, do we eventually always want to remove the local ref ? Or are there cases where we want to keep the local ref ?
Could some kind of RAII design, ensure that when we leave the function, we just delete all local refs to the jthreads ? (avoiding some of the branches bellow?)

ddprof-lib/src/main/cpp/wallClock.cpp
```diff
@@ -161,49 +161,83 @@ void WallClockJVMTI::timerLoop() {
...
+          if (nThread == nullptr) {
+            jni->DeleteLocalRef(thread);
+            continue;
           }
+          jint thread_state;
+          if (jvmti->GetThreadState(thread, &thread_state) == JVMTI_ERROR_NONE &&
+              (thread_state & JVMTI_THREAD_STATE_TERMINATED) == 0) {
+              int tid = VMThread::nativeThreadId(jni, thread);
+              if (tid != -1 && tid != self && (!do_filter || Profiler::instance()->threadFilter()->accept(tid))) {
+                  threads.push_back({nThread, jni->NewWeakGlobalRef(thread)});
```
r1viollet, line 189: I'm assuming NewWeakGlobalRef can fail ?

ddprof-lib/src/main/cpp/wallClock.cpp
```diff
@@ -161,49 +161,83 @@ void WallClockJVMTI::timerLoop() {
...
+          VMThread* nThread = VMThread::fromJavaThread(jni, thread);
+          if (nThread == nullptr) {
+            jni->DeleteLocalRef(thread);
+            continue;
           }
+          jint thread_state;
+          if (jvmti->GetThreadState(thread, &thread_state) == JVMTI_ERROR_NONE &&
+              (thread_state & JVMTI_THREAD_STATE_TERMINATED) == 0) {
+              int tid = VMThread::nativeThreadId(jni, thread);
+              if (tid != -1 && tid != self && (!do_filter || Profiler::instance()->threadFilter()->accept(tid))) {
```
r1viollet, line 188: why would tid be -1 ? should we add a comment ?

ddprof-lib/src/main/cpp/wallClock.cpp
r1viollet: failure probably means OOM, or could it fail in some kind of locked state ?

ddprof-lib/src/main/cpp/wallClock.cpp
```diff
@@ -161,49 +161,83 @@ void WallClockJVMTI::timerLoop() {
...
-    jvmtiError err = jvmti->GetStackTrace(thread_entry.java, 0, max_stack_depth, frame_buffer, &num_frames);
+    if (thread_entry.java_ref == nullptr) {
+      num_failures++;
+      return false;
+    }
+
+    jobject thread = jni->NewLocalRef(thread_entry.java_ref);
+    if (thread == nullptr) {
+        num_failures++;
+        jni->DeleteWeakGlobalRef(thread_entry.java_ref);
```
r1viollet, line 213: this can probably be easier to manage with some RAII design (like defer deletion when something goes out of scope)

ddprof-lib/src/main/cpp/wallClock.cpp
```diff
@@ -161,49 +161,83 @@ void WallClockJVMTI::timerLoop() {
...
+
     if (err != JVMTI_ERROR_NONE) {
       num_failures++;
       if (err == JVMTI_ERROR_THREAD_NOT_ALIVE) {
         threads_already_exited++;
       }
       return false;
     }
+    if (num_frames == 0) {
+      // some JVMTI attached threads are Java-like but have no stack; we can just ignore them
```
r1viollet, line 238: interesting, I wonder if we should represent these in the future (not relevant for this PR)

ddprof-lib/src/main/cpp/wallClock.h
```diff
@@ -76,38 +76,43 @@ class BaseWallClock : public Engine {
...
-        threads.clear();
         // Get a random sleep duration
-        // clamp the random interval to <1,2N-1>
-        // the probability of clamping is extremely small, close to zero
-        OS::sleep(std::min(std::max((long int)1, static_cast<long int>(distribution(generator))), ((_interval * 2) - 1)));
+        // restrict the random interval to <N/2,2N-1>
+        long int delay = _interval;
+        do {
+          delay = static_cast<long int>(distribution(generator));
+        } while (delay < interval / 2 || delay > 2 * interval);
```
r1viollet, line 114: Could std::uniform_int_distribution not directly fit the interval?

jbachorik: Yes, it could - we would switch from normal distribution to a uniform distribution, though.

ddprof-lib/src/main/cpp/wallClock.cpp
jbachorik: Our backend is showing such samples as a single 'Unknown Java' frame - I found it very unhelpful for the wallclock profile, but if we can find how it can be used it is easy to add this back.

ddprof-lib/src/main/cpp/wallClock.cpp
jbachorik: Yes, it can fail. In that case we will enqueue `NULL` and the dequeing logic will just skip it.

ddprof-lib/src/main/cpp/wallClock.h
r1viollet: I did not think too hard about what type of distribution we are looking for. Though are we not biasing the normal distribution by cutting off parts of it ?
I guess having the values centered around the interval does make sense. 
Either way this is probably minor.

ddprof-lib/src/main/cpp/wallClock.cpp
jbachorik: Added a comment

ddprof-lib/src/main/cpp/wallClock.cpp
jbachorik: I have added RAII objects to make this easier to grasp

ddprof-lib/src/main/cpp/wallClock.cpp
jbachorik: Added an extra comment

jbachorik:
I have found some weird behaviour in JVMTI, especially when I try using RAII objects. 
I need to investigate first - turning the PR to draft in the meantime.

PR #195, opened 2025-03-20
Add profiler counters for time spent in stack unwinding

bric3:
**What does this PR do?**:

Use TSC time counters for unwinding.

**How to test the change?**:

**For Datadog employees**:
- [x] JIRA: PROF-11368


jbachorik:
HI @bric3 - you picked the right approach with the counters.
But it will be better if you have only one counter for the whole unwinding - keeping track of each unwound frame type would be too expensive.

jbachorik:
Hm, sorry for not being clear from the beginning - I meant having just one measurement around the unwinding loop in each unwinder implementation (`walkFP`, `walkDwarf` and `walkVM`).
There will always be just one implementation active in the profiler so we will not be mixing up statistics from multiple implementations. And by having just one start/stop timestamp pair we will save quite a nice number of nanoseconds per unwound stack trace.

PR #205, opened 2025-04-14
Prepare infrastructure to gradually absorb the upstream sourcetree

jbachorik:
**What does this PR do?**:
Prepares an infra where we may gradually start using the upstream version of async-profiler.

However, we need to have a ‘locked’ version of dd specific fork async-profiler such that we are not forced to live on the bleeding edge but can move from release to release while carefully cherry-picking any crucial fixes in between.

The idea is to pull the async-profiler upstream via ‘git clone’ (kind of very simple subrepo) and then copy source files to a special location where they will be picked up by our build system instead the ones directly in our source tree. This will allow us to gradually either directly replace the source files with their upstream counterparts or wrap them in a facade or adapter before we get all the necessary changes upstreamed.

If we need to 'wrap' the AP classes, we do so in a namespaced version such that we avoid unexpected clashes. You can check how it works with `vmstructs_dd.[h|cpp]` and `stackWalker_dd.h`. Also, the filenames are suffixed by `_dd` to avoid confusion when eg. including the header files.

**How to work with external async-profiler repo**
The idea is to keep a base branch for our work in [Datadog async-profiler fork](https://github.com/DataDog/async-profiler). This branch is named `dd/master` and is currently based on latest release plus a few useful commits.

To go forward, we will be cherry-picking interesting commits from the upstream master when necessary, but will try to keep the number of cherr-picks to minimum. Once a new release of the upstream is available we will rebase the `dd/master` on that release.

This will give us a good stability while still allowing to pull in critical changes without waiting for the upstream release.

**How to test the change?**:
Run current full test suite

**For Datadog employees**:
- [x] JIRA: [PROF-11648]

[PROF-11648]: https://datadoghq.atlassian.net/browse/PROF-11648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


r1viollet:
I wonder if we could namespace most of the code to ensure we never end up in a weird state where we are unsure of the code being linked. Something like `namespace ap` or `namespace dd` if we choose to namespace the Datadog parts.

jbachorik:
Yes, namespacing sounds useful. However, I would prefer adding namespace only in case we need to create a facade over the async-profiler classes, otherwise we would have to namespace all our classes, just in case. And that feels like a pretty big change for not a lot of gain.

ddprof-lib/src/main/cpp/stackWalker_dd.h
```diff
@@ -0,0 +1,72 @@
...
+ * limitations under the License.
+ */
+
+#ifndef _STACKWALKER_DD_H
+#define _STACKWALKER_DD_H
+
+#include "stackWalker.h"
+
+
+namespace ddprof {
```
r1viollet, line 23: NIT: I am not used to namespace adding indents (though I can adjust)

.gitignore
```diff
@@ -17,6 +17,7 @@
 .tmp
 *.iml
 /ddprof-stresstest/jmh-result.*
+/ddprof-lib/src/main/cpp-external/**/*
```
r1viollet, line 20: good thinking

r1viollet (approved):
LGTM, thank you!

PR #208, opened 2025-04-28
Fix #200 Crash related to aligned_alloc and free in context

### LINKED ISSUE #200: Crash related to aligned_alloc and free in context
Hi, 

I encountered a crash today, from the hs_err log I see the last java frame is `getContextPageOffset0`.

I then used the debug jar and enabled core dump, the stack trace is bellow:

```
#0  0x00007fc775409387 in raise () from /lib64/libc.so.6
#1  0x00007fc77540aa78 in abort () from /lib64/libc.so.6
#2  0x00007fc776008a66 in tcmalloc::Log(tcmalloc::LogMode, char const*, int, tcmalloc::LogItem, tcmalloc::LogItem, tcmalloc::LogItem, tcmalloc::LogItem) () from /usr/lib64/libtcmalloc.so
#3  0x00007fc775ffd644 in (anonymous namespace)::InvalidFree(void*) () from /usr/lib64/libtcmalloc.so
#4  0x00007fc74eb9f701 in Contexts::getPage(int) ()
```

The reason is, if `libtcmalloc.so.4.4.5` is PR_LOADEDed, the `aligned_alloc` is from libc (because libtcmalloc.so.4.4.5 does not have aligned_alloc), while `free` is from tcmalloc, then tcmalloc complains about pointer is not allocated by it.

```
readelf -s -W /usr/lib64/libc-2.26.so|grep aligned_alloc
  1261: 000000000007f0e0     9 FUNC    WEAK   DEFAULT   12 aligned_alloc@@GLIBC_2.16
  6011: 000000000007f0e0     9 FUNC    WEAK   DEFAULT   12 aligned_alloc

readelf -s -W /usr/lib64/libtcmalloc.so.4.4.5|grep aligned_alloc|wc -l
0
```

https://github.com/DataDog/java-profiler/blob/6fa1618db349d42162539e0481fc5337170a53f2/ddprof-lib/src/main/cpp/context.cpp#L57-L68

Consider users are free to use any allocator, it's better not to crash. The simplest fix is to replace `aligned_alloc` with `malloc`. 
Any thoughts ?

Thanks.

yanglong1010:
Hi, Could I have a review of this patch ?

**What does this PR do?**:
Fix #200 

**Motivation**:
Replace `aligned_alloc` with `posix_memalign`

**Additional Notes**:
If `posix_memalign` can not alloc memory and returns a none-zero, we should avoid subsequent memory access, so return an empty `ContextPage` and the Java code check for this.

**How to test the change?**:
The following tests are passed:
- gradle test
- gradle testRelease

**For Datadog employees**:
- [ ] JIRA: [JIRA-XXXX]

### DIFF
--- a/ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
+++ b/ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
@@ -240,6 +240,9 @@ public final class JavaProfiler {
             return;
         }
         long pageOffset = getPageUnsafe(tid);
+        if (pageOffset == 0) {
+            return;
+        }
         int index = (tid % PAGE_SIZE) * CONTEXT_SIZE;
         long base = pageOffset + index;
         UNSAFE.putLong(base + SPAN_OFFSET, spanId);
@@ -252,20 +255,24 @@ public final class JavaProfiler {
             return;
         }
         ByteBuffer page = getPage(tid);
+        if (page == null) {
+            return;
+        }
         int index = (tid % PAGE_SIZE) * CONTEXT_SIZE;
         page.putLong(index + SPAN_OFFSET, spanId);
         page.putLong(index + ROOT_SPAN_OFFSET, rootSpanId);
         page.putLong(index + CHECKSUM_OFFSET, spanId ^ rootSpanId);
     }
 
-
-
     private ByteBuffer getPage(int tid) {
         int pageIndex = tid / PAGE_SIZE;
         ByteBuffer page = contextStorage[pageIndex];
         if (page == null) {
             // the underlying page allocation is atomic so we don't care which view we have over it
-            contextStorage[pageIndex] = page = getContextPage0(tid).order(ByteOrder.LITTLE_ENDIAN);
+            ByteBuffer buffer = getContextPage0(tid);
+            if (buffer != null) {
+                contextStorage[pageIndex] = page = buffer.order(ByteOrder.LITTLE_ENDIAN);
+            }
         }
         return page;
     }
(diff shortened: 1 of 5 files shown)


ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
```diff
@@ -252,20 +255,24 @@ private void setContextByteBuffer(int tid, long spanId, long rootSpanId) {
...
     private ByteBuffer getPage(int tid) {
         int pageIndex = tid / PAGE_SIZE;
         ByteBuffer page = contextStorage[pageIndex];
         if (page == null) {
             // the underlying page allocation is atomic so we don't care which view we have over it
-            contextStorage[pageIndex] = page = getContextPage0(tid).order(ByteOrder.LITTLE_ENDIAN);
+            ByteBuffer buffer = getContextPage0(tid);
+            if (buffer != null) {
+                contextStorage[pageIndex] = page = buffer.order(ByteOrder.LITTLE_ENDIAN);
+            }
```
jbachorik, line 275: This will cause reattempting to allocate even though we are failing eg. due to memory shortage.
Would it make sense to use a sentinel ByteBuffer instance to signal that a page is not allocated but allocation was attempted?

jbachorik (changes requested):
Thanks for the PR!
Mostly, I am fine with the changes - just one small thing related to re-attempted page initialization, as described in the comment.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
yanglong1010: Pushed

jbachorik (approved):

r1viollet:
Thank you for the change!

PR #210, opened 2025-04-28
Explicitly initialize empty context page

jbachorik:
(followup to #308)

**What does this PR do?**:
It fixes compilation error on an old g++

**Motivation**:
We need to build on some rather old systems for compatibility sakes and it requires explicit initialization of the struct


MattAlp (approved):
LGTM. For external context- this is necessary for our `glibc 2.17` builds.

PR #213, opened 2025-05-20
Properly handle the adaptive sampling interval overflow

jbachorik:
**What does this PR do?**:
It adds safety handling for possible sampling interval overflows

**Motivation**:
The adaptive sampling interval can overflow int range when a very allocation heavy activity is present - and wrap back to the default value which is terrible for the profile size.
It is much better to just clamp on INT_MAX - yes, we will miss the rate target but not that brutally as when wrapping around to the default interval.

**For Datadog employees**:
- [x] JIRA: [PROF-11801]

[PROF-11801]: https://datadoghq.atlassian.net/browse/PROF-11801?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/objectSampler.cpp
+++ b/ddprof-lib/src/main/cpp/objectSampler.cpp
@@ -26,6 +26,7 @@
 #include "thread.h"
 #include "vmStructs.h"
 #include <jni.h>
+#include <limits.h>
 #include <math.h>
 #include <string.h>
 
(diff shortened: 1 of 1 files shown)


ddprof-lib/src/main/cpp/objectSampler.cpp
```diff
@@ -188,14 +189,15 @@ Error ObjectSampler::updateConfiguration(u64 events, double time_coefficient) {
       CONFIG_UPDATE_CHECK_PERIOD_SECS, 15);
 
   float signal = pid_controller.compute(events, time_coefficient);
-  int required_interval = _interval - static_cast<int>(signal);
+  u64 required_interval = static_cast<u64>(_interval) - static_cast<int>(signal);
```
r1viollet, line 192: This is probably not OK if `interval < signal`.
You will have both sides converted to a common type, then the subtraction will wrap around (UB).

jbachorik: Ok. I just clamped the `required_interval` in case `_interval` is smaller than `signal` to avoid wrap-around.

ddprof-lib/src/main/cpp/objectSampler.cpp
```diff
@@ -188,14 +189,16 @@ Error ObjectSampler::updateConfiguration(u64 events, double time_coefficient) {
       CONFIG_UPDATE_CHECK_PERIOD_SECS, 15);
 
   float signal = pid_controller.compute(events, time_coefficient);
```
r1viollet, line 191: minor, there is already a conversion here from double to float. Then a truncation to an int.

r1viollet (approved):
LGTM
Thanks for thinking carefully about this!

PR #214, opened 2025-05-21
Re-connect crash recursion protection with VM stackwalker

jbachorik:
**What does this PR do?**:
It fixes an omission from the migration to the vanilla upstream stackwalker implementations.
Our stackwalker was providing a callback to reset the crash handler recursion protection when it needed to longjump to recover. Without that, it is easier to get into an inconsistent state when sooner or later the segfaults won't get handled because the protection thinks we are recursing too deeply.

**Motivation**:
Fix up the omission from the migration. Fix the problems we are seeing in some services.

**Additional Notes**:

**How to test the change?**:
Testing is rather cumbersome. We would have to simulate segfault in the vm stackwalking code and make sure we are properly resetting the tracking.
I did this manually by patching `VMMethod:id()` to dereference a nullptr. Without this patch I will get an immediate crash - with this patch everything works as expected.

**For Datadog employees**:
- [x] JIRA: [PROF-11839]

[PROF-11839]: https://datadoghq.atlassian.net/browse/PROF-11839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/build.gradle
+++ b/ddprof-lib/build.gradle
@@ -143,6 +143,9 @@ def copyUpstreamFiles = tasks.register('copyUpstreamFiles', Copy) {
   configure {
     dependsOn cloneAPTask
   }
+  onlyIf {
+    !project.hasProperty("debug-ap")
+  }
   description = 'Copy shared upstream files'
   from("${projectDir}/build/async-profiler/src") {
     include "arch.h"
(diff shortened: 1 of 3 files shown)


r1viollet (approved):
LGTM

MattAlp (approved):

PR #218, opened 2025-05-28
Patch upstream stackWalker.cpp not to fail on unaligned access

jbachorik:
**Motivation**:
Nightly sanitized tests are failing consistently for  Java 21 JVMCI on aarch64 due to unaligned access in vm stackwalking.
This is adding a patch to suppress asan address check in `walkVM` function.


zhengyu123 (approved):
Looks good.

PR #223, opened 2025-06-10
Remap thread id to avoid bitmap contention

zhengyu123:
**What does this PR do?**:

An application starts a few threads in short period of windows, the newly created threads are likely to have consecutive or close thread ids. When  `ThreadFilter` maps them to the bits on bitmap, there is high chance that they may be mapped to a single cache line, that results cache line contention when updating the bitmap concurrently.

The PR purposes to remap thread id  by swapping lower 2 bytes of thread id, so that consecutive/close thread ids map to different cache lines.

**Motivation**:

Reduce cache line contention, improve performance.

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:
- [ ] JIRA: [JIRA-XXXX]

### DIFF
--- a/ddprof-lib/src/main/cpp/threadFilter.cpp
+++ b/ddprof-lib/src/main/cpp/threadFilter.cpp
@@ -17,6 +17,7 @@
 #include "threadFilter.h"
 #include "counters.h"
 #include "os.h"
+#include <cassert>
 #include <stdlib.h>
 #include <string.h>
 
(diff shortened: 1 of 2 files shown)


r1viollet:
Would the contention not remain if you only swap lower bits ?
I also thought of swapping all bits, however that would lead to other issues (fragmentation in the radix tree). Happy to talk this through. I'll bench it to see what the results have to say.

ddprof-lib/src/main/cpp/threadFilter.cpp
```diff
@@ -85,12 +86,23 @@ void ThreadFilter::clear() {
   _size = 0;
 }
 
+int ThreadFilter::mapThreadId(int thread_id) {
```
jbachorik, line 89: Can you add a comment that this function is an involution ( f(f(x)) == x ) ?

jbachorik (approved):
How do the benchmarks look with the latest changes?
Otherwise, LGTM!

zhengyu123:
JFR recordings from Erwan's benchmark:

Basline:
before

After patch:
after

jbachorik:
Is this superseded by #229  ?
If yes, we should close this one.

PR #229, opened 2025-06-17
Remap thread id to avoid bitmap contention

zhengyu123:
**What does this PR do?**:

An application starts a few threads in short period of window, the newly created threads are likely to have consecutive or close thread ids. When  `ThreadFilter` maps them to the bits on bitmap, there are high chances that they may be mapped to a single cache line, that results cache line contention when updating the bitmap concurrently.

The PR purposes to remap thread id  by reversing lower 2 bytes of thread id, so that consecutive/close thread ids map to different cache lines.

The reason for only reversing lower 2 bytes, is that, thread id is remapped within the same bitmap.

**Motivation**:

Reduce cache line contention, improve performance.

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:
- [x] JIRA: [PROF-12002](https://datadoghq.atlassian.net/browse/PROF-12002)

[PROF-12002]: https://datadoghq.atlassian.net/browse/PROF-12002?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/reverse_bits.h
+++ b/ddprof-lib/src/main/cpp/reverse_bits.h
@@ -1,0 +1,23 @@
+//
+// Borrow the implementation from openjdk
+// https://github.com/openjdk/jdk/blob/master/src/hotspot/share/utilities/reverse_bits.hpp
+//
+
+#ifndef REVERSE_BITS_H
+#define REVERSE_BITS_H
+#include "arch_dd.h"
+#include <stdint.h>
+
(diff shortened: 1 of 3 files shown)


jbachorik (approved):

r1viollet (approved):
LGTM

PR #238, opened 2025-07-07
Thread filter optim

r1viollet:
**What does this PR do?**:

- Reserve padded slots
- Introduce a register / unregister to retrieve slots
- manage a free list

**Motivation**:

Improve throughput of applications that run on many threads with many context updates.

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:
- [ ] JIRA: [JIRA-XXXX]


ddprof-lib/src/main/cpp/threadFilter.cpp
```diff
@@ -1,175 +1,271 @@
...
+ThreadFilter::SlotID ThreadFilter::registerThread() {
+    // First, try to get a slot from the free list (lock-free stack)
+    SlotID reused_slot = popFromFreeList();
+    if (reused_slot >= 0) {
+        return reused_slot;
+    }
+
+    // Allocate a new slot
+    SlotID index = _next_index.fetch_add(1, std::memory_order_relaxed);
+    if (index >= kMaxThreads) {
```
zhengyu123, line 66: Not sure if it is important, but it can race `unregisterThread`, you may want to check return value of `fetch_sub` to determinate if it is really full.

ddprof-lib/src/main/cpp/threadFilter.cpp
```diff
@@ -1,175 +1,271 @@
...
+        return -1;
     }
 
-    filter = end + 1;
-  } while (*end);
+    const int chunk_idx = index >> kChunkShift;
 
-  _enabled = true;
+    // Ensure the chunk is initialized (lock-free)
+    if (chunk_idx >= _num_chunks.load(std::memory_order_acquire)) {
```
zhengyu123, line 75: I don't quite understand, are `index` and `chunk_idx` 1-to-1 matched?

ddprof-lib/src/main/cpp/threadFilter.cpp
r1viollet: I don't think I would be able to do anything different even if we have another thread that adds something concurrently

ddprof-lib/src/main/cpp/threadFilter.cpp
r1viollet: no, chunks are really groups of indices that will grow if we have a large amount of threads. The idea is to have roughly chunks of 256 threads and to grow if needed

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -123,19 +124,101 @@ Java_com_datadoghq_profiler_JavaProfiler_getSamples(JNIEnv *env,
...
+    return;  // Failed to register thread
+  }
+  thread_filter->add(tid, slot_id);
+}
+
+extern "C" DLLEXPORT void JNICALL
+Java_com_datadoghq_profiler_JavaProfiler_filterThreadRemove0(JNIEnv *env,
+                                                             jobject unused) {
+  ProfiledThread *current = ProfiledThread::current();
+  if (unlikely(current == nullptr)) {
```
zhengyu123, line 162: I think `assert(current != nllptr)` should be sufficient, otherwise, we have bigger problem.

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -123,19 +124,101 @@ Java_com_datadoghq_profiler_JavaProfiler_getSamples(JNIEnv *env,
...
+    assert(false);
+    return;
+  }
+  int tid = current->tid();
```
zhengyu123, line 166: Is it possible?

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -123,19 +124,101 @@ Java_com_datadoghq_profiler_JavaProfiler_getSamples(JNIEnv *env,
...
                                                        jboolean enable) {
-  int tid = ProfiledThread::currentTid();
-  if (tid < 0) {
+  ProfiledThread *current = ProfiledThread::current();
+  if (unlikely(current == nullptr)) {
+    assert(false);
+    return;
+  }
+  int tid = current->tid();
+  if (unlikely(tid < 0)) {
```
zhengyu123, line 194: Same as above

ddprof-lib/src/main/cpp/threadFilter.h
```diff
@@ -13,73 +13,80 @@
...
-
-  bool accept(int thread_id);
-  void add(int thread_id);
-  void remove(int thread_id);
-  u64* bitmapAddressFor(int thread_id);
-
-  void collect(std::vector<int> &v);
+private:
+    // Optimized slot structure with padding to avoid false sharing
+    struct alignas(64) Slot {
```
zhengyu123, line 55: We have definition of `DEFAULT_CACHE_LINE_SIZE` in `dd_arch.h`. I would suggest following code for readability.

```
  struct alignas(DEFAULT_CACHE_LINE_SIZE) Slot {
      std::atomic<int> value{-1};
      char padding[DEFAULT_CACHE_LINE_SIZE - sizeof(value)];
  };
```

ddprof-lib/src/main/cpp/threadFilter.cpp
```diff
@@ -13,163 +13,282 @@
...
     }
 
-    if (*end == '-') {
-      int to = strtol(end + 1, &end, 0);
-      while (id <= to) {
-        add(id++);
-      }
+    // Try to install it atomically
+    ChunkStorage* expected = nullptr;
+    if (_chunks[chunk_idx].compare_exchange_strong(expected, new_chunk, std::memory_order_acq_rel)) {
```
zhengyu123, line 82: `memory_order_release` should be sufficient.

ddprof-lib/src/main/cpp/threadFilter.cpp
```diff
@@ -13,163 +13,282 @@
...
     } else {
-      add(id);
+        // Another thread beat us to it - clean up our allocation
+        delete new_chunk;
+    }
+}
+
+ThreadFilter::SlotID ThreadFilter::registerThread() {
+    // If disabled, block new registrations
+    if (!_enabled.load(std::memory_order_acquire)) {
```
zhengyu123, line 92: I don't see any memory ordering `_enabled` providing. Could you explain what it `release`s

zhengyu123 (changes requested):
I did partial second round reviewing, I think there are many inconsistencies in memory ordering.

ddprof-lib/src/main/cpp/javaApi.cpp
r1viollet: I think this happens on unloading. JVMTI cleanup can be removed before all threads are finished ?

ddprof-lib/src/main/cpp/threadFilter.cpp
r1viollet: We want the init to be called (so that constructor is finished)
This means other threads might not have the init finished
Though we can always lazily register them later (even if the on thread start path is better)
So this acts as a load barrier. I think it makes sense. Feel free to challenge.

ddprof-lib/src/main/cpp/threadFilter.cpp
```diff
@@ -13,163 +13,282 @@
...
+    if (unlikely(!_enabled.load(std::memory_order_relaxed))) {
+        return true;
+    }
+    if (unlikely(slot_id < 0)) return false;
+
+    int chunk_idx = slot_id >> kChunkShift;
+    int slot_idx = slot_id & kChunkMask;
+
+    // Fast path: assume valid slot_id from registerThread()
+    ChunkStorage* chunk = _chunks[chunk_idx].load(std::memory_order_relaxed);
```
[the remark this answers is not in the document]
r1viollet: This was intentional, the idea is that we either have null, or a valid chunk. The impact of missing one trace is acceptable if we don't see the chunk in the current thread (which is pretty unlikely). wdyt ?

ddprof-lib/src/main/cpp/threadFilter.cpp
```diff
@@ -13,163 +13,282 @@
...
+    if (likely(chunk != nullptr)) {
+        return chunk->slots[slot_idx].value.load(std::memory_order_acquire) != -1;
```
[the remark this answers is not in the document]
r1viollet: I think this is a tradeoff. I hope this is a path where acquire is acceptable. If you think otherwise I can adjust.

ddprof-lib/src/main/cpp/threadFilter.cpp
```diff
@@ -13,163 +13,282 @@
...
+    }
+    for (int i = 0; i < kFreeListSize; ++i) {
+        _free_list[i].value.store(-1, std::memory_order_relaxed);
+        _free_list[i].next.store(-1, std::memory_order_relaxed);
+    }
+    // Publish 0 chunks to stop range scans (collect)
+    _num_chunks.store(0, std::memory_order_release);
+    // Detach and delete chunks
+    for (int i = 0; i < kMaxChunks; ++i) {
+        ChunkStorage* chunk = _chunks[i].exchange(nullptr, std::memory_order_acq_rel);
```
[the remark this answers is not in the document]
r1viollet: adjusting

ddprof-lib/src/main/cpp/threadFilter.cpp
r1viollet: agreed

zhengyu123 (changes requested):

zhengyu123 (approved):

jbachorik (approved):

PR #251, opened 2025-07-24
Remove gtest during build

r1viollet:
**What does this PR do?**:

Remove gtest during build
Clarify outputs for async-profiler check

**Motivation**:

Consecutive builds are faster

**Additional Notes**:

NA

**How to test the change?**:

We already call test commands which will run gtests

**For Datadog employees**:
- [ ] JIRA: [JIRA-XXXX]

### DIFF
--- a/ddprof-lib/build.gradle
+++ b/ddprof-lib/build.gradle
@@ -112,6 +112,9 @@ def createDebugLinkTask(config, linkTask, extractDebugTask) {
     }
     dependsOn extractDebugTask
     description = 'Add debug link to the original library'
+    
+    inputs.files linkTask, extractDebugTask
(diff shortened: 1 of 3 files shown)


jbachorik (commented):
Is it ok to run gtest only when testing Java?
You could then do linkDebug and get *.so file without running gtests. Is this the intention?

For CI I think this does make any difference - the gtests will run a bit later. For local dev, you can use `-Pskip-gtest` gradle paramater.

The rest of the changes looks good.

jbachorik:
> Though I agree this is a matter of taste.

It's mostly about whether we feel it is ok to be able get the library binary without going through the gtest by default.
I have no strong opinion in this case - as long as we run the gtests eventually :)

jbachorik (approved):

PR #257, opened 2025-08-05
Avoid calling profiler.stop() twice in ThreadFilterBenchmark

zhengyu123:
**What does this PR do?**:
Avoid  `IllegalStateException` exception due to double stop, which can result no output from jmh profiler.

**Motivation**:
Cleanup and make benchmark more robust.

**Additional Notes**:

**How to test the change?**:
Run the benchmark, `ThreadFilterBenchmark` no longer throws `IllegalStateException` exception.

**For Datadog employees**:
- [ ] JIRA: [PROF-12247](https://datadoghq.atlassian.net/browse/PROF-12247)

[PROF-12247]: https://datadoghq.atlassian.net/browse/PROF-12247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-stresstest/src/jmh/java/com/datadoghq/profiler/stresstest/scenarios/counters/ThreadFilterBenchmark.java
+++ b/ddprof-stresstest/src/jmh/java/com/datadoghq/profiler/stresstest/scenarios/counters/ThreadFilterBenchmark.java
@@ -103,13 +103,6 @@ public class ThreadFilterBenchmark extends Configuration {
             Thread.currentThread().interrupt();
         }
 
-        // Stop the profiler if it's active
-        try {
-            profiler.stop();
-        } catch (IllegalStateException e) {
-            System.out.println("Profiler was not active at teardown.");
-        }
-
         long endTime = System.currentTimeMillis();
         long totalOps = operationCount.get();
         double durationSecs = (endTime - startTime) / 1000.0;


ddprof-stresstest/src/jmh/java/com/datadoghq/profiler/stresstest/scenarios/counters/ThreadFilterBenchmark.java
```diff
@@ -103,13 +103,6 @@ public void tearDown() {
             Thread.currentThread().interrupt();
         }
 
-        // Stop the profiler if it's active
-        try {
```
r1viollet, line 107: better to stop in tear down or setup in case it is still running ?

r1viollet (approved):
I had both in case one was not called for some reason.
I'm not sure which makes more sense to remove.
I don't think the exception is a huge deal, though I'm fine removing either stops.

zhengyu123:
> I had both in case one was not called for some reason. I'm not sure which makes more sense to remove. I don't think the exception is a huge deal, though I'm fine removing either stops.

The exception prevents jmh profiler (e.g. prefasm) from generating output.

PR #280, opened 2025-10-20
Fix calltracestorage concurrent and reentrant issues

jbachorik:
**What does this PR do?**:
The PR is improving the robustness of the calltrace storage implementation such that it does not get tripped by eg. profiling signals coming in the middle of the code which makes sure the concurrent access is safe.

The detailed description is in the [arch doc](https://github.com/DataDog/java-profiler/blob/jb/trace_storage_fix/docs/architecture/CallTraceStorage.md)

**Motivation**:
The memory corruption is pretty bad and we need to prevent it.

**Additional Notes**:
I have added a bunch of special stress tests  for the calltrace storage with the help of Claude Code.
Also, I had to somehow placate ASAN so instead of some pseudo issues it actually started showing real issues in the stresstests.

As a side-effect, I have realized that the calltrace storage is using malloc, potentially in signal handler, when the underlying LongHashTable gets resized - and that can happen when we are trying to put a calltrace from eg. signal handler. And allocating from signal handler is never good ...

**How to test the change?**:
Added a bunch of focused  tests for the calltrace storage functionality. All the existing tests are still passing.

**For Datadog employees**:
- [x] JIRA: [PROF-12778]

[PROF-12778]: https://datadoghq.atlassian.net/browse/PROF-12778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/callTraceStorage.cpp
+++ b/ddprof-lib/src/main/cpp/callTraceStorage.cpp
@@ -17,10 +17,11 @@
 
 // HazardPointer static members
 std::atomic<CallTraceHashTable*> HazardPointer::global_hazard_list[HazardPointer::MAX_THREADS];
+std::atomic<std::thread::id> HazardPointer::slot_owners[HazardPointer::MAX_THREADS];
 
 // HazardPointer implementation
 int HazardPointer::getThreadHazardSlot() {
-    // Signal-safe: use thread ID hash to assign slots
+    // Signal-safe collision resolution: use thread ID hash with semi-random prime step probing
     // This avoids thread_local allocation issues in signal handlers
     std::thread::id tid = std::this_thread::get_id();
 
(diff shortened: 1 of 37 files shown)


zhengyu123 (changes requested):
I have yet gone through tests.

zhengyu123 (changes requested):

ddprof-lib/src/main/cpp/callTraceStorage.h
```diff
@@ -34,43 +36,112 @@ class CallTraceStorage {
...
+    std::atomic<CallTraceHashTable*> _scratch_storage;
+    
+    // Generation counter for ABA protection during table swaps
+    std::atomic<u32> _generation_counter;
+    
+    // Liveness checkers - protected by simple spinlock during registration/clear
+    // Using vector instead of unordered_set since std::function cannot be hashed
     std::vector<LivenessChecker> _liveness_checkers;
-    SpinLock _lock;
+    volatile int _liveness_lock;  // Simple atomic lock for rare liveness operations
```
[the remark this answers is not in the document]
jbachorik: yep

ddprof-lib/src/main/cpp/callTraceStorage.cpp
```diff
@@ -8,195 +8,390 @@
...
+    // Apply Knuth multiplicative hash directly to thread ID hash
+    size_t hash = std::hash<std::thread::id>{}(tid) * KNUTH_MULTIPLICATIVE_CONSTANT;
+
+    // Use high bits for better distribution (shift right to get top bits)
+    return static_cast<int>((hash >> (sizeof(size_t) * 8 - 8)) % MAX_THREADS);
+}
+
+HazardPointer::HazardPointer(CallTraceHashTable* resource) : active_(true) {
+    // Get thread hazard slot using signal-safe method
+    my_slot_ = getThreadHazardSlot();
```
zhengyu123: What happens if there is collision?

jbachorik: 🤦 Right .. for CPU and wallclock, 256 slots should be quite enough. 
But for the JVMTI based samples we can have many more threads than the CPU cores.

I have expanded the hazard slot table, added bounded probing for conflict resolution and if we exhaust the table, we would just drop the trace instead of creating crash hazard.

The hazard slot table is now taking 64KiB which seems like a very small memory overhead we would pay for having pretty stable hazard pointer system.

zhengyu123 (approved):
LGTM

PR #282, opened 2025-10-24
[PROF-12714] Add CriticalNative support

zhengyu123:
**What does this PR do?**:

**Motivation**:

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:
- [ ] JIRA: [JIRA-XXXX]

### DIFF
--- a/ddprof-lib/src/main/cpp/javaApi.cpp
+++ b/ddprof-lib/src/main/cpp/javaApi.cpp
@@ -68,11 +68,6 @@ Java_com_datadoghq_profiler_JavaProfiler_init0(JNIEnv *env, jclass unused) {
   return VM::initProfilerBridge(nullptr, true);
 }
 
-extern "C" JNIEXPORT jboolean JNICALL
-JavaCritical_com_datadoghq_profiler_JavaProfiler_init0() {
-  return VM::initProfilerBridge(nullptr, true);
-}
-
 extern "C" DLLEXPORT void JNICALL
 Java_com_datadoghq_profiler_JavaProfiler_stop0(JNIEnv *env, jobject unused) {
   Error error = Profiler::instance()->stop();
@@ -86,12 +81,6 @@ extern "C" DLLEXPORT jint JNICALL
 Java_com_datadoghq_profiler_JavaProfiler_getTid0(JNIEnv *env, jclass unused) {
   return OS::threadId();
 }
-
-extern "C" DLLEXPORT jint JNICALL
-JavaCritical_com_datadoghq_profiler_JavaProfiler_getTid0() {
-  return OS::threadId();
-}
-
 
 extern "C" DLLEXPORT jstring JNICALL
 Java_com_datadoghq_profiler_JavaProfiler_execute0(JNIEnv *env, jobject unused,
@@ -136,12 +125,12 @@ Java_com_datadoghq_profiler_JavaProfiler_getSamples(JNIEnv *env,
   return (jlong)Profiler::instance()->total_samples();
 }
 
-extern "C" DLLEXPORT jlong JNICALL
-JavaCritical_com_datadoghq_profiler_JavaProfiler_getSamples() {
-  return (jlong)Profiler::instance()->total_samples();
-}
-
 // some duplication between add and remove, though we want to avoid having an extra branch in the hot path
+
+// JavaCritical is faster JNI, but more restrictive - parameters and return value have to be
+// primitives or arrays of primitive types.
+// We direct corresponding JNI calls to JavaCritical to make sure the parameters/return value
+// still compatible in the event of signature changes in the future.
 extern "C" DLLEXPORT void JNICALL
 JavaCritical_com_datadoghq_profiler_JavaProfiler_filterThreadAdd0() {
   ProfiledThread *current = ProfiledThread::current();
@@ -173,12 +162,6 @@ JavaCritical_com_datadoghq_profiler_JavaProfiler_filterThreadAdd0() {
 }
 
 extern "C" DLLEXPORT void JNICALL
-Java_com_datadoghq_profiler_JavaProfiler_filterThreadAdd0(JNIEnv *env,
-                                                          jclass unused) {
-  JavaCritical_com_datadoghq_profiler_JavaProfiler_filterThreadAdd0();
-}
-
-extern "C" DLLEXPORT void JNICALL
 JavaCritical_com_datadoghq_profiler_JavaProfiler_filterThreadRemove0() {
   ProfiledThread *current = ProfiledThread::current();
   if (unlikely(current == nullptr)) {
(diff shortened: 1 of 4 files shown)


ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -68,6 +68,11 @@ Java_com_datadoghq_profiler_JavaProfiler_init0(JNIEnv *env, jclass unused) {
   return VM::initProfilerBridge(nullptr, true);
 }
 
+extern "C" JNIEXPORT jboolean JNICALL
+JavaCritical_com_datadoghq_profiler_JavaProfiler_init0() {
+  return VM::initProfilerBridge(nullptr, true);
+}
+
```
ivoanjo: I'm curious, why have a `JavaCritical` variant for ìnit0`? This one doesn't look like it would ever be on the critical path so I don't see when the JVM would pick this one? :eyes:

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -182,11 +202,15 @@ Java_com_datadoghq_profiler_JavaProfiler_filterThreadRemove0(JNIEnv *env,
...
+  JavaCritical_com_datadoghq_profiler_JavaProfiler_filterThreadRemove0();
+}
+
 // Backward compatibility for existing code
 extern "C" DLLEXPORT void JNICALL
-Java_com_datadoghq_profiler_JavaProfiler_filterThread0(JNIEnv *env,
-                                                       jobject unused,
-                                                       jboolean enable) {
+JavaCritical_com_datadoghq_profiler_JavaProfiler_filterThread0(jboolean enable) {
   ProfiledThread *current = ProfiledThread::current();
```
ivoanjo: Wait, is this correct at all? I know this hasn't changed in this PR, but looking at `AsyncProfiler::filterThread`:

```java
    private void filterThread(Thread thread, boolean enable) {
        if (thread == null || thread == Thread.currentThread()) {
            filterThread0(null, enable);
        } else {
            // Need to take lock to avoid race condition with a thread state change
            synchronized (thread) {
                Thread.State state = thread.getState();
                if (state != Thread.State.NEW && state != Thread.State.TERMINATED) {
                    filterThread0(thread, enable);
                }
            }
        }
    }
```

...that `else` branch seems to be completely broken?

(I looked a bit at `java-profiler` and `dd-trace-java` and it doesn't look like this code is still in-use... Perhaps easier to just remove it?)

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -379,23 +421,39 @@ Java_com_datadoghq_profiler_JavaProfiler_recordQueueEnd0(
...
 
 extern "C" DLLEXPORT jlong JNICALL
 Java_com_datadoghq_profiler_JavaProfiler_tscFrequency0(JNIEnv *env,
-                                                       jobject unused) {
+                                                       jclass unused) {
+  return TSC::frequency();
+}
+
+extern "C" DLLEXPORT jlong JNICALL
+JavaCritical_com_datadoghq_profiler_JavaProfiler_tscFrequency0() {
```
ivoanjo: Similar note to `init0` -- this value seems to be saved in a `static final` on the Java side and never called again.

ivoanjo (commented):
Hey :wave: . As requested, I gave it a stab at reviewing the PR!

I'm missing some context on this code, in particular it was hard to determine which methods were performance-critical and which were used only for setup/testing/etc off the critical path.

I wonder if it's worth documenting and/or reflecting that in the method names.

I saw the benchmarks in JIRA -- I guess the important part is that `ThreadFilterBenchmark` is faster? May be worth posting them here too, to give a bit more context.

ddprof-lib/src/main/cpp/vmStructs_dd.cpp
```diff
@@ -98,6 +99,16 @@ namespace ddprof {
...
 
+  void VMStructs_::initCriticalJNINatives() {
+#ifdef __aarch64__
+    // aarch64 does not support CriticalJNINatives
+    JVMFlag* flag = JVMFlag::find("CriticalJNINatives", {JVMFlag::Type::Bool});
+    if (flag != nullptr && flag->get()) {
+        flag->set(0);
+    }
+#endif // __aarch64__
+  }
```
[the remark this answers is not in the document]
zhengyu123: The is a bug in musl/aarch64 jdk8 port, it did not disable it.

ivoanjo:
> However, given the simplicity of JavaCritical variants, it does not hurt to add them for none performance critical ones.

Hmm... I'll be honest and say that I don't think this is the way to go. Specifically:

1. Adding `JavaCritical` variants means more code + more maintenance for it generically. We need to remember to always keep the copy-pasted stuff up-to-date.
2. It makes it _look_ like those methods are performance-sensitive, which they aren't, so it communicates a weird intent.
3. If we ever want to refactor some of those methods, we're now constrained by existence of the `JavaCritical` variant. What happens if it can't be a `JavaCritical` anymore because we'd like to pass in an object? It may not be clear for many of them we could trivially remove the `JavaCritical` variant and then change the method signature.

Can I maybe convince you to revisit your approach here? :smile: :wink: :eyes:

ivoanjo (approved):
:+1: Looks reasonable! I'd still strongly recommend being a lot more explicit about the ARM64 workaround code, so it's extremely clear for future maintainers why it exists (and when we can get rid of it)

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -113,22 +113,26 @@ Java_com_datadoghq_profiler_JavaProfiler_execute0(JNIEnv *env, jobject unused,
...
-                                                    jobject unused) {
+                                                    jclass unused) {
   return (jlong)Profiler::instance()->total_samples();
 }
 
 // some duplication between add and remove, though we want to avoid having an extra branch in the hot path
+
+// JavaCritical is faster JNI, but more restrictive - parameters and return value have to be
+// primitives or arrays of primitive types.
+// We direct corresponding JNI calls to JavaCritical to make sure the parameters/return value
```
jbachorik: Seems like a word is missing. Maybe this?

```suggestion
// We add direct corresponding JNI calls to JavaCritical to make sure the parameters/return value
```

ddprof-lib/src/main/cpp/vmStructs_dd.cpp
```diff
@@ -98,6 +99,16 @@ namespace ddprof {
     }
   }
 
+  void VMStructs_::initCriticalJNINatives() {
+#ifdef __aarch64__
+    // aarch64 does not support CriticalJNINatives
+    JVMFlag* flag = JVMFlag::find("CriticalJNINatives", {JVMFlag::Type::Bool});
```
jbachorik: IIUC, `-XX:+CriticalJNINatives` requires also `-XX:+UnlockExperimentalVMOptions` :/

This would be a hard sell to our customers.

jbachorik (approved):

PR #284, opened 2025-10-24
Make jmh task to honor $JAVA_TEST_HOME variable

zhengyu123:
**What does this PR do?**:
Allows jmh task to pick up $JAVA_TEST_HOME environment variable.

**Motivation**:
Allows jmh task to pick up $JAVA_TEST_HOME environment variable, to be consistent with other tasks.

**Additional Notes**:

**How to test the change?**:
- Set $JAVA_TEST_HOME to different JDK versions
- Run `gradlew jmh` task
- Check output message and make sure the correct JDK is used.

**For Datadog employees**:
- [ ] JIRA: [JIRA-XXXX]

### DIFF
--- a/ddprof-stresstest/build.gradle
+++ b/ddprof-stresstest/build.gradle
@@ -23,6 +23,11 @@ sourceSets {
 }
 
 jmh {
+  def javaHome = System.getenv("JAVA_TEST_HOME")
+  if (javaHome == null) {
+    javaHome = System.getenv("JAVA_HOME")
+  }
+  jvm = "${javaHome}/bin/java"
   iterations = 3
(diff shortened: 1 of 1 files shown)


ddprof-stresstest/build.gradle
```diff
@@ -23,6 +23,11 @@ sourceSets {
 }
 
 jmh {
+  def javaHome = System.getenv("JAVA_TEST_HOME")
```
r1viollet: Should this not be at a higher level ? Other tests would want to benefit from this ?
The way I understand this is valid only for the stress test ?

jbachorik (approved):

PR #290, opened 2025-11-06
Fix use-after-free in CallTraceStorage and optimize trace collection

jbachorik:
**What does this PR do?**:
The original issue was a use-after-free bug in CallTraceStorage::processTraces() where trace pointers were accessed after the memory was deallocated during triple-buffer rotation.

Key changes:
- Fixed use-after-free by copying preserved traces immediately using putWithExistingId()
- Added optional hook parameter to CallTraceHashTable::collect() for inline processing
- Eliminated separate trace buffers (_standby_traces_buffer, _active_traces_buffer)
- Simplified processTraces() to collect directly to _traces_buffer with hooks
- Removed double iteration pattern (collect then iterate) in favor of single pass
- Added regression test to catch use-after-free with frame content access

The fix ensures memory safety while improving performance by eliminating unnecessary buffer copying and reducing iteration overhead.

**Motivation**:
Sadly, my Java habits made me to miss the fact that I was copying around pointers and then clearing/deallocating the trace holding collections. This leading to potentially invalid memory access.

**How to test the change?**:
Extended the gtests such that the use-after-free is immediately caught there. The test is passing after the changes are applied.

**For Datadog employees**:
- [x] JIRA: [PROF-12970]

🤖 Generated with [Claude Code](https://claude.ai/code)

[PROF-12970]: https://datadoghq.atlassian.net/browse/PROF-12970?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/test_workflow.yml
+++ b/.github/workflows/test_workflow.yml
@@ -124,7 +124,7 @@ jobs:
           name: failures-glibc-${{ matrix.java_version }}-${{ matrix.config }}-amd64
           path: failures_glibc-${{ matrix.java_version }}-${{ matrix.config }}-amd64.txt
       - name: Prepare reports
-        if: always()
+        if: steps.set_enabled.outputs.enabled == 'true'
         run: |
           .github/scripts/prepare_reports.sh
       - name: Upload unwinding reports
@@ -247,7 +247,7 @@ jobs:
           name: failures-musl-${{ matrix.java_version }}-${{ matrix.config }}-amd64
           path: failures_musl-${{ matrix.java_version }}-${{ matrix.config }}-amd64.txt
       - name: Prepare reports
-        if: always()
+        if: steps.set_enabled.outputs.enabled == 'true'
         run: |
           .github/scripts/prepare_reports.sh
       - name: Upload unwinding reports
@@ -378,7 +378,7 @@ jobs:
           name: failures-glibc-${{ matrix.java_version }}-${{ matrix.config }}-aarch64
           path: failures_glibc-${{ matrix.java_version }}-${{ matrix.config }}-aarch64.txt
       - name: Prepare reports
-        if: always()
+        if: steps.set_enabled.outputs.enabled == 'true'
         run: |
           .github/scripts/prepare_reports.sh
       - name: Upload unwinding reports
@@ -480,7 +480,7 @@ jobs:
            name: failures-musl-${{ matrix.java_version }}-${{ matrix.config }}-aarch64
            path: failures_musl-${{ matrix.java_version }}-${{ matrix.config }}-aarch64.txt
        - name: Prepare reports
-         if: always()
+         if: steps.set_enabled.outputs.enabled == 'true'
          run: |
            .github/scripts/prepare_reports.sh
        - name: Upload unwinding reports
--- a/gradle/patching.gradle
+++ b/gradle/patching.gradle
@@ -220,7 +220,7 @@ ext.upstreamPatches = [
 
   // Symbol implementation patches for DataDog-specific API extensions
   "symbols_linux.cpp": [
-    validations: [[contains: "#ifdef __linux__"], [contains: "_parsed_inodes"]],
+    validations: [[contains: "#ifdef __linux__"], [contains: "_parsed_inodes"], [contains: "loadSymbolTable"]],
     operations: [
       [
         type: "method_implementation",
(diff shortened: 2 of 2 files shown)


ddprof-lib/src/main/cpp/callTraceStorage.cpp
```diff
@@ -349,65 +342,52 @@ void CallTraceStorage::processTraces(std::function<void(const std::unordered_set
...
-    }
+    });
     
-    // Step 2: Clear standby after collection, prepare for rotation
-    current_standby->clear();
-    u64 new_instance_id = getNextInstanceId();
-    current_standby->setInstanceId(new_instance_id);
+    // Step 3: Clear original_standby after collection
+    original_standby->clear();
+    // u64 new_instance_id = getNextInstanceId();
```
rkennke: Remove commented out code, unless it is still needed - then uncomment it.

ddprof-lib/src/main/cpp/primeProbing.h
```diff
@@ -0,0 +1,57 @@
...
+        return PRIME_STEPS[(hash >> 4) % PRIME_STEP_COUNT];
+    }
+
+    inline static u32 advanceSlot(u32 slot, u32 prime_step, u32 capacity) {
+        return (slot + prime_step) & (capacity - 1);
+    }
+
+public:
+    HashProbe(u64 hash, u32 capacity) 
+        : _slot(hash & (capacity - 1))
```
zhengyu123: Looks like `capacity` has to be power of 2, right? If so, probably need an assert.

ddprof-lib/src/main/cpp/primeProbing.h
```diff
@@ -0,0 +1,57 @@
...
+    // Prime probing constants for better hash distribution and collision resolution
+    static constexpr int PRIME_STEP_COUNT = 16;    // Number of prime steps for collision resolution
+
+    // Prime numbers for semi-random probing, selected to provide good distribution
+    // These primes are carefully chosen to avoid patterns and clustering
+    static constexpr u32 PRIME_STEPS[PRIME_STEP_COUNT] = {
+        1009, 1013, 1019, 1021, 1031, 1033, 1039, 1049,
+        1051, 1061, 1063, 1069, 1087, 1091, 1093, 1097
+    };
+
```
zhengyu123: Suggest:

```
static constexpr u32 PRIME_STEPS[] = { .... };
static constexpr int PRIME_STEP_COUNT = sizeof(PRIME_STEPS) / sizeof(u32);
```

ddprof-lib/src/main/cpp/callTraceStorage.cpp
```diff
@@ -349,65 +351,52 @@ void CallTraceStorage::processTraces(std::function<void(const std::unordered_set
...
-    _active_traces_buffer.clear();
 
-    // Step 1: Collect from current standby FIRST (preserved traces from previous cycle)
-    current_standby->collect(_standby_traces_buffer);
-    
-    // Immediately preserve standby traces that need to be kept for next cycle
-    for (CallTrace* trace : _standby_traces_buffer) {
+    // Step 1: Collect from current standby directly to _traces_buffer with hook for immediate preservation
+    // Only 'original_active' can be used from multuple threads at this moment.
+    // Both 'original_standby' and 'original_scratch' can be used only from the single threa executing 'processTraces'
```
zhengyu123: typo: `threa` -> `thread`

ddprof-lib/src/main/cpp/callTraceStorage.cpp
```diff
@@ -349,65 +351,52 @@ void CallTraceStorage::processTraces(std::function<void(const std::unordered_set
...
     CallTraceHashTable* old_scratch = _scratch_storage.exchange(old_active, std::memory_order_acq_rel);
     _standby_storage.store(old_scratch, std::memory_order_release);
     
-    // Step 5: NOW collect from scratch (old active, now read-only)
-    old_active->collect(_active_traces_buffer);
-    
-    // Preserve traces from old active too
-    for (CallTrace* trace : _active_traces_buffer) {
+    // Step 6: Collect from old active directly to _traces_buffer with hook for immediate preservation
+    old_active->collect(_traces_buffer, [&](CallTrace* trace) {
```
zhengyu123: Even after `CAS`, there may be threads that reference to now  stalled `old_active`, is it safe to collect here?

zhengyu123 (changes requested):

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
```diff
@@ -70,7 +71,24 @@ class LongHashTable {
...
+    }
+    
+    // Move to next slot
+    if (probe.hasNext()) {
+      probe.next();
+      return probe.slot();
+    }
+    
+    // Fallback: wrap around
+    return (slot + 1) & (_capacity - 1);
```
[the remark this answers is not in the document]
jbachorik: I reworked this and capacity can be an arbitrary number

ddprof-lib/src/main/cpp/callTraceStorage.cpp
jbachorik: True. Good catch!

I have added hazard pointer wait to make sure all the pending put operation on the 'original' active are finished before starting the collection.

PR #292, opened 2025-11-10
Fix incorrect freeing order in ProfiledThread

jbachorik:
**What does this PR do?**:
It changes the order of deleting a pointer ad 'nulling' it

**Motivation**:
The current code is wrong. The pointer gets deleted first and only then 'nulled'. This can lead to use-after-free, as pointed by ASAN in one of our nightly sanitized runs.

**Additional Notes**:

**For Datadog employees**:


jbachorik:
> (In general, I would prefer if unrelated changes could be done separately...)

Yes, in general, agreed. Although there is a thin line here - do we want to have separate PRs for comment typos and updates?

PR #297, opened 2025-11-19
Solid TLS Priming

jbachorik:
**What does this PR do?**:
Improve the TLS priming for reliability and ability to handle new native threads.

**Motivation**:
Currently, we are priming all existing threads for profiler specific TLS slot - we use a special signal to run the priming code and assign pre-allocated TLS instance to each existing thread.
There are two issues with that:
- the signal we are using is not guaranteed to get delivered
- we don't see any new native threads that start after the initial priming

The solution is to:
- use an RT* signal for priming
- have a `/proc` FS based 'watcher' which will try and prime new native threads

**Additional Notes**:
This is a prerequisite for the OTEL thread-level-context work.
The description of the feature/changes is in [TlsPriming.md doc](https://github.com/DataDog/java-profiler/blob/jb/tls_priming/docs/architecture/TlsPriming.md)

**How to test the change?**:

**For Datadog employees**:
- [x] JIRA: [PROF-13093]

[PROF-13093]: https://datadoghq.atlassian.net/browse/PROF-13093?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/os_linux_dd.cpp
+++ b/ddprof-lib/src/main/cpp/os_linux_dd.cpp
@@ -1,9 +1,19 @@
 #ifdef __linux__
 
 #include "os_dd.h"
+#include "common.h"
 #include <signal.h>
 #include <unistd.h>
 #include <sys/types.h>
+#include <sys/syscall.h>
+#include <sys/inotify.h>
+#include <dirent.h>
+#include <errno.h>
+#include <pthread.h>
+#include <atomic>
+#include <cstdio>
+#include <cstring>
+#include <memory>
 
 #ifndef __musl__
 #include <malloc.h>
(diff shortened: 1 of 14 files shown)


ddprof-lib/src/main/cpp/os_linux_dd.cpp
```diff
@@ -42,4 +89,221 @@ SigAction ddprof::OS::replaceSigbusHandler(SigAction action) {
...
+  
+  if (pthread_create(&g_watcher_thread, nullptr, threadDirectoryWatcherLoop, nullptr) != 0) {
+    TEST_LOG("Failed to create thread directory watcher thread: %s", strerror(errno));
+    g_watcher_running.store(false);
+    g_watcher_fd.store(-1);
+    close(inotify_fd);
+    return false;
+  }
+
+  g_watcher_thread_created.store(true);
```
rkennke: Is this sufficient to avoid races? In other words, is this method only ever called non-concurrently? Because otherwise we have a window between pthread_create() and store() where a concurrent load() would still see false.

rkennke (commented):
Good change overall. I have some questions below (mostly for my own understanding).

rkennke (approved):

PR #300, opened 2025-11-21
Change the context storage implementation

jbachorik:
**What does this PR do?**:
It changes the tid-based huge-map for context with TLS based approach.

**Motivation**:
There are two main points here:
- TLS based implementation is much more conservative than the huge-map, which had to reserve memory for all possible tids
- Making the context storage mechanism compatible with the OTel ebpf profiler thread-level context proposal

**Additional Notes**:
The actual context payload hasn't been changed yet to mimic the PolarSignal PoC for the OTel thread-level context propagation.

The JVMTI wallclock will not work with TLS context as the sampling is done from a separate thread and accessing TLS is problematic. This is probably not a blocker, since it is marked as experimental and very probably be superseded by some other solution.

**How to test the change?**:
Added a bunch of native and Java tests exercising the context manipulation. 
All existing tests must pass.

**For Datadog employees**:
- [x] JIRA: [PROF-13116]

[PROF-13116]: https://datadoghq.atlassian.net/browse/PROF-13116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -1281,8 +1281,26 @@ void Recording::writeUnwindFailures(Buffer *buf) {
...
-  buf->putVar64(context.spanId);
-  buf->putVar64(context.rootSpanId);
+  u64 spanId = 0;
+  u64 rootSpanId = 0;
+  u64 stored = context.checksum;
+  if (stored) {
+    spanId = context.spanId;
+    rootSpanId = context.rootSpanId;
+
+    u64 swappedRootSpanId = ((rootSpanId & 0xFFFFFFFFULL) << 32) | (rootSpanId >> 32);
```
rkennke: Should this use the Context::checksum() method instead?

ddprof-lib/src/main/cpp/thread.h
```diff
@@ -177,13 +181,18 @@ class ProfiledThread : public ThreadLocalData {
   int getHazardSlot() { return _hazard_slot; }
 
   // context sharing TLS
-  inline void markContextTlsInitialized() {
+  inline void markContextTlsInitialized(Context* ctx_ptr) {
```
rkennke: The `inline` here is redundant.

ddprof-lib/src/main/cpp/thread.h
```diff
@@ -177,13 +181,18 @@ class ProfiledThread : public ThreadLocalData {
...
+    _ctx_tls_ptr = ctx_ptr;
     _ctx_tls_initialized = true;
   }
 
   inline bool isContextTlsInitialized() {
     return _ctx_tls_initialized;
   }
+
+  inline Context* getContextTlsPtr() {
```
rkennke: And here.

ddprof-lib/src/main/java/com/datadoghq/profiler/BufferWriter.java
```diff
@@ -0,0 +1,181 @@
...
+         * <p>Ordered write guarantees that this write will not be reordered with respect to prior writes,
+         * but subsequent operations may be reordered before this write. This is more efficient than a
+         * volatile write when full bidirectional ordering is not required.
+         *
+         * @param buffer the direct ByteBuffer to write to
+         * @param offset the offset in bytes from the buffer's base address
+         * @param value the long value to write
+         * @return the written value
+         */
+        long writeLong(ByteBuffer buffer, int offset, long value);
```
rkennke: Maybe rename to writeLongOrdered() to make it clear that this is not just a plain write?
Also, have you verified that calls to the implementation are properly inlined? Otherwise you would lose the advantage of ordered vs volatile writes.

rkennke (changes requested):

ddprof-lib/src/main/java/com/datadoghq/profiler/BufferWriter8.java
```diff
@@ -0,0 +1,34 @@
+package com.datadoghq.profiler;
+
+import sun.nio.ch.DirectBuffer;
+
+import java.nio.ByteBuffer;
+
+public final class BufferWriter8 implements BufferWriter.Impl {
+    @Override
+    public long writeLong(ByteBuffer buffer, int offset, long value) {
+        JavaProfiler.UNSAFE.putOrderedLong(null, ((DirectBuffer) buffer).address() + offset, value);
```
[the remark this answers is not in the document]
jbachorik: Yeah, makes sense.

ddprof-lib/src/main/cpp/context.h
```diff
@@ -19,51 +19,41 @@
...
+  static Context& get();
+
+  inline static u64 checksum(u64 spanId, u64 rootSpanId) {
+    u64 swappedRootSpanId = ((rootSpanId & 0xFFFFFFFFULL) << 32) | (rootSpanId >> 32);
+    u64 computed = (spanId * KNUTH_MULTIPLICATIVE_CONSTANT) ^ (swappedRootSpanId * KNUTH_MULTIPLICATIVE_CONSTANT);
+    return computed == 0 ? 0xffffffffffffffffull : computed;
+  }
 };
 
+DLLEXPORT extern thread_local Context context_tls_v1;
```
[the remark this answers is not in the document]
jbachorik: Ultimately, we want this to be possible to be looked up from an ebpf profiler and as such it needs a stable exported symbol.

rkennke (commented):
Much better! I have one new comment, see below. And you might want to check copyright headers for new files. (Not sure how important that is for Datadog?)

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -553,3 +529,53 @@ Java_com_datadoghq_profiler_OTelContext_readProcessCtx0(JNIEnv *env, jclass unus
...
+  ctx.checksum = Contexts::checksum(spanId, rootSpanId);
+
+  return ctx.checksum;
+}
+
+extern "C" DLLEXPORT jlong JNICALL
+Java_com_datadoghq_profiler_ThreadContext_setContextSlot0(JNIEnv* env, jclass unused, jint offset, jint value) {
+  Context& ctx = Contexts::get();
+  ctx.tags[offset].value = (u32)value;
+  return offset * value;
```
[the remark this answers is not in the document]
jbachorik: It was used only in benchmarks to have something for `Blackhole` to consume, but probably not important.
Gone.

rkennke (approved):
Looks good, thank you!

PR #301, opened 2025-11-26
Fix another hidden use-after-free in CallTraceStorage

jbachorik:
**What does this PR do?**:
Here we address the (almost certain) last use-after-free bug in `CallTraceStorage` implementation

**Additional Notes**:
The gist of the fix is to make the `LinearAllocator` cleanup 2-phased - first just reset the allocator to start handing out new chunks and actually deallocate the previously used chunks only when it is safe, after the all the collected traces had been processed.

I have generated the reproducer test, hence the more verbose comments, but I think they might be useful for history keeping. The test is using ASAN poisoning and will reliably crash when run under ASAN.

**How to test the change?**:
Added a new gtest which is supposed to be run under ASAN.

**For Datadog employees**:
- [x] JIRA: [PROF-13156]

[PROF-13156]: https://datadoghq.atlassian.net/browse/PROF-13156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/callTraceHashTable.cpp
+++ b/ddprof-lib/src/main/cpp/callTraceHashTable.cpp
@@ -102,10 +102,10 @@ CallTraceHashTable::~CallTraceHashTable() {
 }
 
 
-void CallTraceHashTable::clear() {
-  // Wait for all hazard pointers to clear before deallocation to prevent races
+ChunkList CallTraceHashTable::clearTableOnly() {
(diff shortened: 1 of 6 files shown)


rkennke (approved):
Looks good!

PR #303, opened 2025-11-27
TLS priming optimization

jbachorik:
**What does this PR do?**:

Optimizes TLS (Thread-Local Storage) priming by introducing a lock-free bitset to track Java/JVM threads. This allows the profiler to skip sending TLS priming signals to Java threads (which are already initialized via JVMTI callbacks), reducing overhead by ~95% for typical Java workloads.

Key changes:
- **New `LockFreeBitset` template class** (`lockFree.h`): Reusable double-hashing bitset for concurrent membership tracking
  - Uses two independent hash functions to minimize false positives (probability ≈ (M/N)²)
  - Interleaved memory layout for L1 cache locality
  - Lock-free atomic operations safe for signal handlers
  - 32 KB memory footprint (fits entirely in L1 cache)

- **Java thread tracking**: Register Java threads in the bitset via JVMTI callbacks, skip TLS priming signals for them

- **Thread watcher optimization**: 20ms delay after thread creation allows JVMTI registration before checking the bitset

**Motivation**:

Before this optimization, the TLS priming system sent signals to ALL threads (Java + native) when new threads were detected. Since Java threads are already initialized via JVMTI `ThreadStart` callbacks, these signals were wasteful.

Additionally, before TLS priming was implemented, native threads created after profiling started were **completely invisible** to the profiler. With this change:
- The vast majority of native threads (99.6-99.997%) are correctly primed and profiled
- Only a tiny fraction might be skipped due to false positives in the Java thread filter
- This is dramatically better than 100% of late-created native threads being invisible

**Additional Notes**:

- **Configuration**: TLS priming can be controlled via `DD_PROFILER_TLS_WATCHER` environment variable or JVM system property (set to `0` to disable, `1` to enable - default enabled)
- **False positive analysis**: Double-hashing reduces false positives from p to p². With 16384 bits per array:
  - 100 threads → 0.003% false positive rate
  - 500 threads → 0.09% false positive rate
  - 1000 threads → 0.37% false positive rate
- **Memory layout**: Interleaved array `[word1_0, word2_0, word1_1, word2_1, ...]` ensures both hash lookups access adjacent cache lines

**How to test the change?**:

1. **Unit tests**: New gtest suites added:
   - `test_lockFreeBitset.cpp`: 15 tests for the generic bitset implementation
   - `test_javaThreadBitset.cpp`: 9 tests for Java thread tracking

2. **All existing tests pass**: 113 gtest tests, full test suite

**For Datadog employees**:
[...]

PR #305, opened 2025-12-05
Backport critical fixes to 1.34.4

jbachorik:
Backport critical fixes for 1.34.4

Unfortunately, the TLS priming related changes I did earlier did cause performance overhead - it is addressed in https://github.com/DataDog/java-profiler/pull/303 but I would prefer delaying 1.35.0 until we have run all the changes there for a while and have the crash and deadlock preventing fixes be a part of 1.34.4


zhengyu123 (approved):
LGTM

PR #308, opened 2025-12-09
Instrument native threads

zhengyu123:
**What does this PR do?**:
Instrument/profile native threads, the threads that are created/started outside of JVM, on hotspot/non-musl based JVM. 

**Motivation**:
Enhance Java profiler to profiler native threads.

**Additional Notes**:
The feature is now only enabled for hotspot based JVM running on non-musl Linux platform. The reasons:
- A crash seen on aarch64/musl /JDK11. It might not be related to this change, but it is hard to confirm. Disable the feature for musl based Linux for now.
- J9 has issues to walk native only thread stack in release build, it shows only
` .no_java_frame`
 while debug build shows correct stack.

**How to test the change?**:
- Regular tests 
- JDK tier1 tests with profiler agent. Although, there are failures, but match java-profiler main line.
   There are failures that are expected:
    * `HeapMonitor` uses agent, which conflicts with profiler agent
    * Compiler frame is not compatible with agent.

**For Datadog employees**:
- [X] JIRA: [PROF-11577](https://datadoghq.atlassian.net/browse/PROF-11577)

[PROF-11577]: https://datadoghq.atlassian.net/browse/PROF-11577?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/ctimer_linux.cpp
+++ b/ddprof-lib/src/main/cpp/ctimer_linux.cpp
@@ -88,12 +88,13 @@ static Error patch_libraries() {
 
   const CodeCacheArray& native_libs = Libraries::instance()->native_libs();
   int count = 0;
-  size_t size = count * sizeof(PatchEntry);
+  int num_of_libs = native_libs.count();
+  size_t size = num_of_libs * sizeof(PatchEntry);
   patched_entries = (PatchEntry*)malloc(size);
   memset((void*)patched_entries, 0, size);
   TEST_LOG("Patching libraries");
 
-  for (int index = 0; index < native_libs.count(); index++) {
+  for (int index = 0; index < num_of_libs; index++) {
      CodeCache* lib = native_libs.at(index);
      // Don't patch self
      if (strcmp(lib->name(), info.dli_fname) == 0) {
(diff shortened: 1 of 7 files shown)


ddprof-lib/src/main/cpp/ctimer_linux.cpp
```diff
@@ -62,21 +136,52 @@ static int pthread_setspecific_hook(pthread_key_t key, const void *value) {
...
-  if (VM::isZing()) {
-    CodeCache *libazsys = Libraries::instance()->findLibraryByName("libazsys");
-    if (libazsys != NULL) {
-      void **entry = libazsys->findImport(im_pthread_setspecific);
-      if (entry != NULL) {
-        return entry;
-      }
-    }
+static Error patch_libraries_for_J9_or_musl() {
+   CodeCache *lib = Libraries::instance()->findJvmLibrary("libj9thr");
```
jbachorik: Is this right? 
The `libj9thr` will be available only on J9, regardless of whether we are running on musl or not.

jbachorik (changes requested):
Having the libraries patching tied to ctimer (CPU profiling) does not sound right.
We have a bunch of other engines that can be used either for CPU or wallclock profiling and their functionality would be very inconsistent.

I think the library patching should go to a more generic place - and also be called from a more generic place, like eg. `profiler.cpp`

zhengyu123:
> How will this work with dynamically loaded libraries? Am I reading the code right that we will not patch those?

Let's open another ticket for this as well.

jbachorik (approved):
Approved with the two followup tickets

PR #314, opened 2025-12-16
Fix TLS re-entrancy crash in signal handlers

jbachorik:
**What does this PR do?**:
Fixes a critical crash caused by re-entrant TLS allocation in CPU sampling signal handlers. Replaces `VM::jni()` calls with direct thread state checks to distinguish Java threads from JVM internal threads without triggering `__tls_get_addr`.

**Motivation**:
The profiler was crashing with heap corruption when SIGPROF interrupted ForkJoinWorkerThread during TLS initialization. The crash occurred because:
1. Thread interrupted during `__tls_get_addr` (TLS allocation)
2. Signal handler called `VM::jni()` → `JavaVM->GetEnv()` → `jni_GetEnv` → `__tls_get_addr`
3. Re-entrant call to `__tls_get_addr` corrupted heap metadata
4. Later `free()` crashed with corrupted heap

Stack trace showed:
```
C  [libc.so.6+0xadd55]  __libc_free+0x25
C  [ld-linux-x86-64.so.2+0x1785c]  __tls_get_addr+0x3c
V  [libjvm.so+0x9b6567]  jni_GetEnv+0x47
C  [libjavaProfiler...so+0x1d24a]  CTimer::signalHandler+0xea
C  [ld-linux-x86-64.so.2+0x1785c]  __tls_get_addr+0x3c  ← Re-entrant!
```

**Additional Notes**:
- Fixed in 4 signal handlers: CTimer, ITimer, PerfEvents, WallClock
- Thread state range [4, 12) corresponds to Java threads (_thread_in_native to _thread_max_state)
- Direct memory read of `vm_thread->state()` is async-signal-safe
- Added detailed comments with OpenJDK reference links

**How to test the change?**:
The crash is race-condition dependent and hard to reproduce deterministically. Testing requires:
1. Running with ZGC (which has more TLS usage)
2. High-frequency CPU sampling during ForkJoinPool thread creation
3. Extended stress testing under load

The fix has been validated through:
- Code review against async-signal-safety requirements
- Verification that thread state checks match wallClock.cpp existing pattern
- Confirmation that no TLS access occurs in signal handlers

**For Datadog employees**:
- [x] JIRA: [PROF-13176]

[PROF-13176]: https://datadoghq.atlassian.net/browse/PROF-13176?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/ctimer_linux.cpp
+++ b/ddprof-lib/src/main/cpp/ctimer_linux.cpp
@@ -22,6 +22,7 @@
 #include "debugSupport.h"
 #include "libraries.h"
 #include "profiler.h"
+#include "threadState.h"
 #include "vmStructs.h"
 #include <assert.h>
 #include <dlfcn.h>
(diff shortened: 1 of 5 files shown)


rkennke (approved):
Good!
(And I've been scratching my head about what we have to do with SSL *duh*)

zhengyu123 (approved):
LGTM

PR #347, opened 2026-01-29
Add OTEP #4947 OTEL-compatible context storage

jbachorik:
**What does this PR do?**:

Exposes trace context via an OTEP #4947-compliant TLS pointer (`otel_thread_ctx_v1`),
discoverable by external profilers through the ELF dynsym table. OTEL context storage
is the sole context storage mode; there is no legacy profiler-only mode.

Each thread gets a 640-byte `OtelThreadContextRecord` containing:
- W3C trace_id (128-bit) and span_id (64-bit) in big-endian byte encoding
- Custom attributes encoded as key-index + length + UTF-8 value pairs in `attrs_data[612]`

A `ContextApi` class provides signal-safe read and snapshot operations on the OTEP record.

**Context data flow**

```mermaid
flowchart LR
    T["Java tracer"]

    subgraph thread ["Per-thread (ProfiledThread)"]
        REC["OtelThreadContextRecord\nvalid · trace_id · span_id · attrs_data"]
        SC["JFR sidecar\nu32 encodings · LRS"]
    end

    T -- "put(): valid=0 → write → valid=1" --> REC
    T -- "put(): LRS\nsetContextAttribute(): u32 encoding" --> SC
    T -- "setContextAttribute(): UTF-8 value" --> REC

    REC -- "otel_thread_ctx_v1 (dynsym)" --> EP["External profiler\n(OTEP #4947)"]
    SC -- "O(1), signal-safe" --> SH["Signal handler\n→ JFR writer"]
    REC -- "valid + span_id" --> SH
```

**Publication protocol (detach/attach)**

```mermaid
sequenceDiagram
    participant W as Java tracer (writer)
    participant rec as OtelThreadContextRecord
    participant R as Reader (signal handler / external profiler)

    Note over W,R: Thread init: otel_thread_ctx_v1 permanently points to record

    W->>rec: valid = 0
    W->>rec: storeFence()
    Note over R: load valid → 0, skip record
    W->>rec: write trace_id, span_id, attrs_data, LRS
    W->>rec: storeFence()
    W->>rec: valid = 1
    Note over R: load valid → 1, read record safely
    R->>rec: read trace_id, span_id, attrs_data
```

**Context setting API:**

- **New**: `ThreadContext.put(localRootSpanId, spanId, traceIdHigh, traceIdLow)` — writes
  trace_id + span_id to the OTEP record, stores localRootSpanId in `ProfiledThread` for
  O(1) signal-safe reads (used for endpoint correlation in JFR).
- **Existing (deprecated)**: `ThreadContext.put(spanId, rootSpanId)` — delegates to the
  4-arg overload with `traceIdHigh=0` and `traceIdLow=spanId`. Marked `@Deprecated`; callers
  needing correct OTEP interop must switch to the 4-arg API.
- **New**: `ThreadContext.setContextAttribute(keyIndex, value)` — writes custom attributes.
  Encodes the string value into both the sidecar (u32 Dictionary encoding for JFR) and
  `attrs_data` (UTF-8 for external profilers) via ByteBuffer. JNI is called only on cache
  miss to register a new constant.

**JFR tag encoding sidecar:**

`ProfiledThread` caches pre-computed Dictionary encodings (`_otel_tag_encodings[]`) and
`_otel_local_root_span_id` alongside the OTEP record. The ByteBuffer write path populates
both the OTEP `attrs_data` (raw strings for external profilers) and the sidecar (u32
encodings for our JFR writer). Signal handlers read only the sidecar — O(1) array index,
zero string lookups or hash-table probes per sample.

**Motivation**:

Enable external profilers (e.g. OTel eBPF profiler) to discover and read the Java
profiler's trace context using the standardized OTEP #4947 protocol, while preserving
localRootSpanId for Datadog endpoint correlation.

**Additional Notes**:

- Signal safety: attribute string values are pre-registered in the Dictionary from the
  JNI thread; the signal handler reads only the sidecar (no Dictionary access at all)
- Publication protocol: 3-step with release barriers — set valid=0 + storeFence, write
  trace/span/attrs data, storeFence + set valid=1 — ensuring signal handlers never see
  partially-written records. The TLS pointer `otel_thread_ctx_v1` is set permanently on
  thread init and never modified during context writes.
- Thread lifecycle: `OtelThreadContextRecord` is zeroed and the TLS pointer nulled in
  `releaseFromBuffer()` to prevent external profilers from dereferencing a recycled record
- Sidecar fields are zeroed in `releaseFromBuffer()` before the OTEL record is cleared
- Architecture docs: `doc/architecture/TLSContext.md`

**How to test the change?**:

Automated tests (7 tests):
```bash
./gradlew :ddprof-test:testRelease --tests "com.datadoghq.profiler.context.OtelContextStorageModeTest"
```

Covers: OTEL context round-trip with 128-bit trace ID + localRootSpanId, custom
attributes, attribute overflow, sequential context updates, thread isolation,
span transition clearing attributes, attribute cache isolation.

Symbol verification:
```bash
nm -D libjavaProfiler.so | grep otel_thread_ctx_v1
```

**For Datadog employees**:

- [x] JIRA: [PROF-14224](https://datadoghq.atlassian.net/browse/PROF-14224)

[PROF-14224]: https://datadoghq.atlassian.net/browse/PROF-14224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -1450,6 +1451,21 @@ void Recording::writeContext(Buffer *buf, Context &context) {
...
+  ContextApi::get(spanId, rootSpanId);
+  buf->putVar64(spanId);
+  buf->putVar64(rootSpanId);
+
+  // Tags still come from TLS Context (even in OTEL mode, for compatibility)
+  Context &context = Contexts::get();
+  for (size_t i = 0; i < Profiler::instance()->numContextAttributes(); i++) {
+    Tag tag = context.get_tag(i);
+    buf->putVar32(tag.value);
+  }
```
ivoanjo: We could put them in `attrs-data` for otel. In particular `trace endpoint` would be cool to keep, but in general we could fully support custom attributes for otel.

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -460,6 +462,21 @@ Java_com_datadoghq_profiler_OTelContext_setProcessCtx0(JNIEnv *env,
   JniString version_str(env, version);
   JniString tracer_version_str(env, tracer_version);
 
+  // Default TLS configuration for profiler context slots (0-9)
+  // New format: key names in index order (position = key index)
+  static const char* default_key_map[] = {
+    "tag.0", "tag.1", "tag.2", "tag.3", "tag.4",
+    "tag.5", "tag.6", "tag.7", "tag.8", "tag.9",
+    NULL
+  };
```
ivoanjo: I guess right now we don't know the custom attributes key names in advance, so we drop the names...?

ddprof-lib/src/main/cpp/otel_process_ctx.h
```diff
@@ -24,6 +24,24 @@ extern "C" {
...
+ *
+ * These fields allow external profilers to discover and decode thread-local context records.
+ */
+typedef struct {
+  char *schema_version;      // TLS schema version string (e.g. "tlsdesc_v1_dev")
+  int max_record_size;       // Maximum bytes per TLS record
+  // Key index to name mapping (NULL-terminated array of key names)
+  // Position in array = key index (e.g. ["method", "route", NULL] means index 0 = "method", index 1 = "route")
+  char **attribute_key_map;
+} otel_tls_config;
```
ivoanjo: Ah, right, the upstream library needs to get support for the `attribute_key_map`... I guess we'll eventually want to turn this into a PR for  https://github.com/open-telemetry/sig-profiling/blob/main/process-context/c-and-cpp/ .

ddprof-lib/src/main/cpp/otel_process_ctx.cpp
```diff
@@ -66,19 +68,37 @@ static const otel_process_ctx_data empty_data = {
...
+ * - version:           uint32 = 2
+ * - payload_size:      uint32 (size of protobuf payload)
+ * - published_at_ns:   uint64 (timestamp in nanoseconds since epoch, 0 = update in progress)
+ * - payload:           pointer to protobuf-encoded Resource message
  */
 typedef struct __attribute__((packed, aligned(8))) {
-  char otel_process_ctx_signature[8]; // Always "OTEL_CTX"
-  // TODO: Is version useful? Should we just get rid of it?
-  uint32_t otel_process_ctx_version;  // Always > 0, incremented when the data structure changes
-  // TODO: Is size useful? Should we just get rid of it?
```
ivoanjo: Wow this is a lot of changes to the upstream library... I'm slightly worried that some detail may be lost and diverge from upstream with all these changes.

Maybe worth trying to extract a PR against  https://github.com/open-telemetry/sig-profiling/blob/main/process-context/c-and-cpp/otel_process_ctx.c ? (I can poke at it if needed)

ddprof-test/src/test/java/com/datadoghq/profiler/context/ProcessContextTest.java
```diff
@@ -49,23 +49,38 @@ private static class OtelMappingInfo {
...
+                }
+                Matcher memfdMatcher = memfdPattern.matcher(line);
+                if (memfdMatcher.matches()) {
                     return new OtelMappingInfo(
-                        matcher.group(1),
-                        matcher.group(2),
-                        matcher.group(3)
+                        memfdMatcher.group(1),
+                        memfdMatcher.group(2),
+                        memfdMatcher.group(3)
```
ivoanjo: I suggest matching only on `OTEL_CTX` -- may be easier?

ivoanjo (commented):
Gave it a pass! There were a few of "uuuuh what? 👀 moments". I guess AI?

ddprof-lib/src/main/cpp/context_api.cpp
```diff
@@ -0,0 +1,266 @@
...
+    if (__atomic_load_n(&_initialized, __ATOMIC_ACQUIRE)) {
+        return true;
+    }
+
+    ContextStorageMode mode = args._context_storage;
+    if (mode == CTX_STORAGE_OTEL) {
+        if (!OtelContexts::initialize()) {
+            mode = CTX_STORAGE_PROFILER;
+            __atomic_store_n(&_mode, mode, __ATOMIC_RELEASE);
+            return false;
```
jbachorik: Returning CTX_STORAGE_PROFILER would be more understandable than just `false` - the initialization did technically succeed, we just did fall back to other mode.

ddprof-lib/src/main/cpp/context_api.cpp
```diff
@@ -0,0 +1,266 @@
...
+            key_ptrs[i] = _attribute_keys[i];
+        }
+        key_ptrs[_attribute_key_count] = nullptr;
+
+        otel_thread_ctx_config_data config = {
+            .schema_version = "tlsdesc_v1_dev",
+            .attribute_key_map = key_ptrs,
+        };
+
+#ifndef OTEL_PROCESS_CTX_NO_READ
```
jbachorik: Why this ifdef?

ddprof-lib/src/main/java/com/datadoghq/profiler/ThreadContext.java
```diff
@@ -79,22 +99,79 @@ public ThreadContext(ByteBuffer buffer, int[] offsets) {
...
+
     public long getChecksum() {
         return buffer.getLong(checksumOffset);
     }
 
     public long put(long spanId, long rootSpanId) {
-        return useJNI ? setContext0(spanId, rootSpanId) : putContextJava(spanId, rootSpanId);
+        // OTEL mode must always use JNI to route through ContextApi::set()
+        // which writes to the OTEP record (ByteBuffer path only writes to Context struct)
+        return (useJNI || otelMode) ? setContext0(spanId, rootSpanId) : putContextJava(spanId, rootSpanId);
```
jbachorik: We should use just `putContextJava()` - we want to benchmark both modes and we should be comparing the same codepaths.

ddprof-lib/src/main/cpp/context_api.cpp
```diff
@@ -0,0 +1,169 @@
...
+    // Use atomic load for mode check - may be called from signal handlers
+    ContextStorageMode mode = __atomic_load_n(&_mode, __ATOMIC_ACQUIRE);
+
+    TEST_LOG("ContextApi::setOtel: tid=%d mode=%s trace_high=0x%llx trace_low=0x%llx span=0x%llx",
+             OS::threadId(), mode == CTX_STORAGE_OTEL ? "OTEL" : "PROFILER",
+             (unsigned long long)trace_id_high, (unsigned long long)trace_id_low,
+             (unsigned long long)span_id);
+
+    if (mode == CTX_STORAGE_OTEL) {
+        OtelContexts::set(trace_id_high, trace_id_low, span_id);
```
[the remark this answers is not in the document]
ivoanjo: Not for OTel in general, but it's interesting for Datadog, so having it as an extra attribute means at least in our profilers we can keep offering the features that rely on it.

(E.g. the OTel eBPF Profiler will gather this info and the Datadog backend will know what to do with it)

Same as for "trace endpoint" -- [this doc lists the key names I'm proposing we use for these, when we have them](https://docs.google.com/document/d/1IwjjVJzEChcFPcnVV2N5Kkjg-4_Q4v4Q3ojpxntbdvY/edit?pli=1&tab=t.efaosgjya44c#bookmark=id.700gvw31vb7h).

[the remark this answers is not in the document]
ivoanjo: If we don't keep local root span id here, would that mean we impact dd-trace-java profiles when running in OTel mode? E.g. we'd be missing the local root span?

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -1477,6 +1479,59 @@ void Recording::writeContext(Buffer *buf, Context &context) {
...
+
+  size_t numAttrs = Profiler::instance()->numContextAttributes();
+
+  if (ContextApi::getMode() == CTX_STORAGE_OTEL) {
+    // In OTEL mode: decode attrs_data from the OTEP record only.
+    u32 tag_values[DD_TAGS_CAPACITY] = {};
+
+    ProfiledThread* thrd = ProfiledThread::currentSignalSafe();
+    if (thrd != nullptr && thrd->isOtelContextInitialized()) {
+      OtelThreadContextRecord* record = thrd->getOtelContextRecord();
```
ivoanjo: I'm possibly missing some very subtle detail here -- why is "getting getting a pointer to current OTel context" different from the `Contexts::get()` from the other branch? I was kinda expecting that this part would look the same for both branches -- getting a pointer to "the context" and from there the difference is the format things are written down in the context.

ivoanjo (commented):
Here's my first pass of comments before lunch, be back soon!

ddprof-lib/src/main/cpp/context_api.cpp
```diff
@@ -0,0 +1,323 @@
...
+        free(key);
+    }
+
+    __atomic_store_n(&_mode, CTX_STORAGE_PROFILER, __ATOMIC_RELEASE);
+    __atomic_store_n(&_initialized, false, __ATOMIC_RELEASE);
+}
+
+bool ContextApi::isInitialized() {
+    return __atomic_load_n(&_initialized, __ATOMIC_ACQUIRE);
+}
```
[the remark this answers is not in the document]
jbachorik: Done — all methods now use `getMode()` consistently.

ivoanjo (commented):
Another round of comments, haven't gone through everything 😅 C++ is cool but hard :)

ddprof-lib/src/main/cpp/otel_context.cpp
```diff
@@ -0,0 +1,183 @@
...
+    if (record == nullptr) {
+        return;
+    }
+
+    // OTEP publication protocol:
+    // 1. Detach — external readers see no record during construction
+    __atomic_store_n(&custom_labels_current_set_v2, (OtelThreadContextRecord*)nullptr, __ATOMIC_RELEASE);
+
+    // 2. Invalidate — internal readers (signal handlers using cached pointer) see invalid record
+    __atomic_store_n(&record->valid, (uint8_t)0, __ATOMIC_RELEASE);
```
ivoanjo: So a small but important detail from the spec is that we don't need both of these here. Specifically, `valid == 0` is enough to indicate that a reader shouldn't use the the record, and we could move the set of `custom_labels_current_set_v2` directly into `initializeOtelTls` in `context_api.cpp`.

The `custom_labels_current_set_v2` vs `valid` thing is because the spec is on purpose written to give some leeway to runtimes. For instance, if we kept a `OtelThreadContextRecord` _per Java virtual thread_, we could for instance flip directly the `custom_labels_current_set_v2` when switching virtual threads.

But since we're tying the record to the `ProfiledThread` (which I think is the correct option in this situation), then we can initialize the thread-local once and from them on only flip valid to indicate that an update is going on.

ivoanjo (commented):
Gave it a full pass! Really excited to see how this is evolving! Overall I think there's a bit of code that we can rip out/simplify/improve :D

ddprof-lib/src/main/cpp/thread.cpp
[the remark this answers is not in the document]
jbachorik: Actually, this is not used any more. Very recently the TLS priming was removed but this got lost.
I will remove it in a separate PR as it would increase the noise in this one.

ddprof-lib/src/main/java/com/datadoghq/profiler/ThreadContext.java
```diff
@@ -17,121 +17,359 @@
...
+            for (int i = 0; i < valueLen; i++) {
+                recordBuffer.put(base + 2 + i, utf8[i]);
+            }
+            currentSize += entrySize;
+            recordBuffer.putShort(attrsDataSizeOffset, (short) currentSize);
+            return true;
+        }
+        recordBuffer.putShort(attrsDataSizeOffset, (short) currentSize);
+        return false;
+    }
```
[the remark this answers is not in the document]
jbachorik: You mean we would pad the value with spaces or 0 to fill the full 16 bytes?

ddprof-lib/src/main/java/com/datadoghq/profiler/ThreadContext.java
```diff
@@ -17,121 +17,360 @@
...
+    private final int validOffset;
+    private final int traceIdOffset;
     private final int spanIdOffset;
-    private final int rootSpanIdOffset;
-    private final int checksumOffset;
-    private final int customTagsOffset;
+    private final int attrsDataSizeOffset;
+    private final int attrsDataOffset;
+    private final int maxAttrsDataSize;
+    private final int lrsSidecarOffset; // localRootSpanId offset in sidecar
```
ivoanjo: Since these are fixed in the C code -- I wonder if we'd benefit from making them `static final` or somesuch -- make it clear to the JVM that these are always the same and do not vary per-thread.

(They are actually so fixed that we could hardcode the actual offsets here, we kinda know all of the constants)

ddprof-lib/src/main/java/com/datadoghq/profiler/ThreadContext.java
```diff
@@ -17,121 +17,400 @@
...
+            if (encoding < 0) {
+                // Dictionary full: clear sidecar AND remove the OTEP attrs_data entry
+                // so both views stay consistent (both report no value for this key).
+                int otepKeyIndex = keyIndex + 1;
+                detach();
+                BUFFER_WRITER.writeOrderedInt(sidecarBuffer, keyIndex * 4, 0);
+                removeOtepAttribute(otepKeyIndex);
+                attach();
+                return false;
+            }
```
ivoanjo: Minor: This is the same as `clearContextAttribute`

ivoanjo (commented):
Gave it another pass! :D

ddprof-lib/src/main/cpp/otel_process_ctx.h
jbachorik: Agreed — tracking this. Will open a PR to upstream once the attribute_key_map approach stabilizes. Happy if you want to poke at it.

ddprof-lib/src/main/java/com/datadoghq/profiler/ThreadContext.java
ivoanjo: Of course, up to you! But we could do a "hardcode and enforce in-sync" kinda thing.

That is -- we hardcode here the offsets and:

1. Have a test-only JNI method that also exposes the real offsets => Then add a test that checks both
2. Go even further and have a "self-test" somewhere in the code. E.g. at some point during profiler boot there's a bit of code that always runs, and if the offsets differ, it just stops any profiling right there (so basically take 1 and "test in production before starting the profiler")

This way we could still set things up so the JVM would know as much as possible, while avoiding the "oops we changed it and got to debug some fun corruption from a customer core dump".

ivoanjo (commented):
Looks very good! I think other than some small notes the main two changes to land this is a) the question of the `null` + `registerAttributeKeys`; b) the search and replace for `otel_thread_ctx_v1`.

ddprof-lib/src/main/cpp/otel_context.h
```diff
@@ -0,0 +1,88 @@
...
+public:
+    /**
+     * Read span_id from a record pointer (signal-safe).
+     * Checks the valid flag with acquire semantics before reading.
+     * Returns false if the record is invalid (being mutated). record must not be null.
+     */
+    static bool getSpanId(OtelThreadContextRecord* record, u64& span_id);
+
+    // Byte conversion helper (big-endian, W3C trace context)
+    static u64 bytesToU64(const uint8_t* in);
```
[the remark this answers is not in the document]
jbachorik: Good point, `getSpanId` is only used in `context_api.cpp`. Will move it to a static in `otel_context.cpp`.

ddprof-lib/src/main/java/com/datadoghq/profiler/BufferWriter8.java
```diff
@@ -56,12 +49,12 @@ public void writeInt(ByteBuffer buffer, int offset, int value) {
     }
 
     @Override
-    public void writeAndReleaseInt(ByteBuffer buffer, int offset, int value) {
-        UNSAFE.putIntVolatile(null, ((DirectBuffer) buffer).address() + offset, value);
+    public void fullFence() {
+        UNSAFE.fullFence();
     }
```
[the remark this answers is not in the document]
jbachorik: Push missing due to conflicts :/

ivoanjo (approved):
👍 Looks great to me! 

Thanks for sticking with this one -- really happy to see the current performance + I think looking at it as a whole the level of complexity needed to get to this point seems reasonable + if we ignored the markdown files this was net-neutral in terms of added/removed lines.

jbachorik:
@ivoanjo Thanks for the thorough review!

ivoanjo:
:pray: Hopefully we can dogfood the eBPF profiler side with this soon!

PR #348, opened 2026-01-29
Add alignment checks to NativeFunc mark methods

jbachorik:
**What does this PR do?**:

Fixes NativeFunc mark methods to prevent ASAN segfaults and adds CI workflow improvements for label-based test configuration selection.

**Core fix - NativeFunc marking:**
- Adds alignment validation to all NativeFunc mark-related methods (`isMarked`, `mark` getter/setter)
- Prevents segfaults when accessing marks on non-NativeFunc allocated strings
- Uses the same `is_aligned()` check pattern as `libIndex()`
- Invalid pointers now return safe defaults (false, 0, or no-op) instead of crashing
- Fixes inverted logic in `is_aligned()` helper
- Adds comprehensive tests for marking behavior

**CI workflow improvements:**
- Adds label-based test configuration: `test:release`, `test:asan`, `test:tsan`
- PRs always run `debug` config; labels add additional configurations
- 30-second debounce on label events allows adding multiple labels before run starts
- Fixes pre-existing bugs in `check-for-pr` job (broken jq expression)
- Removes dead `gh-release` job with invalid workflow syntax

**Motivation**:

ASAN tests revealed a segfault in `NativeFunc::mark()` at codeCache.h:77. The profiler was calling `mark()` on function names that weren't allocated via `NativeFunc::create()`, causing invalid memory access.

The CI changes enable easily running ASAN tests on PRs by adding the `test:asan` label, which was needed to properly validate this fix.

**How to test the change?**:

1. Run ASAN tests: `./gradlew :ddprof-test:testAsan`
2. Or add `test:asan` label to the PR to trigger ASAN CI run

The fix prevents crashes when the profiler encounters native frames from libraries without proper NativeFunc metadata.

**For Datadog employees**:

  credentials of any kind, I've requested a review from @DataDog/security-design-and-guidance.

### DIFF
--- a/.github/actions/upsert-pr-comment/action.yml
+++ b/.github/actions/upsert-pr-comment/action.yml
@@ -7,6 +7,10 @@ inputs:
   body-file:
     description: "Path to file whose contents become the comment body"
     required: true
+  comment-id:
+    description: "Unique identifier for this comment type (used to find/update existing comment)"
+    required: false
+    default: "upsert-pr-comment"
   repo:        # optional; defaults to triggering repo
     description: "Repository (owner/repo)."
     required: false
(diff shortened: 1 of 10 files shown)


zhengyu123 (dismissed):

rkennke (changes requested):
Looks good, but what about Zhengyu's suggestion?

rkennke (approved):
Thanks! Looks good!

PR #350, opened 2026-01-30
Possible using duplicated keys to access MethodInfo hash table

zhengyu123:
**What does this PR do?**:
Improves memory safety in signal handler contexts and hardens frame type handling for J9/OpenJ9 JVMs:
- Uses `SafeAccess` for `NativeFunc::read_mark()` to prevent SIGSEGV in signal handlers
- Uses `SafeAccess` for frame pointer dereference in `walkFP` stack walking
- Sanitizes J9 frame types to prevent invalid `FrameTypeId` values from corrupted JVMTI data
- Clamps `FrameType::decode()` to valid range to defend against corrupted BCI values

**Motivation**:
Signal handlers must not crash on invalid memory access. The previous direct struct field access in `read_mark()` and `walkFP` could cause crashes if the pointer was invalid. Additionally, J9's JVMTI extension can return unexpected frame type values that need sanitization before use.

**Additional Notes**:
- Also fixes NPE and timing issues in flaky profiler tests (`JfrDumpTest`, `BaseContextWallClockTest`)
- Adds J9/OpenJ9 JDK support to `run-docker-tests.sh` (8-j9, 11-j9, 17-j9, 21-j9 variants)
- Version bump to 1.38.0

**How to test the change?**:
- Existing unit tests cover the functionality
- `gtestDebug` includes `NativeFunc` tests for `read_mark()` behavior
- J9 tests can be run via `./utils/run-docker-tests.sh --jdk=21-j9`

**For Datadog employees**:
- [ ] JIRA: [PROF-13622](https://datadoghq.atlassian.net/browse/PROF-13622)

[PROF-13622]: https://datadoghq.atlassian.net/browse/PROF-13622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/flightRecorder.h
+++ b/ddprof-lib/src/main/cpp/flightRecorder.h
@@ -109,6 +109,15 @@ public:
 #define HIGHEST_BIT_MASK 0x8000000000000000ULL
 #define HIGHEST_2_BITS_MASK 0xC000000000000000ULL
 
+// MethodMap's key can be derived from 3 sources:
+// 1) jmethodID for Java methods
+// 2) void* address for native method names
+// 3) Encoded RemoteFrameInfo
+// The values of the keys are potentially overlapping, so we use 
+// the highest 2 bits to distinguish them.
+// 00 - jmethodID
+// 10 - void* address
+// 11 - RemoteFrameInfo
 class MethodMap : public std::map<unsigned long, MethodInfo> {
 public:
   MethodMap() {}
@@ -122,7 +131,7 @@ public:
     return ((unsigned long)addr | HIGHEST_BIT_MASK);
   }
 
-  static unsigned long makeRemoteKey(unsigned long key) {
+  static unsigned long makeKey(unsigned long key) {
     assert((key & HIGHEST_2_BITS_MASK) == 0);
     return (key | HIGHEST_2_BITS_MASK);
   }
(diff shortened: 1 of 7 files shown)


ddprof-lib/src/main/cpp/flightRecorder.h
```diff
@@ -106,9 +106,35 @@ class MethodInfo {
...
+// the highest 2 bits to distinguish them.
+// 00 - jmethodID
+// 10 - void* address
+// 11 - RemoteFrameInfo
+class MethodMap : public std::map<unsigned long, MethodInfo> {
 public:
   MethodMap() {}
+
+  static unsigned long makeKey(jmethodID method) {
+    return (unsigned long)method;
```
jbachorik: Should this method clear the highest 2 bits?
I can imagine that if jmethodid comes with any of the highest 2 bits set, it will confuse the key to be considered either plain function address key or a remote symbol key?

jbachorik (changes requested):
Please, add regtets to the gtest suite

ddprof-lib/src/main/cpp/flightRecorder.h
zhengyu123: It is not possible - jmethodID is generated from a monotonic counter, to reach 2^62 probably will take a few centuries :-)

jbachorik (approved):
Nice cleanup!

LGTM!

PR #358, opened 2026-02-06
Migrate build system to Kotlin build-logic

jbachorik:
## Summary

Migrates the native build system from `buildSrc` to a Gradle composite build (`build-logic/`) with modern Kotlin DSL plugins. This provides better IDE support, faster incremental builds, and a cleaner architecture.

**Key changes:**
- Replace `buildSrc/` with `build-logic/conventions/` composite build
- Convert all build scripts from Groovy to Kotlin DSL (`.gradle` → `.gradle.kts`)
- Create reusable Gradle plugins: `NativeBuildPlugin`, `GtestPlugin`, `ScanBuildPlugin`
- Add automatic musl libc detection with `-D__musl__` compiler flag
- Add comprehensive documentation in `build-logic/README.md` and `QUICKSTART.md`
- Create tooling-agnostic `AGENTS.md` for AI coding assistants
- Reorganize documentation with PascalCase naming and clear directory structure

**Build system improvements:**
- Platform-aware native compilation (Linux/macOS, x64/arm64)
- Configurable build configurations (release, debug, asan, tsan, fuzzer)
- Automatic debug symbol extraction for release builds
- Google Test integration with per-test task granularity
- scan-build static analysis integration (Linux only)

**Documentation improvements:**
- Reorganized `doc/` into `architecture/`, `build/`, `reference/`, `temp/` directories
- Standardized all docs to PascalCase naming
- Added `doc/README.md` navigation index
- Added PascalCase naming rule to `AGENTS.md`

## Test plan

- [x] Build compiles on macOS arm64
- [x] Build compiles on Linux x64
- [x] Build compiles on Linux arm64
- [x] Build compiles on Linux x64 musl
- [x] All Java tests pass (`testDebug`)
- [x] All C++ tests pass (`gtestDebug`)
- [x] CI pipeline passes

```bash
# Local testing
./gradlew buildDebug
./gradlew testDebug
./gradlew gtestDebug
```


rkennke (commented):
So wait a minute...
Without having looked at much of the changes (which seems many):
- Now I need to learn Kotlin, in addition to Groovy, for the build system?
- How does it improve IDE support? I don't think my IDE (CLion) deals very well with Gradle, let alone Gradle with Kotlin stuff? (Although, it might actually like the Kotlin parts, CLion comes from the same company that invented Kotlin, after all).

jbachorik:
@rkennke There is no Groovy left. 
The main issue is that using groovy and `buildSrc` is 'old/deprecated` and any bugfixing/improving effort goes to the kotlin based stuff.

And we want to use the plugins because it simplifies the actual build files significantly :(

Eg. to completely setup the native build you just need to do 
```
// Configure native build with the new plugin
nativeBuild {
  version.set(componentVersion)
  cppSourceDirs.set(listOf("src/main/cpp"))
  includeDirectories.set(
    listOf(
      "src/main/cpp",
      "${project(":malloc-shim").file("src/main/public")}"
    )
  )
}
```

No fancy gradle magic stealing configurations from other tasks etc. 

I am planning to progressively move the logic that is not strictly related to configuring the build up to the plugins so the only thing you would need to do to maintain the build would be pretty small and declarative.

And, last but not least, dd-trace-java is also moving to Kotlin and since we cooperate with that project we need to be at least aware of the Kotlin syntax (but the codegen tools can take a brunt - the AGENTS.md is set up in a way that it should help with the build maintenance)


---
> How does it improve IDE support? 

Well, IntelliJ seems to be a lot snappier with this new config. Also, the build _feels_ like starting faster ...

rkennke (approved):
Looks good!

PR #363, opened 2026-02-10
Safer type casting

zhengyu123:
**What does this PR do?**:
- Capture uninitialized field offsets
- Capture invalid memory accesses
- Capture bad type castings
- Cleanup, remove unused code
- Renamed some classes to avoid name collisions with JVM, this improves debugability.  

**Motivation**:
Profiler code heavily depends on `vmStructs` to decode hotspot JVM internal data structures, but `vmStructs` evolves version-by-version, that can results in misinterpreting the data structures and leads to crashes.

This PR is intended to capture the problem early, by:
- check for uninitialized field offsets - likely due to `vmStructs` name changes, etc.
- check underlaid memory is mapped before access.
- check memory range that covers the destination type, is valid before casts to destination type. 
- cleanup: removed unused code, such as:
  `permGen` - no long presented in supported JVM
  `CollectedHeap` - no use
- use macro to generate type declarations and corresponding initializations to avoid boilerplate code.

**Additional Notes**:
This PR also provides a way to mitigate the tedious tasks to define `cast` function for a type, which you have to:
  - define a type size variable
  - initialize the size variable
  - populate the size from `vmStructs`
  - define `cast` and `load_then_cast` functions for the type

This PR introduces a macro:
```
#define DECL_TYPES_DO(f) \
    f(VMClassLoaderData,    MATCH_SYMBOLS("ClassLoaderData")) \
    f(VMConstantPool,       MATCH_SYMBOLS("ConstantPool")) \
    f(VMConstMethod,        MATCH_SYMBOLS("ConstMethod")) \
    f(VMFlag,               MATCH_SYMBOLS("JVMFlag", "Flag")) \
    f(VMJavaFrameAnchor,    MATCH_SYMBOLS("JavaFrameAnchor")) \
    f(VMKlass,              MATCH_SYMBOLS("Klass")) \
    f(VMMethod,             MATCH_SYMBOLS("Method")) \
    f(VMNMethod,            MATCH_SYMBOLS("nmethod")) \
    f(VMSymbol,             MATCH_SYMBOLS("Symbol")) \
    f(VMThread,             MATCH_SYMBOLS("Thread"))
```
For each row, it defines a type and the symbol(s) used in `vmStructs` to populate type size. Notice, there can be more than one symbol for the type, due to different JVM versions that export the same value under different names.

To add a new type, you only need to add a row, then use following macros to define the type in profiler, e.g.
```
DECL_TYPE(VMClassLoaderData)
....
DECL_TYPE_END
```
The definition automatically creates three methods, with safety checks:
```
        // size of the type - sizeof(T)
        static uint64_t type_size();
        // cast from a pointer
        static name * cast(const void* ptr); 
        // load a pointer from given location, then cast the pointer
        static name * load_then_cast(const void* ptr);

```

**How to test the change?**:
Running tests on different version of JDKs, make sure that newly added checks do not fail.

**For Datadog employees**:
- [X] JIRA: [PROF-13721](https://datadoghq.atlassian.net/browse/PROF-13721)

[PROF-13721]: https://datadoghq.atlassian.net/browse/PROF-13721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddprof-lib/src/main/cpp/vmStructs.inline.h
```diff
@@ -0,0 +1,27 @@
...
+#include "vmStructs.h"
+
+VMNMethod* VMMethod::code() {
+    assert(_method_code_offset >= 0);
+    const void* code_ptr = *(const void**) at(_method_code_offset);
+    return VMNMethod::cast(code_ptr);
+}
+
+VMMethod* VMThread::compiledMethod() {
+    assert(_thread_vframe_offset >= 0);
```
jbachorik: `_thread_vframe_offset` is not used here - why the assert?

ddprof-lib/src/main/cpp/vmStructs.h
```diff
@@ -19,6 +19,44 @@
 class GCHeapSummary;
 class HeapUsage;
 
+#define TYPE_SIZE_NAME(name)    _##name##_size
+#define DECL_TYPE_SIZE(name) static uint64_t TYPE_SIZE_NAME(name);
```
jbachorik: Is this different to `DECL_TYPE_SIZE_VAR`? If not, can we have just one?
Otherwise, this macro seems to be unused.

ddprof-lib/src/main/cpp/vmStructs.h
```diff
@@ -19,6 +19,44 @@
...
+#define SIZE_OF(name) VM##name::type_size()
```
jbachorik: Unused macro?

jbachorik (changes requested):

jbachorik (changes requested):
Good effort to get the vmstructs more organized. I wonder, if we can get to a state where the VMStructs class does not need to hold all possible offset holders.

I would really appreciate more cleanup before requesting review. There was a bunch of low-hanging typos and miss-spels here :/

jbachorik (approved):
Conceptually I am fine with this.
Please, consider the nits before merging, but I would be fine with the current shape as well.

PR #407, opened 2026-03-05
Fix sanitizer errors in stack walker and surface sanitizer logs in tests

jbachorik:
**What does this PR do?**:

- Adds `NOADDRSANITIZE` to `VMJavaFrameAnchor` methods (`getFrame`, `lastJavaSP`, `lastJavaFP`, `lastJavaPC`, `fromEntryFrame`) and `VMThread::anchor()` — the `no_sanitize("address")` attribute on `walkVM` doesn't propagate to callees, so ASAN instruments them and flags accesses to JVM-internal memory.
- Fixes a UBSAN null-pointer dereference at `stackWalker.cpp:793` where `anchor->getFrame()` is called after `anchor` was set to NULL by the interpreter frame fallback path.
- Suppresses UBSAN alignment errors in aarch64 `StackFrame` methods (`adjustSP`, `unwindStub`, `unwindCompiled`, `unwindPrologue`, `unwindEpilogue`, `checkInterruptedSyscall`) that dereference `instruction_t*` (4-byte aligned) from potentially misaligned PC values obtained from interrupted signal contexts.
- Surfaces sanitizer log file contents (both ASAN and UBSAN) in test output by reading new entries from log files after each test and printing them to stderr.
- Fixes javadoc task missing dependency on `copyExternalLibs`.
- Disables ASAN signal interception (`handle_segv=0,handle_sigfpe=0`) to avoid conflicts with JVM's own signal handlers.
- Replaces `sameStack` heuristic in `checkFault()` with `crashProtectionActive` flag as the preferred crash recovery check, with `sameStack` retained as a fallback for cases where `ProfiledThread` TLS is unavailable (early init, crash recovery tests).
- Protects raw pointer dereferences in `VMMethod::id()` with `SafeAccess` calls to prevent SIGSEGV when GC/class unloading frees the ConstantPool or holder during signal-handler stack walking.

**Motivation**:

Nightly CI sanitizer builds on GraalVM 17 aarch64 crash during wall-clock profiling. Multiple root causes:

1. ASAN instruments `VMJavaFrameAnchor` methods that access JVM-internal memory, triggering `DEADLYSIGNAL`.
2. UBSAN detects misaligned loads in aarch64 `StackFrame` methods when the profiler's signal handler interrupts a JIT compiler thread with a non-4-byte-aligned PC.
3. UBSAN null-pointer dereference when `anchor->getFrame()` is called after `anchor` was set to NULL.
4. ASAN's default signal interception (`handle_segv=1`) conflicts with JVM signal handlers.
5. `checkFault()` uses a `sameStack` heuristic with an 8KB threshold that fails when ASAN redzones + aarch64 kernel signal frames inflate the stack distance beyond 8KB, preventing crash recovery via `longjmp`.
6. `VMMethod::id()` has unprotected raw dereferences that SIGSEGV when GC/class unloading frees memory concurrently.

**Additional Notes**:

- `checkFault()` now uses OR logic: prefer the semantic `crashProtectionActive` flag (reliable regardless of stack frame sizes, handles ASAN builds), fall back to the `sameStack` heuristic when `ProfiledThread` TLS is not set up (e.g. early init or crash recovery tests where `currentSignalSafe()` returns null).
- On aarch64, `instruction_t` is `unsigned int` (4 bytes). When a wall-clock signal interrupts a thread, the captured PC may point to arbitrary memory, leading to misaligned dereferences that are harmless on hardware but flagged by UBSAN.
- The sanitizer log surfacing parses both `ASAN_OPTIONS` and `UBSAN_OPTIONS` for `log_path`, tracks file sizes before/after each test, and prints only new content so errors are attributed to the test that triggered them.

**How to test the change?**:

- `./utils/run-docker-tests.sh --config=asan --jdk=21` — regression check
- `./utils/run-docker-tests.sh --config=release --jdk=21` — regression check for non-ASAN builds

**For Datadog employees**:
- [x] JIRA: [PROF-13925]

[PROF-13925]: https://datadoghq.atlassian.net/browse/PROF-13925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/build-logic/conventions/src/main/kotlin/com/datadoghq/native/config/ConfigurationPresets.kt
+++ b/build-logic/conventions/src/main/kotlin/com/datadoghq/native/config/ConfigurationPresets.kt
@@ -205,7 +205,7 @@ object ConfigurationPresets {
                 if (libasan != null) {
                     config.testEnvironment.apply {
                         put("LD_PRELOAD", libasan)
-                        put("ASAN_OPTIONS", "allocator_may_return_null=1:unwind_abort_on_malloc=1:use_sigaltstack=0:detect_stack_use_after_return=0:handle_segv=1:halt_on_error=0:abort_on_error=0:print_stacktrace=1:symbolize=1:log_path=/tmp/asan_%p.log:suppressions=$rootDir/gradle/sanitizers/asan.supp")
+                        put("ASAN_OPTIONS", "allocator_may_return_null=1:unwind_abort_on_malloc=1:use_sigaltstack=0:detect_stack_use_after_return=0:handle_segv=0:halt_on_error=0:abort_on_error=0:print_stacktrace=1:symbolize=1:log_path=/tmp/asan_%p.log:suppressions=$rootDir/gradle/sanitizers/asan.supp")
                         put("UBSAN_OPTIONS", "halt_on_error=0:abort_on_error=0:print_stacktrace=1:log_path=/tmp/ubsan_%p.log:suppressions=$rootDir/gradle/sanitizers/ubsan.supp")
                         put("LSAN_OPTIONS", "detect_leaks=0")
                     }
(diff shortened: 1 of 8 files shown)


rkennke (approved):
Ok!

PR #420, opened 2026-03-17
Intercept sigaction to prevent libraries from overwriting signal handlers

jbachorik:
## Summary

- Intercept `sigaction()` calls via GOT patching to prevent other libraries from overwriting our SIGSEGV/SIGBUS handlers
- Store intercepted handlers as chain targets and call them from our handlers
- Protects against libraries like wasmtime that install broken signal handlers calling `malloc()` (not async-signal-safe)

## Problem

Wasmtime's SIGSEGV handler calls `__tls_get_addr` which can call `malloc()`, violating async-signal-safety. When the profiler uses `safefetch`, this causes deadlocks:

1. Application code holds malloc's lock
2. Profiler's signal handler runs and calls `safefetch`
3. `safefetch` triggers SIGSEGV
4. Wasmtime's handler runs and calls `malloc()` → deadlock

(see https://github.com/bytecodealliance/wasmtime/issues/12787 wasmtime issue)

## Solution

- Patch `sigaction` GOT entry in all loaded libraries
- Intercept SIGSEGV/SIGBUS handler installations (SA_SIGINFO only)
- Keep our handler on top, store their handler as chain target
- Call chain target from our handler after our logic completes

## Test plan

- [x] Verify profiler works with wasmtime-based applications without deadlocking
- [x] Verify signal handler chaining works correctly
- [x] Run existing test suite

### DIFF
--- a/ddprof-lib/src/main/cpp/codeCache.cpp
+++ b/ddprof-lib/src/main/cpp/codeCache.cpp
@@ -338,6 +338,11 @@ void CodeCache::addImport(void **entry, const char *name) {
               saveImport(im_realloc, entry);
           }
           break;
+      case 's':
+          if (strcmp(name, "sigaction") == 0) {
+              saveImport(im_sigaction, entry);
+          }
+          break;
     }
 }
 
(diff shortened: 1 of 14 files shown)


rkennke (commented):
Looks good overall (well as good as patching-out call addresses can look, I suppose), I just have two questions.

rkennke (approved):
Ok, thanks for the explanations. Feel free to add a warning, if you like.

jbachorik:
Yes, added warning and also a counter that would tell us how many libraries we missed due to overflow.

PR #340, opened 2026-01-23
[PROF-13528] Fold _dd files back into their upstream counterparts

rkennke:
**What does this PR do?**:
Merge/consolidate all our Datadog adapters (`*_dd*` files) back into their corresponding upstream source files.

**Motivation**:
Simplify our development and maintenance of async-profiler-derived code.

**Additional Notes**:
I did the change in steps, it might be more useful to look at the commits one-by-one.

**How to test the change?**:
I have run testDebug and testRelease with no regressions.

**For Datadog employees**:
- [x] If this PR touches code that signs or publishes builds or packages, or handles
  credentials of any kind, I've requested a review from `@DataDog/security-design-and-guidance`.
- [x] This PR doesn't touch any of that.
- [x] JIRA: [PROF-13528]

[PROF-13528]: https://datadoghq.atlassian.net/browse/PROF-13528?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #199, opened 2025-03-25
A set of fixes related to clang, aarch64 and musl pecularities of vmstructs stack unwinder

jbachorik:
**What does this PR do?**:
It addresses the shortcomings of the current implementation of the vmstructs stack walker on aarch64, clang and musl and their combinations.

**Motivation**:
Iron out the last wrinkles on the vmstructs based stack walker such we can even consider using it as default.

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:
- [ ] If this PR touches code that signs or publishes builds or packages, or handles
  credentials of any kind, I've requested a review from `@DataDog/security-design-and-guidance`.
- [x] This PR doesn't touch any of that.
- [x] JIRA: [PROF-11294], [PROF-11299]

Unsure? Have a question? Request a review!

[PROF-11294]: https://datadoghq.atlassian.net/browse/PROF-11294?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PROF-11299]: https://datadoghq.atlassian.net/browse/PROF-11299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #426, opened 2026-03-19
Detect per-library aarch64 frame layout from DWARF

jbachorik:
**What does this PR do?**:
Wires up the previously-defined-but-unused `FrameDesc::default_clang_frame` by detecting the correct linked-frame size per library from actual DWARF FDE entries, and using it as the fallback when a PC has no DWARF coverage.

**Motivation**:
On aarch64, GCC and Clang both produce standard frames (`stp x29, x30, [sp, #-16]!; mov x29, sp`) with `CFA = FP+16` in their DWARF FDE bodies. The existing fallback `default_frame` encoded `CFA = FP+0` (with a special-case path in the unwinder), which was a heuristic gamble. `default_clang_frame` with `CFA = FP+16` is the ABI-compliant encoding that correctly advances SP to the pre-frame value. The per-library detection was intended when `default_clang_frame` was introduced in #199 but was never implemented — it was lost during the `_dd` file consolidation in #340.

**Additional Notes**:

I am quite sure we already had this functionality before. Somehow got lost when merging with the upstream and redoing the source structure :/

- Detection fires on the first FDE entry with `cfa_reg == DW_REG_FP && cfa_off > 0` during DWARF parsing — terminal records (which use the hardcoded `LINKED_FRAME_SIZE = 0` on aarch64) are correctly skipped by the `cfa_off > 0` guard.
- `detectedDefaultFrame()` only returns `default_clang_frame` when `LINKED_FRAME_CLANG_SIZE != LINKED_FRAME_SIZE`, which is aarch64 only. On x86_64/i386 the two constants are equal so behaviour is unchanged.
- For the `cc == NULL` fallback (PC in a completely unknown library), macOS/aarch64 uses `default_clang_frame` unconditionally since Apple ships Clang-only code.
- `_default_frame` always points to a static `FrameDesc` — no lifetime or thread-safety concern. Verified by review.

**How to test the change?**:
Existing unwinding integration tests cover the aarch64 unwinding paths. The change is conservative: libraries with DWARF info already unwind correctly via the table; this only affects the rare fallback path (PC outside DWARF coverage or library has no `.eh_frame`). Testing on Apple Silicon (macOS/aarch64) and Linux/aarch64 with both GCC- and Clang-compiled JDKs exercises the detection path.

**For Datadog employees**:
- [ ] JIRA: [JIRA-XXXX]

### DIFF
--- a/ddprof-lib/src/main/cpp/codeCache.cpp
+++ b/ddprof-lib/src/main/cpp/codeCache.cpp
@@ -63,6 +63,7 @@ CodeCache::CodeCache(const char *name, short lib_index,
 
   _dwarf_table = NULL;
   _dwarf_table_length = 0;
+  _default_frame = &FrameDesc::default_frame;
 
   _capacity = INITIAL_CODE_CACHE_CAPACITY;
   _count = 0;
@@ -101,6 +102,7 @@ void CodeCache::copyFrom(const CodeCache& other) {
   _dwarf_table = new FrameDesc[_dwarf_table_length];
   memcpy(_dwarf_table, other._dwarf_table,
          _dwarf_table_length * sizeof(FrameDesc));
+  _default_frame = other._default_frame;
 
   _capacity = other._capacity;
   _count = other._count;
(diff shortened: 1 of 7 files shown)


kaahos (commented):
Hi, this PR seems good to me. I included a question/thought which, while not being directly linked to what the PR description involves, is still related.

This is my first review so it might be worth to get the approval of someone else.

rkennke (approved):
Ok looks good, thank you!

PR #430, opened 2026-03-22
Parse __eh_frame DWARF info on macOS for per-PC frame descriptions

jbachorik:
## What

Parse raw `__eh_frame` DWARF CFI data on macOS to provide per-PC frame descriptions for GCC-built libraries, improving stack unwinding accuracy for functions with non-standard frame layouts (leaf functions, `-fomit-frame-pointer`).

### `ddprof-lib/src/main/cpp/dwarf.h` / `dwarf.cpp`
- Add new `DwarfParser` constructor accepting a raw `(eh_frame, size)` pair (no binary-search index).
- Implement `parseEhFrame()`: linearly iterates CIE/FDE records, extracts code/data alignment from each CIE, decodes per-PC frame descriptions from FDEs, then `qsort`s the result.
- Add `_has_z_augmentation` field: set when the CIE augmentation string starts with `'z'`. The FDE augmentation-data-length LEB field is only present in that case; guarding the skip prevents corrupting the instruction stream for CIEs with empty or non-`z` augmentation. The parser assumes a single CIE per module — the DWARF spec allows multiple CIEs with different augmentation strings, but macOS binaries compiled by clang always emit one CIE per module and this path is only exercised for macOS `__eh_frame` sections.
- Pass `detectedDefaultFrame()` to `setDwarfTable()`, matching the Linux path and correctly selecting the clang frame layout on aarch64.
- Add bounds check in CIE parsing before `getLeb()`/`getSLeb()` calls to guard against malformed or truncated `__eh_frame` data.

### `ddprof-lib/src/main/cpp/symbols_macos.cpp`
- Detect `__eh_frame` section in the `__TEXT` segment during Mach-O parsing.
- When `DWARF_SUPPORTED` and the section is present, construct a `DwarfParser` and call `cc->setDwarfTable()` with the detected default frame.
- Add missing `Symbols::clearParsingCaches()` definition (fixes linker error on macOS).

### `ddprof-lib/src/test/cpp/dwarf_ut.cpp` (new)
- Four gtest unit tests covering `parseEhFrame()`: empty section, terminator-only, CIE-only, and CIE+FDE.

## Why

On macOS, GCC-built shared libraries (`libgcc`, Homebrew packages) emit `__eh_frame` with standard DWARF CFI, but no `__unwind_info`. Previously the profiler fell back to a single library-wide heuristic frame layout for these libraries. Parsing the DWARF FDE table yields exact CFA/FP/PC rules per instruction address, eliminating unwinding failures at function prologues and epilogues.

JIRA: [PROF-14088]

[PROF-14088]: https://datadoghq.atlassian.net/browse/PROF-14088?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddprof-lib/src/main/cpp/codeCache.cpp
```diff
@@ -99,7 +99,7 @@ void CodeCache::copyFrom(const CodeCache& other) {
   _imports_patchable = other._imports_patchable;
 
   _dwarf_table_length = other._dwarf_table_length;
-  _dwarf_table = new FrameDesc[_dwarf_table_length];
+  _dwarf_table = (FrameDesc*)malloc(_dwarf_table_length * sizeof(FrameDesc));
```
zhengyu123: When _dwarf_table_length == 0, malloc(0) is impl-defined and memcpy(NULL, NULL, 0) is UB per C standard. Guard with if (_dwarf_table_length > 0) or at least assert.

ddprof-lib/src/main/cpp/dwarf.cpp
```diff
@@ -144,6 +155,93 @@ void DwarfParser::parse(const char *eh_frame_hdr) {
...
+      u32 range_start = (u32)(getPtr() - _image_base);
+      u32 range_len = get32();
+      if (_has_z_augmentation) {
+        _ptr += getLeb(record_end);  // getLeb reads the length; advance past the augmentation data bytes
+        if (_ptr > record_end) {
+          break;
+        }
+      }
+      parseInstructions(range_start, record_end);
+      addRecord(range_start + range_len, DW_REG_FP, LINKED_FRAME_SIZE,
```
zhengyu123: The sentinel addRecord at end of each FDE always uses LINKED_FRAME_SIZE constants. On aarch64 with clang (LINKED_FRAME_CLANG_SIZE=16), this resets to a GCC-style layout — inconsistent for a parser explicitly targeting macOS clang output.

ddprof-lib/src/main/cpp/dwarf.h
```diff
@@ -152,6 +153,18 @@ class DwarfParser {
         }
     }
 
+    u32 getLeb(const char* end) {
+        u32 result = 0;
+        for (u32 shift = 0; _ptr < end; shift += 7) {
+            u8 b = *_ptr++;
+            result |= (b & 0x7f) << shift;
```
zhengyu123: Could this shift result in overflow?

zhengyu123 (changes requested):

ddprof-lib/src/main/cpp/codeCache.cpp
jbachorik: Fixed — `malloc` and `memcpy` are now guarded by `if (_dwarf_table_length > 0)`, with `_dwarf_table = nullptr` in the else branch.

ddprof-lib/src/main/cpp/dwarf.cpp
jbachorik: Fixed — the sentinel `addRecord` now uses `LINKED_FRAME_CLANG_SIZE` instead of `LINKED_FRAME_SIZE`. On x86_64 these are equal so no behavioral change there; on aarch64 this correctly uses 16 instead of 0.

ddprof-lib/src/main/cpp/dwarf.h
jbachorik: Yes — `(b & 0x7f)` promotes to `int` and shifting by 28 overflows signed int (UB). Fixed by casting to `(u32)` before the shift and adding `shift < 32` to the loop condition to handle malformed data with excess continuation bytes.

zhengyu123 (approved):
LGTM

PR #437, opened 2026-03-24
fix(sigaction): prevent infinite loop in signal handler chaining

jbachorik:
**What does this PR do?**:
Fix infinite loop bug in signal handler chaining when wasmtime (or similar libraries) is present.

When intercepting sigaction calls from other libraries, we were returning our handler as `oldact`, causing an infinite loop:
```
profiler -> wasmtime -> profiler -> wasmtime -> ...
```

Now we save and return the original (JVM's) handler, so the chain terminates correctly:
```
profiler -> wasmtime -> JVM
```

Also fixes an intermittent SIGABRT in debug builds on aarch64 GraalVM: `crashProtectionActive()` now falls back to `VMThread::isExceptionActive()` for threads without `ProfiledThread` TLS, so the `cast_to()` debug assert no longer fires when setjmp crash protection is already active in `walkVM`.

**Motivation**:
Gradle builds were hanging with 94% CPU usage in the dd-trace-processor thread, stuck in an infinite signal handler loop when wasmtime was present. Intermittent SIGABRT was seen on `test-linux-glibc-aarch64 (17-graal, debug)` CI jobs.

**Additional Notes**:
Changes:
- Save original handlers in `protectSignalHandlers()` BEFORE installing ours
- Return saved JVM handlers as `oldact` in `sigaction_hook()` for both install and query-only calls
- Extend `crashProtectionActive()` with `VMThread::isExceptionActive()` fallback
- Add `sigaction_interception_ut.cpp` test to catch this bug
- Add `OS::resetSignalHandlersForTesting()` for test isolation
- Wire gtest to run before Java tests in `testDebug`
- Remove broken `test_tlsPriming.cpp` (referenced removed APIs)
- Add `getSigactionHook()` stub for macOS
- Add ASCII flow diagram to `sigaction_hook` documenting the handler chain

**How to test the change?**:
- New unit test `sigaction_interception_ut` validates the fix
- Run `./gradlew :ddprof-lib:gtestDebug` - all tests pass
- Run `./gradlew :ddprof-test:testDebug` - gtest now runs before Java tests

**For Datadog employees**:
- [x] JIRA: [PROF-14144](https://datadoghq.atlassian.net/browse/PROF-14144)

---
🤖 Generated with Claude Code

Co-Authored-By: Claude Opus 4.5 <noreply@anthropic.com>

[PROF-14144]: https://datadoghq.atlassian.net/browse/PROF-14144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rkennke (approved):
Looks good.

PR #459, opened 2026-04-10
Fix signal-handler crash in Libraries::findLibraryByAddress

jbachorik:
**What does this PR do?**:

Fixes a crash/hang in `Libraries::findLibraryByAddress` triggered when a profiler signal fires on a thread whose thread-local storage (TLS) block has not been initialized yet.

The fix extracts the lookup algorithm into `findLibraryImpl.h` as a signal-handler-safe template using a plain `static volatile int` last-hit index instead of any `thread_local` storage.

| File | Change |
|------|--------|
| `ddprof-lib/src/main/cpp/findLibraryImpl.h` | New: signal-safe template `findLibraryByAddressImpl` with `static volatile int` last-hit cache |
| `ddprof-lib/src/main/cpp/libraries.cpp` | Delegate `findLibraryByAddress` to the template (one-liner) |
| `ddprof-lib/src/test/cpp/libraries_ut.cpp` | New: unit tests covering known/null/invalid address and cache hit/miss paths |

**Motivation**:

DTLS initialisation for shared libraries calls `calloc` internally. If a profiler signal fires on a thread whose TLS block has not been set up yet while that thread is inside `malloc`, any `thread_local` access in the signal-handler call chain deadlocks on the allocator lock — manifesting as a crash in `Libraries::findLibraryByAddress`. The `static volatile int` last-hit index avoids TLS entirely: a benign race on the index is acceptable — worst case is a cache miss that falls through to the O(n) linear scan.

The template is also reusable by benchmarks (hotPathBenchmark.cpp can instantiate it with lightweight fake types) without pulling in JVM headers.

Fixes [PROF-13630](https://datadoghq.atlassian.net/browse/PROF-13630).

**Additional Notes**:

The signal-safety guarantee depends on two properties:
1. `static volatile int` reads/writes do not require TLS initialization — there is no DTLS access on any supported target (linux-x64, linux-arm64, macos-arm64).
2. `CodeCacheArray::count()` and `operator[]` use `__atomic_load_n` compiler builtins that expand to a single load instruction — these are signal-safe on glibc, musl, and macOS.

A stale `last_hit` index (e.g., after a concurrent `updateSymbols` adds more libraries) simply falls through to the linear scan. Since `CodeCacheArray` never removes entries, `last_hit < count()` is the only safety check needed.

**How to test the change?**:

- `./gradlew :ddprof-lib:buildDebug` — must compile cleanly
- `./gradlew :ddprof-lib:testDebug` (Linux/musl CI) — new `LibrariesTest.*` cases exercise known/null/invalid addresses and cache hit/miss paths
- Signal-safety under uninitialized TLS cannot be automated in GTest; verified by code inspection (no `thread_local` in signal-handler paths).

**For Datadog employees**:

- [x] JIRA: [PROF-13630](https://datadoghq.atlassian.net/browse/PROF-13630)

[PROF-13630]: https://datadoghq.atlassian.net/browse/PROF-13630?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/findLibraryImpl.h
+++ b/ddprof-lib/src/main/cpp/findLibraryImpl.h
@@ -28,7 +28,7 @@
 // that falls through to the O(n) linear scan.
 
 template<typename CacheEntry, typename CacheArray>
-static inline CacheEntry* findLibraryByAddressImpl(const CacheArray& libs, const void* address) {
+inline CacheEntry* findLibraryByAddressImpl(const CacheArray& libs, const void* address) {
     static volatile int last_hit = 0;
     const int count = libs.count();
     int hint = last_hit;
--- a/ddprof-lib/src/main/cpp/findLibraryImpl.h
+++ b/ddprof-lib/src/main/cpp/findLibraryImpl.h
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2026, Datadog, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef _FINDLIBRARYIMPL_H
+#define _FINDLIBRARYIMPL_H
+
+// Signal-handler-safe last-hit-index cache for findLibraryByAddress.
+//
+// Templated on CacheArray and CacheEntry so that the exact same algorithm
+// can be exercised from benchmarks (using lightweight fake types) without
+// pulling in JVM headers, while production uses CodeCacheArray/CodeCache.
+//
+// Requirements on CacheArray:
+//   int          count()     const  — number of live entries
+//   CacheEntry*  operator[](int i) const  — entry at index i (may be nullptr)
+//
+// Requirements on CacheEntry:
+//   bool  contains(const void* address) const
+//
+// Signal-safety: the last-hit index is a plain static volatile int.
+// DTLS initialisation for shared libraries calls calloc internally; if a
+// profiler signal fires on a thread whose TLS block has not been set up yet
+// while that thread is inside malloc, any thread_local access deadlocks on
+// the allocator lock.  A plain static volatile int avoids TLS entirely.
+// A benign race on the index is acceptable — the worst case is a cache miss
+// that falls through to the O(n) linear scan.
+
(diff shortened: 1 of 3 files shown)


ddprof-lib/src/main/cpp/findLibraryImpl.h
```diff
@@ -0,0 +1,51 @@
...
+// Signal-safety: the last-hit index is a plain static volatile int.
+// DTLS initialisation for shared libraries calls calloc internally; if a
+// profiler signal fires on a thread whose TLS block has not been set up yet
+// while that thread is inside malloc, any thread_local access deadlocks on
+// the allocator lock.  A plain static volatile int avoids TLS entirely.
+// A benign race on the index is acceptable — the worst case is a cache miss
+// that falls through to the O(n) linear scan.
+
+template<typename CacheEntry, typename CacheArray>
+static inline CacheEntry* findLibraryByAddressImpl(const CacheArray& libs, const void* address) {
```
kaahos:
```suggestion
inline CacheEntry* findLibraryByAddressImpl(const CacheArray& libs, const void* address) {
```
I'd suggest removing `static` from the function prototype as it would otherwise mean that the function is private to the translation unit: several header inclusions would imply several independent copies of the function (so several `last_hit`).

It is not a "real" issue though as several independent `last_hit` variables just means falling through to the linear scan which, in the end, is correct.

kaahos (approved):
Minor improvement suggestion which does not modify the final output.

ddprof-lib/src/main/cpp/findLibraryImpl.h
jbachorik: Addressed — removed `static` from the template function. `inline` alone is sufficient; the linker will merge instantiations across TUs, giving a single shared `last_hit` cache.

jbachorik: Fixed

PR #460, opened 2026-04-10
Fix snapshot dump crash on stale jmethodID (PROF-14003)

jbachorik:
**What does this PR do?**:

Fixes a crash in `Lookup::fillJavaMethodInfo` during snapshot dump when profiled jmethodIDs refer to classes that were subsequently unloaded. Two mitigations close the TOCTOU window:

1. **`check_jmethodID_hotspot` (vmStructs.cpp)**: Before reading `class_loader_data` from the `cpool_holder` Klass pointer, verify the full memory range is still readable via `SafeAccess::isReadableRange`. This catches freed/reclaimed Klass pages before JVMTI is called.

2. **`fillJavaMethodInfo` (flightRecorder.cpp)**: Add `method_class != NULL` guard after `GetMethodDeclaringClass` succeeds, preventing `GetClassSignature` from being called with a null JNI handle.

**Motivation**:

On JDK 1.8.0_472, profiling captures jmethodIDs in async signal handlers. By dump time, associated classes may have been unloaded. `GetMethodDeclaringClass` can return a `jclass` wrapping a stale/garbage oop, and the existing J9 sentinel guard (`method_class != (jclass)-1`) is bypassed on HotSpot (`!VM::isOpenJ9()` is always true). This causes `GetClassSignature` to crash inside `oopDesc::is_a()`.

Crash trace: `#0 oopDesc::is_a() → #1 jvmti_GetClassSignature → #2 Lookup::fillJavaMethodInfo`

**Additional Notes**:

This is a mitigation, not a complete fix. The TOCTOU window between `check_jmethodID` and the JVMTI calls is narrowed but not eliminated. A complete fix would require a `ClassUnload` listener to eagerly invalidate cached jmethodIDs, which is a larger change. The `isReadableRange` guard covers the exact memory range accessed at `cpool_holder + class_loader_data_offset`.

**How to test the change?**:

Reproducing the race reliably requires running under a class-loading-heavy workload on JDK 8 with frequent GC cycles during dump. The fix is a defensive guard; unit-level testing is not feasible without a test-only API that forces jmethodID staleness. CI should verify no regression in existing profiling tests.

**For Datadog employees**:

- [x] JIRA: [PROF-14003](https://datadoghq.atlassian.net/browse/PROF-14003)

[PROF-14003]: https://datadoghq.atlassian.net/browse/PROF-14003?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbachorik:

jbachorik:
@codex review

rkennke (approved):

PR #461, opened 2026-04-10
Fix JVM crash when Thread::current() returns nullptr (PROF-13072)

jbachorik:
**What does this PR do?**:
Fixes a JVM crash where `Thread::current()` returns `nullptr` inside ASGCT/JFR allocation paths when a profiling signal fires during thread initialization.

**Motivation**:
There is a race window in `start_routine_wrapper` between `Profiler::registerThread()` (which arms the per-thread CPU timer, enabling SIGPROF delivery) and `routine(params)` (which calls `thread_native_entry` → `pd_set_thread()`, setting `Thread::current()` in JVM TLS). If SIGPROF or SIGVTALRM fires in this window, ASGCT can be invoked on a thread where `Thread::current()` (ELF TLS) is still `nullptr`, crashing in JFR allocation paths (`resource_allocate_bytes` called from JfrStackTrace, JfrArtifactSet, etc.). This only manifests in virtualized environments where OS scheduling makes the race window much more likely to be hit.

**Additional Notes**:
Two complementary changes:

1. **Narrow the race window** (`start_routine_wrapper`, `start_routine_wrapper_spec`): move `Profiler::registerThread()` inside the existing `SignalBlocker` scope. The timer is then armed while SIGPROF/SIGVTALRM are masked; any pending signal fires only after signals are re-enabled (but before `routine(params)`) and is discarded by the guard below.

2. **Signal handler guard with one-shot init window** (`CTimer::signalHandler`, `WallClockASGCT::signalHandler`): when `JVMThread::isInitialized() && JVMThread::current() == nullptr`, skip the sample — but only if the thread's `_init_window` countdown is still active (starts at 1, decremented on first skip). `JVMThread::current()` reads the JVM's own pthread TLS key (found during startup) — the same value `pd_set_thread()` writes.

   The one-shot countdown distinguishes the two cases where `JVMThread::current()` is null:
   - **JVM threads in the race window**: the first signal after the `SignalBlocker` exits is skipped. POSIX guarantees at most one pending signal of each type, so by the time the window expires, `pd_set_thread()` has been called.
   - **Pure native threads** (e.g. `NativeThreadCreator`): `JVMThread::current()` is always null. The countdown expires after one skip, and all subsequent signals are sampled normally — no permanent loss of native thread samples.

**How to test the change?**:
The crash reproduces only in virtualized environments. The benchmark at https://github.com/DataDog/java-profiler/tree/zgu/ctx_benchmark was used to trigger it originally. Unit-level reproduction is not feasible (race window requires specific scheduling that doesn't occur on bare metal reliably).

`DynamicNativeThread` and `NativeThreadTest` on J9/glibc/aarch64 verify that the guard does not permanently suppress native (non-JVM) thread samples.

**For Datadog employees**:
- [x] JIRA: [PROF-13072](https://datadoghq.atlassian.net/browse/PROF-13072)

[PROF-13072]: https://datadoghq.atlassian.net/browse/PROF-13072?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
+++ b/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
@@ -82,12 +82,21 @@ static void init_thread_tls() {
 __attribute__((visibility("hidden")))
 static void* start_routine_wrapper_spec(void* args) {
     RoutineInfo* thr = (RoutineInfo*)args;
-    func_start_routine routine = thr->routine();
-    void* params = thr->args();
-    delete_routine_info(thr);
-    init_thread_tls();
-    int tid = ProfiledThread::currentTid();
-    Profiler::registerThread(tid);
+    func_start_routine routine;
+    void* params;
+    int tid;
+    {
+        // Keep signals blocked across delete_routine_info, init_thread_tls, and
+        // registerThread for the same reasons as start_routine_wrapper: ASAN
+        // lock-ordering and the JVM TLS race window (PROF-13072).
+        SignalBlocker blocker;
+        routine = thr->routine();
+        params = thr->args();
+        delete_routine_info(thr);
+        init_thread_tls();
+        tid = ProfiledThread::currentTid();
+        Profiler::registerThread(tid);
+    }
     void* result = routine(params);
     Profiler::unregisterThread(tid);
     ProfiledThread::release();
@@ -126,6 +135,7 @@ static void* start_routine_wrapper(void* args) {
     RoutineInfo* thr = (RoutineInfo*)args;
     func_start_routine routine;
     void* params;
+    int tid;
     {
         // Block profiling signals while accessing and freeing RoutineInfo
         // and during TLS initialization. Under ASAN, new/delete/
@@ -133,14 +143,19 @@ static void* start_routine_wrapper(void* args) {
         // allocator lock. A profiling signal during any of these calls
         // runs ASAN-instrumented code that tries to acquire the same
         // lock, causing deadlock.
+        // registerThread is also kept inside the blocker so that the CPU
+        // timer is armed while SIGPROF/SIGVTALRM are masked.  Any pending
+        // signal fires only after signals are re-enabled (when the blocker
+        // scope exits), at which point JVMThread::current() is still null
+        // and the guard in CTimer::signalHandler discards the sample safely.
         SignalBlocker blocker;
         routine = thr->routine();
         params = thr->args();
         delete thr;
         ProfiledThread::initCurrentThread();
+        tid = ProfiledThread::currentTid();
+        Profiler::registerThread(tid);
     }
-    int tid = ProfiledThread::currentTid();
-    Profiler::registerThread(tid);
     void* result = nullptr;
     // Handle pthread_exit() bypass - the thread calls pthread_exit()
     // instead of normal termination
--- a/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
+++ b/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
@@ -72,6 +72,17 @@ static void init_thread_tls() {
     ProfiledThread::initCurrentThread();
 }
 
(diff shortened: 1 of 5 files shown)


ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
```diff
@@ -126,21 +137,28 @@ static void* start_routine_wrapper(void* args) {
...
+        // signal fires only after signals are re-enabled (when the blocker
+        // scope exits), at which point JVMThread::current() is still null
+        // and the guard in CTimer::signalHandler discards the sample safely.
         SignalBlocker blocker;
         routine = thr->routine();
         params = thr->args();
         delete thr;
         ProfiledThread::initCurrentThread();
+        tid = ProfiledThread::currentTid();
+        ProfiledThread::currentSignalSafe()->startInitWindow();
```
kaahos, line 160:
```suggestion
if (ProfiledThread *pt = ProfiledThread::currentSignalSafe()) {
    pt->startInitWindow();
}
```

`currentSignalSafe()` can return `null`, so I'd recommend to add a guard.

ddprof-lib/src/main/cpp/thread.h
```diff
@@ -65,6 +65,7 @@ class ProfiledThread : public ThreadLocalData {
   u32 _recording_epoch;
   u32 _misc_flags;
   int _filter_slot_id; // Slot ID for thread filtering
+  uint8_t _init_window; // Countdown for JVM thread init race window (PROF-13072)
```
kaahos, line 68: Related to this modification, I'd suggest adding `_init_window = 0` in [void ProfiledThread::releaseFromBuffer() {](https://github.com/DataDog/java-profiler/blob/f9ea8f2dab7fd5628d69d841f6942464404d643a/ddprof-lib/src/main/cpp/thread.cpp#L89), the window is correctly reset and to avoid any silent sample drop.

kaahos (changes requested):
Claude pointed out something interesting about some code not directly modified in this PR:

> File: `libraryPatcher_linux.cpp`:98-101
> 
> Between `init_thread_tls()` and `start_window_and_register()`, signals are unblocked for one instruction (`ProfiledThread::currentTid()`). At that moment `ProfiledThread` is initialized but `_init_window == 0`, so the guard in `WallClockASGCT::signalHandler` doesn't fire. If `SIGVTALRM` arrives here (possible in no-filter mode, since the TID is in `/proc/self/task/`), ASGCT is called with `JVMThread::current() == nullptr` → crash. This is the exact bug the PR claims to fix, re-introduced on the aarch64/musl path.
> 
> Fix: fold `currentTid()` inside `start_window_and_register` so no signal can arrive with `_init_window == 0` and an initialized `ProfiledThread`:
> 
> ```
> __attribute__((noinline))
> static void start_window_and_register() {
>     SignalBlocker blocker;
>     ProfiledThread::currentSignalSafe()->startInitWindow();
>     Profiler::registerThread(ProfiledThread::currentTid());
> }
> ```
> 
> Caller becomes:
>```
> init_thread_tls();
> start_window_and_register();  // no local tid needed
> ```

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
jbachorik: Addressed — added null guard as suggested.

ddprof-lib/src/main/cpp/thread.h
jbachorik: Addressed — also folded `currentTid()` inside the `SignalBlocker` so no signal can arrive with `_init_window == 0` and an initialised `ProfiledThread`; and added `_init_window = 0` in `releaseFromBuffer()` to reset the guard on thread reuse.

kaahos (approved):
Looks good to me!

PR #462, opened 2026-04-10
Centralize Java thread detection and type-safe thread state handling

zhengyu123:
**What does this PR do?**:
Centralize Java thread detection and improve thread state handling

**Motivation**:
Consolidate and cleanup code for consistence, maintainability and readability.

**Additional Notes**:

*Consolidate Java thread detection*
 Java thread detection was scattered across VMThread instance methods, each calling cachedIsJavaThread() independently. Thread state comparisons used magic integers throughout (>= 4, < 12, == 
  8, == 10, == 11), making the code fragile and hard to audit.

*Centralized VMThread::isJavaThread(VMThread*)*                                                                                                                                               
  Replaces the old cachedIsJavaThread() pattern with a single static method that:                                                                                                                
  - Returns false immediately for nullptr (native/JNI threads outside the JVM)
  - Consults the ProfiledThread cache for previously determined results (TYPE_JAVA_THREAD / TYPE_NOT_JAVA_THREAD)                                                                                
  - Falls back to vtable comparison (hasJavaThreadVtable()) for uncached threads, then stores the result         
                                                                                                                                                                                                 
*JVMJavaThreadState enum*                                                                                                                                                                                       
  Names the previously magic thread state integers with HotSpot-matching constants (_thread_in_Java, _thread_blocked, _thread_blocked_trans, etc.). A new convertJvmExecutionState() switch maps 
  them to ExecutionMode in a type-safe, exhaustive way.                                                                                                                                          
     
*Safer vtable reads*
  hasJavaThreadVtable() now uses SafeAccess::load() for the vtable slot comparison, guarding against unmapped memory access.                                                                     
   
*Cleaner ProfiledThread flag API*                                                                                                                                                                    
  Thread type is stored as a 2-bit field in _misc_flags via setJavaThread(bool) / threadType(), replacing the ad-hoc flag manipulation in the old code. 

**How to test the change?**:
- Passed CI tests
- Passed stress tests

**For Datadog employees**:
- [X] JIRA: [PROF-14160](https://datadoghq.atlassian.net/browse/PROF-14160)

[PROF-14160]: https://datadoghq.atlassian.net/browse/PROF-14160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -777,30 +777,12 @@ int HotspotSupport::getJavaTraceAsync(void *ucontext, ASGCT_CallFrame *frames,
...
-      _thread_in_Java           =  8, // running in Java or in stub code
-      _thread_in_Java_trans     =  9, // corresponding transition state (not used, included for completeness)
-      _thread_blocked           = 10, // blocked in vm 
-      _thread_blocked_trans     = 11, // corresponding transition state
-      _thread_max_state         = 12  // maximum thread state+1 - used forstatistics allocation
-  };
-  **/
-
-  bool in_java = (state == 8 || state == 9);
+  JVMJavaThreadState state = vm_thread->state();
```
jbachorik, line 780: Is this code guaranteed never enter for non-java threads?
If not, consider adding an isJavaThread() guard before the state() call, or remove the assert from state() and restore the defensive return-0 behavior for non-Java threads.

ddprof-lib/src/main/cpp/hotspot/vmStructs.cpp
```diff
@@ -629,19 +629,55 @@ void* VMThread::initialize(jthread thread) {
...
 
-static ExecutionMode convertJvmExecutionState(int state) {
+bool VMThread::isJavaThread(VMThread* vm_thread) {
+    // Not a JVM thread - native thread, e.g. thread launched by JNI code
+    if (vm_thread == nullptr) {
+        return false;
+    }
+
+    // Must be called from current thread
+    assert(vm_thread == VMThread::current());
```
jbachorik, line 639: **Undocumented self-only contract**

`isJavaThread()` has `assert(vm_thread == VMThread::current())` restricting it to self-only use, but the header declares it as `static bool isJavaThread(VMThread* thread)` — the parameter name `thread` and the static signature suggest it could be called for any thread. Future callers iterating ProfiledThread entries for housekeeping would trip this assert in debug.

Consider renaming the parameter to `current_thread` in the header declaration and adding a brief comment: `// Must be called for the current thread only`.

zhengyu123: I believe the assert severs as self documentation. 
- Calling from other threads is always racy
- There are *no* callsites from other threads at this point, so we better enforce that.
- It was implemented without passing in `thread`, but calling `VMThread::current()` inside the method, but Claude code complains can have performance impact.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
zhengyu123: It should be guaranteed - it asserts inside `VMThread::state()`
Calling from non-java thread is a bug.

jbachorik (approved):

PR #470, opened 2026-04-14
Add deadlock event with JVMTI-captured stacktraces

jbachorik:
**What does this PR do?**:

Adds a native JFR deadlock event (`datadog.DeadlockedThread`) that captures real
JVMTI stacktraces for both the deadlocked thread and the lock owner. This provides
proper navigable stacktraces in JFR tools — a significant improvement over the
string-based stacks in the dd-trace-java fallback implementation.

Key components:
- New `T_DEADLOCK` JFR event type with fields: deadlockId, eventThread, stackTrace,
  lockName, lockOwnerThread, lockOwnerStackTrace
- JVMTI `GetStackTrace()` capture for both threads in the deadlock pair, stored in
  the profiler's `CallTraceStorage` as real stacktrace constant pool references
- `DeadlockDetector` Java class using `ThreadMXBean.findDeadlockedThreads()` with
  periodic polling on a daemon thread
- `JavaProfiler.startDeadlockDetector()` / `stopDeadlockDetector()` lifecycle API

**Motivation**:

PROF-11534 — The existing deadlock event in dd-trace-java uses JDK JFR with string
stacktraces. This PR adds the improved native implementation in java-profiler that
captures real stacktraces via JVMTI, which dd-trace-java can use when the profiler
is loaded.

**Additional Notes**:

- dd-trace-java keeps its existing `DeadlockEventFactory` as a fallback for
  environments where the profiler isn't loaded
- Works on JDK 8+ (all APIs are JDK 1.5+; JVMTI is always available)
- One event per deadlocked thread; events in the same cycle share a `deadlockId`
- Lock name truncated to 160 chars to fit within single-byte JFR event size prefix
- Supports both monitor and `java.util.concurrent` lock deadlocks (falls back to
  `findMonitorDeadlockedThreads()` when synchronizer usage is not supported)

**How to test the change?**:

```bash
./gradlew ddprof-test:testRelease --tests '*DeadlockDetectionTest*'
```

The test creates a real deadlock between two threads, starts the detector with a
100ms interval, and verifies that `datadog.DeadlockedThread` events appear in the
profiler's JFR output with non-empty stacktraces and matching deadlock IDs.

**For Datadog employees**:

- [x] JIRA: [PROF-11534](https://datadoghq.atlassian.net/browse/PROF-11534)

🤖 Generated with [Claude Code](https://claude.com/claude-code) via muse implement

[PROF-11534]: https://datadoghq.atlassian.net/browse/PROF-11534?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -1641,6 +1641,30 @@ void Recording::recordCpuLoad(Buffer *buf, float proc_user, float proc_system,
...
+void Recording::recordDeadlock(Buffer *buf, int tid, u64 deadlock_id,
+                               u64 call_trace_id, const char *lock_name,
+                               int lock_owner_tid,
+                               u64 lock_owner_call_trace_id) {
+  // Truncate lock_name to fit within a single-byte event size prefix (max 255).
+  // Fixed overhead: ~75 bytes for 7 varints + event type + size prefix.
+  size_t lock_name_len = strlen(lock_name);
+  if (lock_name_len > 160) {
+    lock_name_len = 160;
+  }
```
jbachorik: Can this be done as 'min' and directly at line 1661?

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -601,6 +601,65 @@ void Profiler::recordQueueTime(int tid, QueueTimeEvent *event) {
...
+      !_locks[lock_index = (lock_index + 1) % CONCURRENCY_LEVEL].tryLock() &&
+      !_locks[lock_index = (lock_index + 2) % CONCURRENCY_LEVEL].tryLock()) {
+    return;
+  }
+  _jfr.recordDeadlock(lock_index, tid, deadlock_id, call_trace_id, lock_name,
+                      lock_owner_tid, lock_owner_call_trace_id);
+  _locks[lock_index].unlock();
+}
+
+static u64 captureStackTrace(jthread thread, CallTraceStorage &storage,
```
jbachorik, line 621: Add a warning comment that this is calling JVMTI::GetStackTrace and depending on the Java version this might cause safepoint so this should never be called from an async signal handler

ddprof-test/src/test/java/com/datadoghq/profiler/deadlock/DeadlockDetector.java
```diff
@@ -0,0 +1,162 @@
...
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Periodic deadlock detector using JMX ThreadMXBean.
+ * When a deadlock is detected, emits native JFR events with real JVMTI-captured
+ * stacktraces for both the deadlocked thread and the lock owner.
+ */
+final class DeadlockDetector {
```
jbachorik, line 43: This is a duplicate of what we have in dd-trace-java, right?
Can this be made a test only and have dd-trace-java drive the detection, with java-profiler only dealing with recording the enhanced event?

jbachorik:
@codex review

ddprof-lib/src/main/cpp/flightRecorder.cpp
jbachorik: Good call — the two-liner can collapse into `size_t lock_name_len = std::min(strlen(lock_name), static_cast<size_t>(160));` right at the declaration, which is cleaner and removes the mutable branch entirely. Will update.

jbachorik: Yes, that's cleaner — `size_t lock_name_len = std::min(strlen(lock_name), (size_t)160);` collapses it to a single line. Will fix.

ddprof-test/src/test/java/com/datadoghq/profiler/deadlock/DeadlockDetector.java
jbachorik: Yes, this is intentionally test-only — it lives in `ddprof-test/src/test/java` and is package-private, so it never ships as production code. The design intent is exactly what you describe: in production, dd-trace-java drives detection and calls into java-profiler only for recording the enhanced JFR event; this helper exists purely to exercise the recording path without a full dd-trace-java dependency in tests.

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -601,6 +601,68 @@ void Profiler::recordQueueTime(int tid, QueueTimeEvent *event) {
...
+  int lock_owner_tid = JVMThread::nativeThreadId(env, lock_owner_thread);
+  if (tid < 0 || lock_owner_tid < 0) {
+    return;
+  }
+
+  u64 call_trace_id = captureStackTrace(thread, _call_trace_storage,
+                                        _max_stack_depth);
+  u64 lock_owner_trace_id = captureStackTrace(lock_owner_thread,
+                                              _call_trace_storage,
+                                              _max_stack_depth);
```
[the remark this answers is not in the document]
jbachorik: Valid concern — `_call_trace_storage.put()` is not reentrant from the same thread, so a profiling signal interrupting `captureStackTrace` on the deadlock-detector thread could corrupt the storage or produce a garbled trace. Wrapping both `captureStackTrace` calls (and ideally the `recordDeadlock` store that follows) in a `CriticalSection` guard — mirroring what `recordEvent` does — is the right fix. I'll add that guard before the first `captureStackTrace` call in `recordDeadlockWithCapture`.

Issue #487, opened 2026-04-16
[Libretto] Implement tracking native socket I/O events


jbachorik:
Track blocking I/O operations at the libc function level.

## Scope
- Track: socket I/O via libc functions `send`, `recv`, `write`, `read` (TCP/SOCK_STREAM only)
- Instrumentation approach: PLT hooking (GOT patching), NOT kernel syscall level
- `write`/`read` hooks filter non-socket fds via a lock-free fd-type cache; non-socket fds pass through with one atomic load overhead
- NOT in scope: `sendto`/`recvfrom` (UDP), `connect`, `close`, `shutdown`
- Feature activation: explicit opt-in via profiler argument `nativesocket`
- Rate limiting: ~83 events/s (~5000/min) via PID-based subsampling using a standalone `RateLimiter` class
- Separate JFR event type `datadog.NativeSocketEvent`
- Linux required; macOS compiles as no-op stub

## Sampling Model
- **Time-weighted Poisson-process sampling**: the sampling unit is call duration in TSC ticks, not bytes
- Probability of sampling a call of duration `d` with mean interval `T`: `P = 1 - exp(-d/T)`
- Inverse-transform weight: `weight = 1/P`; invariant: `E[weight × duration] = duration` — `sum(weight_i × duration_i)` is an unbiased estimator of total I/O time
- Long blocking calls (d >> T) have P ≈ 1 and are virtually always sampled regardless of interval magnitude; short fast calls are down-sampled
- Per-thread `PoissonSampler` instances for outbound (send+write) and inbound (recv+read) directions; both read from a single shared `RateLimiter`
- One shared `RateLimiter` covers all four hooks combined: ~83 events/s (~5000/min) total budget across all directions. A write-heavy workload raises the interval but does not suppress long blocking reads because P → 1 as d >> T
- PID controller adjusts `interval_ticks` once per second based on aggregate fire count across all threads
- Epoch-based lazy TLS reset: `_epoch` is bumped on `start()`; per-thread samplers reinitialise lazily on next hook invocation

## Acceptance Criteria

- [ ] Socket I/O tracking can be explicitly enabled via profiler argument `nativesocket`
- [ ] Instrumentation hooks libc `send`, `recv`, `write`, `read` via PLT hooking; `write`/`read` skip non-socket fds via a lock-free fd-type cache
[...]

PR #488, opened 2026-04-16
feat: native socket I/O tracking via PLT hooks (PROF-10637)

jbachorik:
**What does this PR do?**

Intercepts libc `send`, `recv`, `write`, `read` via PLT hooking and records
`datadog.NativeSocketEvent` JFR events with **time-weighted** inverse-transform
sampling (Poisson process, PID rate control, target ~83 events/s / ~5000/min).

**Motivation**

Track blocking TCP socket I/O at the libc function level to surface socket
latency and throughput in the Datadog profiler. Netty with Java NIO transport
(the primary use case) goes through libc `send`/`recv`/`write`/`read`, making
PLT patching the right interception point. Feature is explicitly opt-in via
the `natsock` profiler argument. Implements PROF-10637.

**Configuration**

| Argument        | Effect                                                                |
|-----------------|-----------------------------------------------------------------------|
| `natsock`       | Enable native socket tracking with the default 1 ms initial period    |
| `natsock=100us` | Enable with an explicit initial sampling period (any `parseUnits` unit: `ns`, `us`, `ms`, `s`) |

The interval is the initial parameter of the time-weighted Poisson process;
the PID controller adapts it at runtime to hold the total event rate near
`TARGET_EVENTS_PER_SECOND` (= 83 events/s) across all four hooks combined.

**Architecture**

```mermaid
flowchart LR
    A["Java code<br/>OutputStream.write()"] --> B["JDK native<br/>libnio/libnet"]
    B --> C{"PLT entry<br/>for write()"}
    C -- patched --> D["NativeSocketSampler::write_hook"]
    D --> E{"fd is TCP<br/>socket?"}
    E -- no --> F["orig write()"]
    E -- yes --> G["orig write()"]
    G --> H{"shouldSample<br/>(duration, op)"}
    H -- no --> I["return"]
    H -- yes --> J["recordEvent"]
    J --> K["JFR buffer"]
    F --> I
```

**Sampling pipeline**

```mermaid
flowchart TD
    A["Hook fires with<br/>duration_ticks"] --> B["PoissonSampler<br/>(thread_local)"]
    B --> C{"interval budget<br/>consumed?"}
    C -- no --> D["reject"]
    C -- yes --> E["compute weight<br/>1 / (1 - exp(-d/interval))"]
    E --> F["RateLimiter::accept()"]
    F --> G{"PID budget<br/>for this epoch?"}
    G -- no --> D
    G -- yes --> H["record event<br/>operation / fd-addr / bytes / weight"]
    H --> I["RateLimiter<br/>maybeUpdateInterval<br/>(PID: P=31 I=511 D=3)"]
    I --> J["next epoch interval"]
```

Two `thread_local PoissonSampler` instances per thread: outbound (`send`+`write`)
and inbound (`recv`+`read`). Both share one `RateLimiter` so the ~83 events/s
target is a single budget across all directions.

**Hook installation**

```mermaid
flowchart TD
    A["NativeSocketSampler::start"] --> B["dlsym RTLD_NEXT<br/>(RTLD_DEFAULT fallback on musl)<br/>resolve _orig_send/recv/write/read"]
    B --> C["patch_socket_functions"]
    C --> D["walk every loaded CodeCache"]
    D --> E{"lib imports<br/>send/recv/write/read?"}
    E -- yes --> F["rewrite PLT entry<br/>to *_hook"]
    E -- no --> G["skip"]
    C --> H["also patch dlopen PLT"]
    H --> I["dlopen_hook wraps<br/>subsequent loads"]
    I --> J["updateSymbols +<br/>install_socket_hooks"]
    J --> E
```

Late-loaded libraries (e.g. HotSpot lazily loads `libnet.so` on the first
`java.net.Socket` use) are caught by the `dlopen` wrapper which re-runs
`patch_socket_functions` after each new library is registered.

**JFR event fields** (`datadog.NativeSocketEvent`)

| Field              | Type      | Description                                          |
|--------------------|-----------|------------------------------------------------------|
| `startTime`        | long      | Event start (ticks)                                  |
| `eventThread`      | Thread    | Calling Java thread (constant pool)                  |
| `stackTrace`       | StackTrace| Captured stack (constant pool)                       |
| `duration`         | long      | Syscall duration (ticks)                             |
| `operation`        | string    | `SEND` or `RECV`                                     |
| `remoteAddress`    | string    | `ip:port` (IPv4 or `[ipv6]:port`); empty if unknown  |
| `bytesTransferred` | long      | Bytes actually transferred                           |
| `weight`           | float     | Inverse sampling probability (`1 / P`)               |
| `spanId`           | long      | Datadog span id from active context                  |
| `localRootSpanId`  | long      | Datadog local root span id                           |
| _context attrs_    | strings   | User context attributes                              |

`sum(weight × duration)` is an unbiased estimator of total socket I/O time;
`sum(weight × bytesTransferred)` estimates total bytes.

**Key design notes**

- **Scope**: TCP blocking `send`/`recv`/`write`/`read` only. UDP (`sendto`/`recvfrom`)
  and Netty's native epoll / io_uring transports are explicitly out of scope.
- **Fast path**: `write`/`read` hooks short-circuit on non-socket fds via a
  lock-free `std::atomic<uint8_t> _fd_type_cache[65536]` — one relaxed atomic
  load per call on the non-socket path.
- **Thread context**: hook bodies run on the calling Java thread (not a signal
  handler), so `malloc` and `std::mutex` are safe.
- **PLT re-patching on dlopen**: `install_socket_hooks()` (called from
  `dlopen_hook` after `updateSymbols`) re-runs `patch_socket_functions` so
  lazily-loaded libraries (`libnet.so`, `libnio.so`, native Netty transports)
  are patched the moment they appear.
- **fd-to-address cache**: intentionally **not** cleared on JFR chunk boundaries;
  stale addresses for reused fds are a known, accepted trade-off to avoid
  racing with in-flight recordings.
- **Original function pointer lifecycle**: `_orig_send/recv/write/read` are
  assigned once (when `_socket_size == 0`) and are intentionally not nulled in
  `unpatch_socket_functions` to avoid a memory-ordering race with in-flight
  hook invocations that may still be executing.
- **Platform scope**: `#if defined(__linux__)` — compiled as no-op stubs on
  macOS. Runs on both glibc and musl Linux; on musl `RTLD_NEXT` returns NULL
  (libc precedes the profiler DSO in the link map), so `RTLD_DEFAULT` is used
  as fallback to locate the real libc symbols. Feature is additionally disabled
  in Java tests on J9/Zing via `Platform.isJ9()` / `Platform.isZing()` guards.
- **`NativeSocketEvent`** lives in `nativeSocketSampler.h` (not the shared
  `event.h`) since it is only used by `NativeSocketSampler`.

**How to test the change?**

- **C++ unit tests** — `NativeSocketSamplerHookTest` in
  `ddprof-lib/src/test/cpp/nativeSocketSampler_ut.cpp` — verifies that
  `send_hook`/`recv_hook`/`write_hook`/`read_hook` each delegate to the
  installed `_orig_*` pointer and return its value.

- **JUnit integration tests** in
  `ddprof-test/src/test/java/com/datadoghq/profiler/nativesocket/`:

  | Test                                 | What it verifies                                                  |
  |--------------------------------------|-------------------------------------------------------------------|
  | `NativeSocketEnabledTest`            | Events produced when feature is enabled                           |
  | `NativeSocketDisabledTest`           | No events when feature is not enabled                             |
  | `NativeSocketEventFieldsTest`        | All required JFR fields present and valid                         |
  | `NativeSocketEventThreadTest`        | `eventThread` populated with the calling thread                   |
  | `NativeSocketStackTraceTest`         | Stack trace captured on events                                    |
  | `NativeSocketRemoteAddressTest`      | `remoteAddress` in `ip:port` format                               |
  | `NativeSocketSendRecvSeparateTest`   | SEND and RECV tracked independently                               |
  | `NativeSocketBytesAccuracyTest`      | `sum(weight × duration)` unbiased estimator within tolerance      |
  | `NativeSocketRateLimitTest`          | Event count ≪ operation count (PID rate control active); `weight > 1` on sampled events |
  | `NativeSocketRestartTest`            | Events produced after profiler stop/restart cycle                 |
  | `NativeSocketUdpExcludedTest`        | UDP (`sendto`/`recvfrom`) produces zero events                    |
  | `NativeSocketMacOsNoOpTest`          | No events on macOS (no-op stub)                                   |

Spec: https://github.com/DataDog/java-profiler/issues/487

**For Datadog employees**

- [x] JIRA: [PROF-10637](https://datadoghq.atlassian.net/browse/PROF-10637)

[PROF-10637]: https://datadoghq.atlassian.net/browse/PROF-10637?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbachorik:
Could not apply automatically — The reviewer comment (ID 4261434558) is an automated CI test-results report showing all 32 jobs passed with no failures. There is no code concern, requested change, or actionable finding to address with a patch.. Manual attention needed.

jbachorik:
@codex review

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -1827,6 +1827,24 @@ void Recording::recordMallocSample(Buffer *buf, int tid, u64 call_trace_id,
...
 }
 
+void Recording::recordNativeSocketSample(Buffer *buf, int tid, u64 call_trace_id,
+                                         NativeSocketEvent *event) {
+  int start = buf->skip(1);
+  buf->putVar64(T_NATIVE_SOCKET);
+  buf->putVar64(event->_start_time);
+  buf->putVar64(tid);
+  buf->putVar64(call_trace_id);
+  buf->putVar64(event->_end_time - event->_start_time);
```
rkennke: TSC::ticks() is not guaranteed monotonic across core migration; a thread that migrates between t0 and t1 can observe t1 < t0. With unsigned subtraction the JFR duration becomes ~`2^64, and since sum(weight × duration)is the headline estimator, one such event can swamp the aggregate. Recommend clamping:event->_end_time >= event->_start_time ? end-start : 0`. Cheap insurance.

ddprof-lib/src/main/cpp/livenessTracker.cpp
```diff
@@ -211,6 +211,10 @@ void LivenessTracker::stop() {
 
 Error LivenessTracker::initialize(Arguments &args) {
   _enabled = args._gc_generations || args._record_liveness;
+  // Per-call toggles that must reflect the current start's arguments even when
+  // the tracker's table is reused from a prior initialize (_initialized=true).
+  // On musl the test JVM is not forked per test, so the singleton survives.
+  _record_heap_usage = args._record_heap_usage;
```
rkennke: Is this related to this PR?

ddprof-lib/src/main/cpp/nativeSocketSampler.h
```diff
@@ -0,0 +1,218 @@
...
+private:
+    static NativeSocketSampler* const _instance;
+
+    // Set once by setOriginalFunctions() (called under _lock, before PLT patching) and
+    // never reset to null while hooks are active.  No atomic needed: the __ATOMIC_RELEASE
+    // on each PLT patch provides a store-store barrier that keeps these assignments
+    // visible before the PLT entry becomes observable; the _socket_active release/acquire
+    // pair establishes happens-before for any hook that sees _socket_active=true.
+    //
+    // ASSUMPTION (documented for any future port to a weaker memory model): we do not
```
rkennke: Or, do the right thing now (std::atomic fields, etc), and avoid headaches later, if we ever support restart-in-prod, because by then we might have forgotten about this place, etc. It seems a cheap-enough change and avoids a potential future trip-wire.

ddprof-lib/src/main/cpp/nativeSocketSampler.h
```diff
@@ -0,0 +1,218 @@
...
+    // 2=non-TCP).  An entry is valid only when its high nibble equals _fd_cache_gen
+    // mod 16.  Incrementing _fd_cache_gen invalidates all entries in O(1) without
+    // touching the 65536-entry array.
+    //
+    // KNOWN LIMITATION (mod-16 generation wrap): _fd_cache_gen is only consulted via
+    // its low 4 bits.  After 16 start() cycles the generation wraps and stale entries
+    // from a previous incarnation become indistinguishable from current ones until each
+    // fd is naturally re-probed.  Profiler restarts are not exercised in production
+    // (only in tests), so the wrap is benign in practice.  If restart-in-prod ever
+    // becomes a supported mode, widen _fd_cache_gen to uint32_t and store the full
```
rkennke, line 162: Or do it now and avoid the headache later. Same as elsewhere.

ddprof-lib/src/main/cpp/nativeSocketSampler.h
```diff
@@ -0,0 +1,218 @@
...
+
+    // fd -> "ip:port" string cache.  Bounded to MAX_FD_CACHE entries; no
+    // eviction is performed (entries for closed/reused fds are stale until
+    // the next stop(), but stale addresses are a known, accepted limitation).
+    // Once the cache fills, _fd_cache_full latches true to skip further
+    // resolveAddr() calls — without this latch the hot path would degrade to
+    // per-event getpeername()+inet_ntop()+std::string allocation that gets
+    // thrown away (since emplace is gated on size).
+    static const int MAX_FD_CACHE = 65536;
+    std::unordered_map<int, std::string> _fd_cache;
```
rkennke: If I understand correctly, once the cache is full, the profiler stops resolving addresses. 64K entries can be reached quickly on a busy server, and once that happens, it silently just starts remoteAddress going empty. Downstream nobody gets an error — it looks like "the profiler stopped resolving peer addresses," which is hard to distinguish from a bug. The only thing that clears it is clearFdCache() and that's called only on start()/stop(), i.e. effectively never in a long-running prod session.

I suggest to do either of:
  1. Make the limitation observable — bump a counter when the latch trips, so in the field you can see      "address resolution disabled, cache full" instead of guessing why remoteAddress went blank.                                                    
  2. Bound it properly — an LRU, or a periodic clear on JFR chunk boundary. You explicitly say it does not clear on chunk boundaries (to avoid racing in-flight recordings), so an LRU keyed on recency is the cleaner fix — it keeps the cache useful indefinitely without unbounded growth.

rkennke (changes requested):
This is a nice feature! I have a few requests.

ddprof-lib/src/main/cpp/flightRecorder.cpp
jbachorik: I need to check if we don't do this assumption elsewhere. IIRC, JFR is not really bothering with the core migration TSC issues ...

ddprof-lib/src/main/cpp/livenessTracker.cpp
jbachorik: Not really. My several rebases must have pulled it ... I will clean this up before moving from draft

ddprof-lib/src/main/cpp/nativeSocketSampler.h
```diff
@@ -0,0 +1,218 @@
...
+    // Fds outside [0, FD_TYPE_CACHE_SIZE) are probed on every call.
+    static const int     FD_TYPE_CACHE_SIZE  = 65536;
+    // FD_TYPE_UNKNOWN is the implicit value-zero sentinel for never-written entries
+    // and gen-mismatch entries; it is decoded by the (cached >> 4) != gen path in
+    // isSocket(), not by an explicit comparison against this constant.
+    static const uint8_t FD_TYPE_UNKNOWN     = 0;
+    static const uint8_t FD_TYPE_SOCKET      = 1;
+    static const uint8_t FD_TYPE_NON_SOCKET  = 2;
+    std::atomic<uint8_t> _fd_cache_gen{0};   // incremented on each cache reset
+    std::atomic<uint8_t> _fd_type_cache[FD_TYPE_CACHE_SIZE];
```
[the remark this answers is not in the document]
jbachorik: The sequence is: constructor (uninitialized cache, gen=0) → start() → clearFdCache() bumps gen to 1 → _socket_active set true → hooks enabled. By the time any hook reads `_fd_type_cache`, gen is 1 and all initial zero-bytes (high nibble 0 ≠ gen 1) are cache misses. Explicit zero-initialization of the array would be correct defensively but provides no runtime behavior change under this call sequence.

ddprof-lib/src/main/cpp/nativeSocketSampler.h
jbachorik: The field at nativeSocketSampler.h line 161 is `std::atomic<uint8_t> _fd_cache_gen{0}` — it is already declared as std::atomic and zero-initialized with `{0}`.

ddprof-lib/src/main/cpp/nativeSocketSampler.cpp
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    int so_type;
+    socklen_t solen = sizeof(so_type);
+    int rc = getsockopt(fd, SOL_SOCKET, SO_TYPE, &so_type, &solen);
+    if (rc == 0) {
+        bool tcp = (so_type == SOCK_STREAM);
+        uint8_t type = tcp ? FD_TYPE_SOCKET : FD_TYPE_NON_SOCKET;
+        _fd_type_cache[fd].store((uint8_t)(((gen & 0xF) << 4) | type),
+                                 std::memory_order_release);
```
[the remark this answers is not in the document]
jbachorik: Fixed in 4fa1940724a2aad41034b8ca78335eb75224ecfc.

ddprof-lib/src/main/cpp/poissonSampler.h
```diff
@@ -0,0 +1,238 @@
...
+        if (_epoch != epoch_now) {
+            reset(interval, epoch_now);
+        }
+        _used += value;
+        if (_used < _threshold) {
+            return false;
+        }
+        // Threshold crossed: advance by a fresh Exp draw so the next
+        // inter-arrival gap is independent of all previous ones.
+        _threshold += nextExp(interval);
```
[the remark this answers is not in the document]
jbachorik: Fixed in 4fa1940724a2aad41034b8ca78335eb75224ecfc.

ddprof-test/src/test/java/com/datadoghq/profiler/nativesocket/NativeSocketUdpExcludedTest.java
```diff
@@ -0,0 +1,84 @@
...
+import org.openjdk.jmc.common.item.IItemCollection;
+import org.openjdk.jmc.common.item.IItemIterable;
+import org.openjdk.jmc.common.item.IMemberAccessor;
+import org.openjdk.jmc.common.unit.UnitLookup;
+
+import java.net.DatagramPacket;
+import java.net.DatagramSocket;
+import java.net.InetAddress;
+
+import static org.junit.jupiter.api.Assertions.*;
```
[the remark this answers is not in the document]
jbachorik: Fixed in 4fa1940724a2aad41034b8ca78335eb75224ecfc.

ddprof-stresstest/src/jmh/java/com/datadoghq/profiler/stresstest/scenarios/throughput/NativeSocketOverheadBenchmark.java
```diff
@@ -0,0 +1,173 @@
...
+
+    private final byte[] buf = new byte[CHUNK];
+
+    @Setup(Level.Trial)
+    public void setup() throws Exception {
+        if (profilerActive) {
+            JavaProfiler profiler = JavaProfiler.getInstance();
+            Path jfr = Files.createTempFile("nativesocket-bench", ".jfr");
+            profiler.execute("start,nativesocket,jfr,file=" + jfr.toAbsolutePath());
+        }
```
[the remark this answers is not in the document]
jbachorik: Fixed in 4fa1940724a2aad41034b8ca78335eb75224ecfc.

ddprof-lib/src/main/cpp/nativeSocketSampler.cpp
```diff
@@ -0,0 +1,376 @@
...
+        return Error("natsock profiling not requested");
+    }
+    return Error::OK;
+}
+
+Error NativeSocketSampler::start(Arguments &args) {
+    // Clear the fd cache on start so stale entries from a prior session don't
+    // produce misattributed events even if stop() was not called.
+    clearFdCache();
+    // Initial sampling period: args._nativesocket_interval (ns) when > 0,
```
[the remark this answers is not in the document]
jbachorik: Fixed in 4fa1940724a2aad41034b8ca78335eb75224ecfc.

ddprof-test/src/test/java/com/datadoghq/profiler/nativesocket/NativeSocketRateLimitTest.java
```diff
@@ -0,0 +1,168 @@
...
+        // Target rate is ~5000 events/min; for a 10-second window ~833 events is the ceiling.
+        // Using operations/2 (10000) provides a generous upper bound that must still be met.
+        assertTrue(eventCount < operations / 2,
+                "Too many events sampled (rate limiting not working): event count (" + eventCount
+                        + ") should be less than operations/2 (" + operations / 2 + ")");
+
+        // At least some events must have weight > 1, indicating byte-weighted sampling
+        assertTrue(foundWeightAboveOne,
+                "Expected at least one event with weight > 1 (byte-weighted inverse-transform sampling)");
+    }
```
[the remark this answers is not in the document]
jbachorik: Fixed in 4fa1940724a2aad41034b8ca78335eb75224ecfc.

ddprof-lib/src/main/cpp/flightRecorder.cpp
jbachorik: Fixed in 4fa1940724a2aad41034b8ca78335eb75224ecfc.

ddprof-lib/src/main/cpp/livenessTracker.cpp
jbachorik: Fixed in 4fa1940724a2aad41034b8ca78335eb75224ecfc.

ddprof-lib/src/main/cpp/nativeSocketSampler.h
jbachorik: Fixed in 4fa1940724a2aad41034b8ca78335eb75224ecfc.

ddprof-lib/src/main/cpp/nativeSocketSampler.h
jbachorik: Fixed in 4fa1940724a2aad41034b8ca78335eb75224ecfc.

ddprof-lib/src/main/cpp/nativeSocketSampler.cpp
```diff
@@ -0,0 +1,411 @@
...
+    NativeSocketEvent event;
+    event._start_time = t0;
+    event._end_time   = t1;
+    event._operation  = op;
+    event._remote_addr[0] = '\0';
+    {
+        std::lock_guard<std::mutex> lock(_fd_cache_mutex);
+        auto it = _fd_cache.find(fd);
+        if (it != _fd_cache.end()) {
+            strncpy(event._remote_addr, it->second.c_str(), sizeof(event._remote_addr) - 1);
```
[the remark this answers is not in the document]
jbachorik: Fixed in fd78c503c: `recordEvent()` now always calls `resolveAddr()` for sampled events, so a reused fd gets a fresh peer address within one sampling interval.

ddprof-lib/src/main/cpp/nativeSocketSampler.cpp
```diff
@@ -0,0 +1,411 @@
...
+    {
+        uint64_t n = _send_hook_calls.fetch_add(1, std::memory_order_relaxed);
+        if (n < 3 || (n & 0x3FF) == 0) {
+            TEST_LOG("NativeSocketSampler::send_hook #%llu fd=%d len=%zu flags=0x%x",
+                     (unsigned long long)(n + 1), fd, len, flags);
+        }
+    }
+#endif
+    u64 t0 = TSC::ticks();
+    return record_if_positive(fd, fn(fd, buf, len, flags), t0, TSC::ticks(), 0);
```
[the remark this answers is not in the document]
jbachorik: Fixed in fd78c503c: `send_hook` and `recv_hook` now call `isSocket(fd)` before proceeding, so connected UDP sockets are excluded the same way `write`/`read` hooks are.

rkennke (approved):
Thanks for addressing the issues. Looks good now!

jbachorik:
@rkennke Thanks for the patient review :)

PR #492, opened 2026-04-20
Fix IBM J9 crash: replace pthread_cleanup_push with abi::__forced_unwind catch (SCP-1154)

jbachorik:
**What does this PR do?**:
Fixes a JVM crash on IBM J9 / OpenJ9 when profiling is enabled (SCP-1154, regression from ddprof 1.39.0).

Replaces `pthread_cleanup_push/pop` in both `start_routine_wrapper` variants with an explicit `catch(abi::__forced_unwind&)` guard. Also adds the same guard to `J9WallClock::timerLoop` to ensure `VM::detachThread()` is always called when J9 cancels the sampler thread during JVM shutdown.

**Motivation**:
IBM J9's thread teardown raises `_Unwind_ForcedUnwind` via libgcc (sourced from `libj9thr29.so`). In C++ mode, `pthread_cleanup_push` expands to a `__pthread_cleanup_class` object whose destructor is **implicitly `noexcept`** (C++11). When J9's forced-unwind propagates through this noexcept context, `std::terminate()` → `abort()` is called, crashing the JVM.

This was a regression introduced in commit `2063c659` ("Prevent potential race in thread startup and cleanup dead code"), which added `pthread_cleanup_push/pop` to `start_routine_wrapper`. The fix replaces that mechanism with an explicit `catch(abi::__forced_unwind&)` + `throw;` pattern, which is the GCC/glibc-recommended approach for cleanup code that must run during POSIX thread cancellation.

Root cause (from Reshmi Anand's investigation):
- Crash stack: `abort() ← libjavaProfiler.so ← _Unwind_ForcedUnwind ← libj9thr29.so`
- Version bisect: ddprof 1.37.0 (no crash) → 1.39.0 (crash)
- Trigger: `-Ddd.profiling.enabled=true` on IBM WAS 9.0.5.16 with OpenJDK 1.8.0_462

**Additional Notes**:
- `abi::__forced_unwind` is correct: J9 calls `_Unwind_ForcedUnwind` from libgcc (confirmed by crash stack frame `_Unwind_ForcedUnwind ← libj9thr29.so`), which uses the GCC/Itanium ABI `__forced_unwind` type.
- `pthread_exit()` is also covered: on glibc it raises its own `__forced_unwind`.
- The `thread_cleanup` static helper is removed (no longer needed).
- In `start_routine_wrapper_spec` (aarch64), `tid` is now captured before the `try` block to avoid the lazy-allocating `ProfiledThread::current()` path inside the catch.
- `J9WallClock::timerLoop` `PushLocalFrame`/`PopLocalFrame` balance: if cancellation fires between push/pop, `DetachCurrentThread` (via `VM::detachThread()`) releases the outstanding JNI local frame per the JNI spec.

**How to test the change?**:
New regression test `ddprof-lib/src/test/cpp/forced_unwind_ut.cpp` runs on Linux without a JVM and verifies:
1. `catch(abi::__forced_unwind&)` fires on `pthread_cancel` and the cleanup block runs.
2. The `throw;` re-throw lets the thread exit as `PTHREAD_CANCELED` (not via `std::terminate()`).
3. `ProfiledThread::release()` completes safely inside the catch block.
4. `pthread_exit()` (which also uses `__forced_unwind` on glibc) is caught by the same pattern.

Run: `./gradlew :ddprof-lib:gtestRelease --tests "ForcedUnwindTest*"`

End-to-end IBM J9 reproduction requires the Docker-based reproducer from the SCP-1154 investigation artifacts.

**For Datadog employees**:
- [x] JIRA: [SCP-1154](https://datadoghq.atlassian.net/browse/SCP-1154)

🤖 Generated with [Claude Code](https://claude.com/claude-code) via muse implement

[SCP-1154]: https://datadoghq.atlassian.net/browse/SCP-1154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/j9/j9WallClock.cpp
+++ b/ddprof-lib/src/main/cpp/j9/j9WallClock.cpp
@@ -58,16 +58,14 @@ void J9WallClock::stop() {
 }
 
 void J9WallClock::timerLoop() {
-  JNIEnv *jni = VM::attachThread("java-profiler Sampler");
-  jvmtiEnv *jvmti = VM::jvmti();
-
-  int max_frames = _max_stack_depth + MAX_NATIVE_FRAMES + RESERVED_FRAMES;
-  ASGCT_CallFrame *frames =
-      (ASGCT_CallFrame *)malloc(max_frames * sizeof(ASGCT_CallFrame));
-
   // IBM J9 may cancel this thread via abi::__forced_unwind during JVM shutdown.
-  // Catch it to ensure free(frames) and VM::detachThread() always run, then
-  // re-throw so the thread exits properly through the normal unwind path.
+  // Catch it to ensure free(frames) and VM::detachThread() always run — including
+  // if the unwind fires during setup (VM::attachThread or malloc) before the
+  // sampling loop starts — then re-throw so the thread exits properly through
+  // the normal unwind path.  jni and frames are declared before the try and
+  // initialized to null so the catch handler's cleanup is safe in either case:
+  // free(nullptr) is a no-op, and detachThread() is only called when attach
+  // succeeded.
   //
   // PushLocalFrame / PopLocalFrame balance: if the forced unwind fires between
   // PushLocalFrame and PopLocalFrame, the outstanding JNI local frame is released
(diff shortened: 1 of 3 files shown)


zhengyu123:
Claude found two relevant issues:

- j9WallClock.cpp:60   
VM::attachThread and malloc run before the try{} block — if a forced-unwind fires in that narrow window, cleanup is skipped despite the comment claiming otherwise. 

- libraryPatcher_linux.cpp :97   
On aarch64/musl/jdk11 path, init_thread_tls() and start_window_and_register() use separate SignalBlocker scopes — signals are  unblocked between the two calls, creating a gap where the init-window guard isn't set.

jbachorik:
Thanks — both gaps are real. Fixed in 1cb5f6ca:

- **`j9WallClock.cpp`**: `VM::attachThread` and `malloc` are now inside the `try{}` block, with `jni` and `frames` declared as `nullptr` before it so the catch handler's `free(frames)` (a no-op on null) and `VM::detachThread()` are safe if the unwind fires mid-setup. Comment updated accordingly.
- **`libraryPatcher_linux.cpp`**: `init_thread_tls()` and `start_window_and_register()` merged into a single `__attribute__((noinline)) init_tls_and_register()` that holds one `SignalBlocker` spanning `initCurrentThread()` + `startInitWindow()` + `registerThread()`, closing the gap on aarch64/musl/jdk11. `noinline` is preserved so `sigset_t` still stays off `start_routine_wrapper_spec`'s frame.

zhengyu123 (approved):

PR #502, opened 2026-04-24
add early thread-state check and TaskBlock emission for wall-clock

kaahos:
**What does this PR do?**:
Improves wall-clock sampling by adding an early thread-state check so we avoid full sampling work when threads are in cheap-to-detect idle states (e.g. sleeping / condition-variable wait). Adds TLS-based park enter/exit handling so parked threads can be treated efficiently during wall-clock sampling. Introduces TaskBlock recording for blocked intervals. Centralizes TaskBlock eligibility, extends wall-clock counters and `datadog.WallClockSamplingEpoch` JFR fields for observability.

**Motivation**:
Wall-clock sampling should spend less work on threads that are idle or parked while still surfacing blocking time as TaskBlock events for analysis. This aligns native behavior with tracer instrumentation and improves visibility through counters and epoch metrics.

**Additional Notes**:
- JVMTI `MonitorWait` / `MonitorWaited` callbacks are registered only for JDK < 21. For JDK21+, Java-level instrumentation is preferred.
- linked PR: https://github.com/DataDog/dd-trace-java/pull/11247

**How to test the change?**:
```bash
./gradlew :ddprof-lib:gtestDebug_park_state_ut
./gradlew :ddprof-test:testDebug -Ptests=PrecheckTest
./gradlew :ddprof-test:testDebug -Ptests=PrecheckEfficiencyTest
./gradlew :ddprof-test:testDebug -Ptests=ParkTaskBlockTest
./gradlew :ddprof-test:testDebug -Ptests=WallclockMitigationsCombinedTest
```

**For Datadog employees**:
- [x] JIRA: [JIRA-14534]


kaahos:
@copilot review

kaahos:
@copilot+gpt-5.2-codex review

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -1574,6 +1576,23 @@ void Recording::recordQueueTime(Buffer *buf, int tid, QueueTimeEvent *event) {
...
 }
 
+void Recording::recordTaskBlock(Buffer *buf, int tid, TaskBlockEvent *event) {
+  int start = buf->skip(1);
+  buf->putVar64(T_TASK_BLOCK);
+  buf->putVar64(event->_start);
+  buf->putVar64(event->_end - event->_start);
+  buf->putVar64(tid);
+  buf->putVar64(event->_ctx.spanId);
+  buf->putVar64(event->_ctx.rootSpanId);
```
jbachorik: We already have the generic way of capturing context. It would be better to use that instead.
It will also take care of the custom attribute, simplifying the code here.

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -313,6 +320,48 @@ Java_com_datadoghq_profiler_JavaProfiler_recordQueueEnd0(
...
+Java_com_datadoghq_profiler_JavaProfiler_recordTaskBlock0(
+    JNIEnv *env, jclass unused, jlong startTicks, jlong endTicks, jlong spanId,
+    jlong rootSpanId, jlong blocker, jlong unblockingSpanId) {
+  int tid = ProfiledThread::currentTid();
+  if (tid < 0) {
+    return;
+  }
+  Context context = ContextApi::snapshot();
+  context.spanId = (u64)spanId;
+  context.rootSpanId = (u64)rootSpanId;
```
jbachorik: The same - let's just use the generic way of getting and storing context in the events.

ddprof-lib/src/main/cpp/livenessTracker.cpp
```diff
@@ -217,6 +217,13 @@ Error LivenessTracker::initialize(Arguments &args) {
   }
 
   if (_initialized) {
+    // Tracker settings are sticky across recordings. Preserve the historical
+    // table/config behavior, but allow HeapUsage recording to be enabled later
+    // (e.g. if an earlier test initialized liveness without ':L'). Once heap usage
+    // recording has been requested for this JVM lifetime, keep emitting HeapUsage
+    // even if a later recording omits the flag — avoids dropping JVM heap telemetry
+    // when liveness was turned on first without ':L'.
+    _record_heap_usage = _record_heap_usage || args._record_heap_usage;
```
jbachorik: This seems to be not really related (although valid) code change

doc/reference/EventTypeSystem.md
```diff
@@ -232,3 +232,22 @@ The eventTypeFromBCI() conversion function provides a clean bridge between Datad
...
+## Datadog JFR extensions (wall-clock / TaskBlock)
+
+Defined in `ddprof-lib/src/main/cpp/jfrMetadata.h` / `jfrMetadata.cpp`:
+
+- **`datadog.TaskBlock`** (`T_TASK_BLOCK`) — blocking intervals with `blocker` carrying the same
+  **32-bit identity hash bits** as `System.identityHashCode` (JVMTI wait path and JNI park path).
+- **`datadog.WallClockSamplingEpoch`** (`T_WALLCLOCK_SAMPLE_EPOCH`) — timer aggregation snapshot;
+  appended numeric fields must remain backward-compatible for parsers that tolerate trailing varints.
+
+**JVMTI**: `MonitorWait` / `MonitorWaited` are registered only for `java_version() < 21`
```
jbachorik: For Java 21+ there are no such callbacks?

doc/architecture/TLSContext.md
```diff
@@ -500,6 +500,36 @@ Full benchmark data and analysis:
...
+otherwise only Java park-flag suppression applies.
+
+On JDK **21+**, `Thread.sleep` often maps to **CONDVAR_WAIT** (not only legacy **SLEEPING**),
+same broad category as condvar-based park — see `PrecheckTest` and `wallClock.cpp`.
+
+### JVMTI `MonitorWait` / TaskBlock (JDK before 21)
+
+JVMTI `MonitorWait` / `MonitorWaited` callbacks are registered only when
+`java_version() < 21` (`vmEntry.cpp`). They populate TaskBlock with a **blocker** value
+matching **`System.identityHashCode(waitedObject)`** (same convention as the JNI park path).
```
jbachorik: This does not seem to be the right place. This is about the TLS context implementation.

ddprof-test/src/test/java/com/datadoghq/profiler/wallclock/WallClockThreadFilterTest.java
```diff
@@ -38,6 +38,9 @@ public void test() throws InterruptedException {
 
     @Override
     protected String getProfilerCommand() {
-        return "wall=~1ms,filter=0";
+        // wallprecheck=false: this test exercises thread filter + JFR thread identity, not sleep
+        // suppression (see PrecheckTest). With wallprecheck=true and VMThread set on the filter
+        // slot, a thread sleeping the whole window receives no wall signals.
+        return "wall=~1ms,filter=0,wallprecheck=false";
```
jbachorik: This should not be necessary - wallprecheck is false by default, right?

ddprof-lib/src/main/cpp/wallClock.h
```diff
@@ -151,7 +163,7 @@ class WallClockASGCT : public BaseWallClock {
     void timerLoop() override;
 
   public:
-    WallClockASGCT() : BaseWallClock(), _collapsing(false) {}
+    WallClockASGCT() : BaseWallClock(), _collapsing(false), _precheck(true) {}
```
jbachorik: **Constructor default for `_precheck` is inverted**

`WallClockASGCT()` initialises `_precheck` to `true`, but `Arguments._wall_precheck` defaults to `false`. If `initialize(args)` is ever skipped (or a caller constructs the engine without calling it), precheck is silently active, suppressing `SIGVTALRM` for sleeping/parked threads without user opt-in — contrary to the documented opt-in semantics.

Consider matching the `Arguments` default:

```suggestion
    WallClockASGCT() : BaseWallClock(), _collapsing(false), _precheck(false) {}
```

ddprof-lib/src/main/cpp/threadFilter.cpp
```diff
@@ -284,6 +290,48 @@ void ThreadFilter::collect(std::vector<int>& tids) const {
...
+    int num_chunks = _num_chunks.load(std::memory_order_relaxed);
+    for (int chunk_idx = 0; chunk_idx < num_chunks; ++chunk_idx) {
+        ChunkStorage* chunk = _chunks[chunk_idx].load(std::memory_order_acquire);
+        if (chunk == nullptr) continue;
+
+        for (const auto& slot : chunk->slots) {
+            int slot_tid = slot.value.load(std::memory_order_acquire);
+            if (slot_tid != -1) {
+                VMThread* vm = slot.vm_thread.load(std::memory_order_acquire);
+                ProfiledThread* pt = slot.profiled_thread.load(std::memory_order_acquire);
```
jbachorik: **`profiled_thread` is collected but never consumed**

`collectWithState` performs an atomic `acquire` load of `ProfiledThread*` into every `ThreadEntry`, but no caller in the timer path (`sampleThreads`, `wallClock.cpp`) dereferences it. Carrying a raw pointer to a non-lifetime-guaranteed object through the sampling loop is a footgun — a future change that consumes `entry.profiled_thread` will race with `ProfiledThread::release()`/`delete` unless every site adds a lifetime guard.

Either remove the field from `ThreadEntry` (and the corresponding load) until there is a concrete consumer, or document the lifetime contract on the field declaration so future users add a null check before dereferencing.

ddprof-lib/src/main/cpp/thread.h
```diff
@@ -239,6 +259,70 @@ class ProfiledThread : public ThreadLocalData {
...
+    u32 flags = __atomic_load_n(&_misc_flags, __ATOMIC_ACQUIRE);
+    if ((flags & FLAG_PARKED) == 0) {
+      return ParkedWallClockState::NOT_PARKED;
+    }
+    return (flags & FLAG_PARKED_WITH_SPAN) != 0
+        ? ParkedWallClockState::PARKED_ACTIVE_SPAN
+        : ParkedWallClockState::PARKED_SPANLESS;
+  }
+
+  inline void monitorWaitEnter(u64 start_ticks, const Context& context, u64 blocker) {
```
jbachorik: **Misleading comment: `_monitor_wait_active` is not read from the wall-clock path**

The comment claims the wall-clock path reads `_monitor_wait_*` fields, but the signal handler in `wallClock.cpp` never reads any of them. Both `MonitorWait` and `MonitorWaited` JVMTI callbacks fire on the affected (waited) thread itself, so there is no cross-thread access pattern here. Leaving the comment in place implies a memory-ordering constraint that doesn't exist and could mislead future contributors into adding unnecessary synchronisation.

Consider updating it to reflect the actual access pattern, e.g.:

> Written and read only on the waited thread via JVMTI `MonitorWait`/`MonitorWaited` callbacks; no cross-thread access.

ddprof-lib/src/main/cpp/wallClock.cpp
```diff
@@ -165,6 +183,7 @@ bool BaseWallClock::isEnabled() const {
 
 void WallClockASGCT::initialize(Arguments& args) {
   _collapsing = args._wall_collapsing;
+  _precheck   = args._wall_precheck;
```
jbachorik, line 210: **Initialization-ordering reliance is implicit**

`_precheck = args._wall_precheck` correctly mirrors the requested setting, but the constructor's default of `true` (see `wallClock.h`) means there is a window between construction and `initialize(args)` where the field is enabled. Today this is benign because `start()` calls `pthread_create()` only after `initialize()` returns, but the contract is undocumented — any future code that runs the timer loop or invokes `sampleThreads` before `initialize()` will silently observe the wrong default.

No code change required if the existing `initialize() → start()` order is guaranteed, but a brief comment on `initialize()` ("must be called before `start()`/timer-loop entry; otherwise `_precheck` is unset") would lock the invariant in.

jbachorik:
Nit: Either change `_num_skipped_precheck_os` to `u64` to match `putVar64`, or use `putVar32` consistent with the `T_INT` JFR metadata declaration.

ddprof-test/src/test/java/com/datadoghq/profiler/wallclock/SleepTest.java
```diff
@@ -26,6 +26,6 @@ public void testSleep() {
 
     @Override
     protected String getProfilerCommand() {
-        return "wall=1ms";
+        return "wall=1ms,wallprecheck=false";
```
jbachorik: Not really necessary, I assume

ddprof-lib/src/main/cpp/hotspot/vmStructs.inline.h
```diff
@@ -9,12 +9,21 @@
...
 #include "hotspot/vmStructs.h"
 #include "jvmThread.h"
+#include "vmEntry.h"
 
 VMThread* VMThread::current() {
+    // JVMThread::current() is the native thread self pointer. On OpenJ9/Zing it
+    // is not a HotSpot JavaThread*; only HotSpot may reinterpret it as VMThread*.
+    if (!VM::isHotspot() || JVMThread::current() == nullptr) {
+        return nullptr;
+    }
```
jbachorik, line 19: These changes are coincidental, right?

jbachorik:
I did the first pass. I have some concerns about the basics - let's take it off-line and decide how to proceed. Until then you should not spend time on addressing my inline comments.

PR #504, opened 2026-04-28
Implement option to use JVMTI/JFR stack-walker for CPU- and wall-clock-profiler

rkennke:
**What does this PR do?**:

We are working to contribute a JVMTI/JFR-based stack-walker into upstream JDK (see [JEP Draft: Native Profiler Hook for Unbiased Stack Traces (Experimental)](https://bugs.openjdk.org/browse/JDK-8380294)). This feature adds a JVMTI extension to trigger a stack-trace to be emitted via the JVM's JFR. This PR wires that feature to our CPU and wall-clock profilers. When the feature is enabled, and the JVMTI extension is detected, it will prefer that over the other stack-walkers.

**Motivation**:

I would like to be able to test the new JDK feature end-to-end. Eventually, when the feature is released in upstream JDK, it will be ready to go on our side.

**Additional Notes**:

The feature needs to be enabled by the jvmtistacks option.

The stack-traces will be emitted via the JVM's JFR stream. We merge the JVM JFR with the agent's JFR, the backend will receive the combined JFR. I'll post a separate PR in the backend which implements the necessary post-processing.

I also made a small unrelated addition to vmStructs to deal with removed uncompressed class-pointers support in JDK 27.

**How to test the change?**:

I manually verified that the CPU and wall-clock profiler produce the desired AsyncStackTrace events.

**For Datadog employees**:
- [x] If this PR touches code that signs or publishes builds or packages, or handles
- [x] JIRA: [PROF-14350]

[PROF-14350]: https://datadoghq.atlassian.net/browse/PROF-14350?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -1520,6 +1520,7 @@ void Recording::writeEventSizePrefix(Buffer *buf, int start) {
 }
 
 void Recording::recordExecutionSample(Buffer *buf, int tid, u64 call_trace_id,
+                                      u64 correlation_id,
                                       ExecutionEvent *event) {
   int start = buf->skip(1);
   buf->putVar64(T_EXECUTION_SAMPLE);
@@ -1529,12 +1530,14 @@ void Recording::recordExecutionSample(Buffer *buf, int tid, u64 call_trace_id,
   buf->put8(static_cast<int>(event->_thread_state));
   buf->put8(static_cast<int>(event->_execution_mode));
   buf->putVar64(event->_weight);
+  buf->putVar64(correlation_id);
   writeCurrentContext(buf);
   writeEventSizePrefix(buf, start);
   flushIfNeeded(buf);
 }
 
 void Recording::recordMethodSample(Buffer *buf, int tid, u64 call_trace_id,
+                                   u64 correlation_id,
                                    ExecutionEvent *event) {
   int start = buf->skip(1);
   buf->putVar64(T_METHOD_SAMPLE);
@@ -1544,6 +1547,7 @@ void Recording::recordMethodSample(Buffer *buf, int tid, u64 call_trace_id,
   buf->put8(static_cast<int>(event->_thread_state));
   buf->put8(static_cast<int>(event->_execution_mode));
   buf->putVar64(event->_weight);
+  buf->putVar64(correlation_id);
   writeCurrentContext(buf);
   writeEventSizePrefix(buf, start);
   flushIfNeeded(buf);
@@ -1848,11 +1852,11 @@ void FlightRecorder::recordEvent(int lock_index, int tid, u64 call_trace_id,
       RecordingBuffer *buf = rec->buffer(lock_index);
       switch (event_type) {
       case BCI_CPU:
-          rec->recordExecutionSample(buf, tid, call_trace_id,
+          rec->recordExecutionSample(buf, tid, call_trace_id, 0,
                                      (ExecutionEvent *)event);
           break;
         case BCI_WALL:
-          rec->recordMethodSample(buf, tid, call_trace_id,
+          rec->recordMethodSample(buf, tid, call_trace_id, 0,
                                   (ExecutionEvent *)event);
           break;
         case BCI_ALLOC:
@@ -1875,6 +1879,32 @@ void FlightRecorder::recordEvent(int lock_index, int tid, u64 call_trace_id,
         rec->flushIfNeeded(buf);
         rec->addThread(lock_index, tid);
       }
+  }
+}
+
+void FlightRecorder::recordEventDelegated(int lock_index, int tid,
+                                          u64 correlation_id, int event_type,
+                                          Event *event) {
+  OptionalSharedLockGuard locker(&_rec_lock);
+  if (locker.ownsLock()) {
+    Recording* rec = _rec;
+    if (rec != nullptr) {
+      RecordingBuffer *buf = rec->buffer(lock_index);
+      switch (event_type) {
+        case BCI_CPU:
+          rec->recordExecutionSample(buf, tid, 0, correlation_id,
+                                     (ExecutionEvent *)event);
+          break;
+        case BCI_WALL:
+          rec->recordMethodSample(buf, tid, 0, correlation_id,
+                                  (ExecutionEvent *)event);
+          break;
+        default:
+          // Delegation is only wired for CPU/wall samples in v1.
+          break;
+      }
+      rec->addThread(lock_index, tid);
+    }
   }
 }
 
--- a/ddprof-lib/src/main/cpp/hotspot/vmStructs.h
+++ b/ddprof-lib/src/main/cpp/hotspot/vmStructs.h
@@ -300,7 +300,7 @@ typedef void* address;
         field(_flag_type_offset, offset, MATCH_SYMBOLS("_type", "type"))                                            \
     type_end()                                                                                                      \
     type_begin(VMOop, MATCH_SYMBOLS("oopDesc"))                                                                     \
-        field(_oop_klass_offset, offset, MATCH_SYMBOLS("_metadata._klass"))                                         \
+        field(_oop_klass_offset, offset, MATCH_SYMBOLS("_metadata._klass", "_compressed_klass"))                    \
     type_end()                                                                                                      \
     type_begin(VMUniverse, MATCH_SYMBOLS("Universe", "CompressedKlassPointers"))                                    \
         field(_narrow_klass_base_addr, address, MATCH_SYMBOLS("_narrow_klass._base", "_base"))                      \
--- a/ddprof-lib/src/main/cpp/itimer.cpp
+++ b/ddprof-lib/src/main/cpp/itimer.cpp
@@ -17,6 +17,7 @@
 
 #include "itimer.h"
 #include "debugSupport.h"
+#include "jvmThread.h"
 #include "os.h"
 #include "profiler.h"
 #include "stackWalker.h"
@@ -95,4 +96,86 @@ Error ITimer::start(Arguments &args) {
 void ITimer::stop() {
   struct itimerval tv = {{0, 0}, {0, 0}};
   setitimer(ITIMER_PROF, &tv, NULL);
+}
+
+volatile bool ITimerJvmti::_enabled = false;
+long ITimerJvmti::_interval = 0;
+
+void ITimerJvmti::signalHandler(int signo, siginfo_t *siginfo, void *ucontext) {
+  CriticalSection cs;
+  if (!cs.entered()) {
+    return;
+  }
+  int saved_errno = errno;
+  if (!__atomic_load_n(&_enabled, __ATOMIC_ACQUIRE)) {
+    errno = saved_errno;
+    return;
+  }
+  ProfiledThread *current = ProfiledThread::currentSignalSafe();
+  if (current != nullptr && JVMThread::isInitialized() && JVMThread::current() == nullptr
+      && current->inInitWindow()) {
+    current->tickInitWindow();
+    errno = saved_errno;
+    return;
+  }
+  int tid = current ? current->tid() : OS::threadId();
+  if (current) {
+    current->noteCPUSample(Profiler::instance()->recordingEpoch());
+  }
+  Shims::instance().setSighandlerTid(tid);
+
+  ExecutionEvent event;
+  event._execution_mode = getThreadExecutionMode();
+  // setitimer(ITIMER_PROF) delivers SIGPROF to an arbitrary thread chosen by
+  // the OS, so ucontext may be from a JVM-internal thread.  Pass nullptr to
+  // force the JVM into safepoint-based stack walking instead.
+  Profiler::instance()->recordSampleDelegated(nullptr, _interval, tid,
+                                               BCI_CPU, &event);
+  Shims::instance().setSighandlerTid(-1);
+  errno = saved_errno;
+}
+
+Error ITimerJvmti::check(Arguments &args) {
+  if (!VM::canRequestStackTrace()) {
+    return Error("HotSpot RequestStackTrace JVMTI extension not available");
+  }
+
+  struct sigaction oldsa;
+  struct sigaction ign = {};
+  sigemptyset(&ign.sa_mask);
+  ign.sa_handler = SIG_IGN;
+  sigaction(SIGPROF, &ign, &oldsa);
+
+  struct itimerval tv_on = {{1, 0}, {1, 0}};
+  Error err = setitimer(ITIMER_PROF, &tv_on, nullptr) != 0
+      ? Error("ITIMER_PROF is not supported on this system")
+      : Error::OK;
+  struct itimerval tv_off = {{0, 0}, {0, 0}};
+  setitimer(ITIMER_PROF, &tv_off, nullptr);
+
+  sigaction(SIGPROF, &oldsa, nullptr);
+  return err;
+}
+
+Error ITimerJvmti::start(Arguments &args) {
+  if (args._interval < 0) {
+    return Error("interval must be positive");
+  }
+  _interval = args.cpuSamplerInterval();
+
+  OS::installSignalHandler(SIGPROF, signalHandler);
+
+  time_t sec = _interval / 1000000000;
+  suseconds_t usec = (_interval % 1000000000) / 1000;
+  struct itimerval tv = {{sec, usec}, {sec, usec}};
+  if (setitimer(ITIMER_PROF, &tv, nullptr) != 0) {
+    return Error("ITIMER_PROF is not supported on this system");
+  }
+  return Error::OK;
+}
+
+void ITimerJvmti::stop() {
+  struct itimerval tv = {};
+  setitimer(ITIMER_PROF, &tv, nullptr);
+  OS::installSignalHandler(SIGPROF, nullptr, SIG_IGN);
 }
--- a/ddprof-lib/src/main/cpp/profiler.cpp
+++ b/ddprof-lib/src/main/cpp/profiler.cpp
@@ -63,9 +63,12 @@ static Engine noop_engine;
 static MallocTracer malloc_tracer;
 static PerfEvents perf_events;
 static WallClockASGCT wall_asgct_engine;
+static WallClockJvmti wall_jvmti_engine;
 static J9WallClock j9_engine;
 static ITimer itimer;
+static ITimerJvmti itimer_jvmti;
 static CTimer ctimer;
+static CTimerJvmti ctimer_jvmti;
 
 void Profiler::onThreadStart(jvmtiEnv *jvmti, JNIEnv *jni, jthread thread) {
   ProfiledThread::initCurrentThread();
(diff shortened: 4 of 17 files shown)


ddprof-lib/src/main/cpp/vmEntry.cpp
```diff
@@ -426,6 +431,46 @@ bool VM::initProfilerBridge(JavaVM *vm, bool attach) {
...
+          Counters::increment(JVMTI_STACKS_INIT_FAILED);
+        }
+      } else {
+        Log::warn("jvmtistacks requested but HotSpot RequestStackTrace extension "
+                  "is not available on this JVM");
+        Counters::increment(JVMTI_STACKS_INIT_FAILED);
+      }
+    }
+  }
+
```
jbachorik: A stylistic nit - can this be extracted to a function?

ddprof-lib/src/main/cpp/hotspot/vmStructs.h
```diff
@@ -300,7 +300,7 @@ typedef void* address;
         field(_flag_type_offset, offset, MATCH_SYMBOLS("_type", "type"))                                            \
     type_end()                                                                                                      \
     type_begin(VMOop, MATCH_SYMBOLS("oopDesc"))                                                                     \
-        field(_oop_klass_offset, offset, MATCH_SYMBOLS("_metadata._klass"))                                         \
+        field(_oop_klass_offset, offset, MATCH_SYMBOLS("_metadata._klass", "_compressed_klass"))                    \
```
jbachorik, line 303: This is not related directly to the JVMTI stack walker, right?

rkennke: No, this is only there to make it work with JDK 27, which is a prerequisite for the JVMTI stack-walker. Want me to extract this into a separate PR?

jbachorik: No, not necessary. Just a clarification. Thanks!

jbachorik:
**`ddprof-lib/src/main/cpp/profiler.cpp:609`**

I think Error::OK is 0, so the condition is correct. But it might be good to use explicit comparison against Error::OK for posterity

jbachorik:
**`ddprof-lib/src/main/cpp/vmEntry.h:804`**

**Data Race Uninitialized Pointers**

Plain non-atomic reads of _request_stack_trace and _request_stack_trace_initialized in canRequestStackTrace() race with initialization writes in probeJFRRequestStackTrace(). Signal handlers may see partial initialization on other CPUs due to lack of memory ordering. CWE-131 pattern: pointer_race_no_sync.

Use __atomic_load_n for both _request_stack_trace and _request_stack_trace_initialized with __ATOMIC_ACQUIRE semantics in canRequestStackTrace(). Initialize with __atomic_store_n in probeJFRRequestStackTrace().

jbachorik:
**`ddprof-lib/src/main/cpp/profiler.h:99`**

**False Sharing**

_sample_seq is placed adjacent to _total_samples (both write-mostly atomics incremented on every sample), sharing a single cache line. This doubles coherence traffic on multi-socket/many-core hardware.

Separate with explicit padding (alignas(DEFAULT_CACHE_LINE_SIZE) on _sample_seq in profiler.h:99) or co-locate _sample_seq with a read-mostly field so the dirtied line is not also being read by other paths.

jbachorik:
**`ddprof-lib/src/main/cpp/profiler.cpp:597`**

**Contended Atomic**

_sample_seq is incremented with atomicInc (full barrier __sync_fetch_and_add) on every CPU/wall sample from every thread, creating a single globally-contended cache line in the sample hot path. Combined with adjacent _total_samples and adjacent write-mostly operations, this doubles coherence traffic on multi-socket/many-core hardware.

Use atomicIncRelaxed(_sample_seq) (or __atomic_fetch_add(..., __ATOMIC_RELAXED)) like _total_samples; the value is opaque to the JVM and only needs uniqueness, not synchronization. Separate with explicit padding (alignas(DEFAULT_CACHE_LINE_SIZE) on _sample_seq in profiler.h:99) to avoid false-sharing with _total_samples.

jbachorik:
**`ddprof-lib/src/main/cpp/profiler.cpp:597`**

**Integer Overflow**

Correlation ID generation via atomicInc(_sample_seq) + 1 can wrap at u64 boundaries; when _sample_seq reaches ULLONG_MAX, (ULLONG_MAX + 1) overflows to zero, violating the wire-format contract where zero means 'no correlation'. CWE-190. This becomes more severe when profiler restarts and resets _sample_seq to 0 (line 1067), potentially reusing correlation IDs that match stale AsyncStackTrace events from prior JFR recordings (replay attack vector).

Implement saturating increment: u64 seq = atomicIncRelaxed(_sample_seq); u64 corr = (seq == ULLONG_MAX) ? 1 : (seq + 1); to ensure correlation_id is never zero. Alternatively prevent zero by saturating/modular arithmetic on restart.

jbachorik:
**`ddprof-lib/src/main/cpp/profiler.cpp:1032`**

**Misleading Comment**

The new comment block above selectWallEngine's jvmtistacks branch states 'The engine-level _wallclock_sampler knob still takes precedence for users who explicitly request JVMTI/ASGCT', but the code returns wall_jvmti_engine before the switch(args._wallclock_sampler) is even reached, so jvmtistacks in fact OVERRIDES the knob. Documentation contradicts behavior.

Rewrite comment to: 'jvmtistacks overrides _wallclock_sampler when the HotSpot extension is available'.

jbachorik:
**`ddprof-lib/src/main/cpp/wallClock.cpp:878`**

**Signal Safety Errno**

WallClockJvmti::signalHandler does not save/restore errno across the body, unlike the sibling CTimerJvmti and ITimerJvmti handlers added in the same diff. Calls into ProfiledThread, OS::threadId, and recordSampleDelegated may set errno and clobber the interrupted thread's errno value.

Bracket the body with `int saved_errno = errno; ...; errno = saved_errno;` matching the pattern used in CTimerJvmti/ITimerJvmti.

jbachorik:
**`ddprof-lib/src/main/cpp/vmEntry.cpp:765`**

**Null Pointer Dereference**

probeJFRRequestStackTrace() reads args._jvmtistacks (line 763) and calls _init_request_stack_trace->() (line 765) without checking if the InitializeRequestStackTrace function pointer was actually set. If the extension is not found in GetExtensionFunctions, _init_request_stack_trace remains nullptr, causing a null deref.

Add an explicit null check before calling _init_request_stack_trace: 'if (_init_request_stack_trace == nullptr) { Log::warn(...); Counters::increment(JVMTI_STACKS_INIT_FAILED); return; }' after line 764.

jbachorik:
**`ddprof-lib/src/main/cpp/ctimer_linux.cpp:187`**

**Duplication And Lsp**

CTimerJvmti::start duplicates CTimer::start almost verbatim (max_timers/_timers calloc, registerThread loop) but skips OS::primeSignalOriginCheck() and origin validation that CTimer::start/signalHandler perform. Without origin validation, CTimerJvmti::signalHandler processes every SIGPROF without verifying it originates from setitimer(ITIMER_PROF); foreign SIGPROF (e.g. from raise/pthread_kill) will trigger unintended JVMTI stack-trace requests.

Either reuse CTimer::start (e.g. by parameterizing the signal handler) or add OS::primeSignalOriginCheck() call in CTimerJvmti::start() and OS::shouldProcessSignal guard in CTimerJvmti::signalHandler.

jbachorik:
**`ddprof-lib/src/main/cpp/ctimer_linux.cpp:226`**

**Silent Fail**

CTimerJvmti::start() returns Error::OK unconditionally, discarding thread-registration failures from lines 219-222. Swallows per-thread registerThread failures and reports success even when no threads were actually registered.

Change line 226 to `return result;` instead of `return Error::OK;`

jbachorik:
**`ddprof-lib/src/main/cpp/vmEntry.cpp:760`**

**Resource Leak**

GetExtensionFunctions returns a table where each jvmtiExtensionFunctionInfo entry owns sub-allocations (id, short_description, params array, and each param's name); only the top-level table is Deallocated, leaking every per-entry allocation for the lifetime of the JVM.

Iterate ext_functions before deallocating and call _jvmti->Deallocate on each entry's id, short_description, params[j].name for j in [0,param_count), and params itself, then Deallocate(ext_functions).

jbachorik:
**`ddprof-lib/src/main/cpp/arguments.cpp:364`**

**Argument Parsing Fallthrough**

jvmtistacks argument parser has a logic bug: case 'f' falls through to 'default' which sets true, but comments say 'false'. All non-matched characters implicitly set true.

Add explicit break after case 'f': case 'f': _jvmtistacks = false; break;. Replace default with explicit error handling for invalid values.

jbachorik:
**`ddprof-lib/src/main/cpp/profiler.cpp:597`**

**Lock Contention**

recordSampleDelegated attempts tryLock on three different lock indices (582-586) before falling back. If all three fail, the sample is dropped ('dangling AsyncStackTrace'). Under high lock contention, silently loses correlation between profiler events and JFR stack traces.

Add a counter (similar to JVMTI_STACKS_FAILED_OTHER) to track dropped samples due to lock contention, and surface this in the stop() diagnostic warnings.

jbachorik:
**`ddprof-lib/src/main/cpp/profiler.cpp:597`**

**Return Value Semantics**

recordSampleDelegated returns `true` even when the lock could not be acquired and the sample event was dropped (only the dangling JVM-side stack-trace request remains). The header comment says 'true if the request was accepted by the JVM ... and the sample event was recorded' — the sample event was NOT recorded in that branch, so callers cannot rely on `true` meaning what the contract says.

Either return `false` on the lock-skip path so callers can count it as a drop, or update the header doc to: 'true if the JVM accepted the stack-trace request; the local sample event may still have been dropped due to lock contention'.

jbachorik:
**`ddprof-lib/src/main/cpp/profiler.cpp:641`**

**Documentation Accuracy**

Error message at line 675 suggests JFR flag syntax '-XX:StartFlightRecording=...,+jdk.AsyncStackTrace#enabled=true' but standard JFR syntax uses '.enabled=true' not '#enabled=true'.

Replace '+jdk.AsyncStackTrace#enabled=true' with 'jdk.AsyncStackTrace.enabled=true'.

jbachorik:
**`ddprof-lib/src/main/cpp/itimer.cpp:131`**

Maybe add `__atomic_store_n(&_enabled, false, __ATOMIC_RELEASE);` at the top of ITimerJvmti::stop before disarming the timer to close the in-flight signal window.

jbachorik:
**`ddprof-lib/src/main/cpp/wallClock.cpp:843`**

NIT: Add a runtime check in WallClockJvmti::initialize() or add a comment stating that initialization must be guarded by the caller checking VM::canRequestStackTrace() first.

jbachorik:
**`ddprof-lib/src/main/cpp/profiler.cpp:1045`**

**Initialization**

_sample_seq is reset to 0 on each profiler start/reset. No test verifies that the reset does not cause correlation ID collisions if profiler is stopped and restarted with existing JFR recordings.

Add integration test or comment explaining that stale AsyncStackTrace events from prior recording are expected.

rkennke:
> **`ddprof-lib/src/main/cpp/profiler.cpp:609`**
> 
> I think Error::OK is 0, so the condition is correct. But it might be good to use explicit comparison against Error::OK for posterity

It is more complicated than that. We cannot compare with another `Error::OK`, because the `Error` type does not have such a == operator. It *does* have a `operator bool()` and that is what those if statements are using. I would also prefer to use an `== Error::OK` pattern, but this requires to add such operator, and it would then be preferable to change this across the code-base.

rkennke:
> **`ddprof-lib/src/main/cpp/profiler.cpp:597`**
> 
> **Integer Overflow**
> 
> Correlation ID generation via atomicInc(_sample_seq) + 1 can wrap at u64 boundaries; when _sample_seq reaches ULLONG_MAX, (ULLONG_MAX + 1) overflows to zero, violating the wire-format contract where zero means 'no correlation'. CWE-190. This becomes more severe when profiler restarts and resets _sample_seq to 0 (line 1067), potentially reusing correlation IDs that match stale AsyncStackTrace events from prior JFR recordings (replay attack vector).
> 
> Implement saturating increment: u64 seq = atomicIncRelaxed(_sample_seq); u64 corr = (seq == ULLONG_MAX) ? 1 : (seq + 1); to ensure correlation_id is never zero. Alternatively prevent zero by saturating/modular arithmetic on restart.

correlationId=0 is a perfectly fine value, there is no wire-format where zero means no correlationId. When there is no deferred stack-trace, the correlationId attribute itself is absent. I'll change the misleading comment.

rkennke:
> **`ddprof-lib/src/main/cpp/ctimer_linux.cpp:226`**
> 
> **Silent Fail**
> 
> CTimerJvmti::start() returns Error::OK unconditionally, discarding thread-registration failures from lines 219-222. Swallows per-thread registerThread failures and reports success even when no threads were actually registered.
> 
> Change line 226 to `return result;` instead of `return Error::OK;`

This looks like a pre-existing bug in CTimer::start() as well. I'm going to do the fix here.

rkennke:
> **`ddprof-lib/src/main/cpp/itimer.cpp:131`**
> 
> Maybe add `__atomic_store_n(&_enabled, false, __ATOMIC_RELEASE);` at the top of ITimerJvmti::stop before disarming the timer to close the in-flight signal window.

I believe _enabled is already cleared by disableEngines() before calling stop(), so this is not needed.

rkennke:
I addressed all review comments, please re-review. Thanks!

ddprof-lib/src/main/cpp/vmEntry.h
```diff
@@ -190,6 +196,18 @@ class VM {
...
   }
 
+  static bool canRequestStackTrace() {
+    return __atomic_load_n(&_request_stack_trace, __ATOMIC_ACQUIRE) != nullptr;
+  }
+
+  // Must not be called from a signal handler — invokes JVMTI which is not async-signal-safe.
+  static bool initializeRequestStackTrace();
+
+  static jvmtiError requestStackTrace(void* ucontext, jlong user_data) {
```
jbachorik, line 206: **Potential null dereference in requestStackTrace()**

`requestStackTrace()` atomically loads `_request_stack_trace` into a local `fn` variable but does not null-check `fn` before calling it. If `initializeRequestStackTrace()` were ever to clear the pointer concurrently after the caller's `canRequestStackTrace()` check returns true, the call would crash with a null function pointer dereference.

Consider adding a null guard:
```cpp
auto fn = __atomic_load_n(&_request_stack_trace, __ATOMIC_ACQUIRE);
if (!fn) return JVMTI_ERROR_INTERNAL;
return fn(_jvmti, (jthread)nullptr, ucontext, user_data);
```

ddprof-lib/src/main/cpp/wallClock.cpp
```diff
@@ -220,3 +220,123 @@ void WallClockASGCT::timerLoop() {
...
+  CriticalSection cs;
+  if (!cs.entered()) {
+    return;
+  }
+  int saved_errno = errno;
+  ProfiledThread *current = ProfiledThread::currentSignalSafe();
+  if (current != nullptr && JVMThread::isInitialized() && JVMThread::current() == nullptr
+      && current->inInitWindow()) {
+    current->tickInitWindow();
+    return;
```
jbachorik, line 271: **errno not restored on early return**

`WallClockJvmti::signalHandler` saves `errno` on entry but the early-return path here (the `inInitWindow`/`tickInitWindow` path) exits without restoring it, corrupting `errno` for the interrupted thread.

Both `CTimerJvmti::signalHandler` and `ITimerJvmti::signalHandler` correctly restore `errno` on this path — this looks like an oversight.

Add `errno = saved_errno;` before the `return;`.

ddprof-lib/src/main/cpp/itimer.cpp
```diff
@@ -96,3 +97,79 @@ void ITimer::stop() {
...
+  Shims::instance().setSighandlerTid(-1);
+  errno = saved_errno;
+}
+
+Error ITimerJvmti::check(Arguments &args) {
+  if (!VM::canRequestStackTrace()) {
+    return Error("HotSpot RequestStackTrace JVMTI extension not available");
+  }
+
+  OS::installSignalHandler(SIGPROF, nullptr, SIG_IGN);
```
jbachorik: **ITimerJvmti::check() has a persistent side effect**

`check()` installs `SIG_IGN` for `SIGPROF` without saving or restoring the previous handler. If `check()` is called during engine selection but `ITimerJvmti::start()` is never reached — because `CTimerJvmti` was selected instead — `SIGPROF` remains permanently `SIG_IGN` for the process lifetime.

Consider saving the previous handler before the probe and restoring it unconditionally before returning from `check()`, or moving the `SIG_IGN` installation to `start()` where it is committed.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -1856,6 +1860,33 @@ void FlightRecorder::recordEvent(int lock_index, int tid, u64 call_trace_id,
...
+          break;
+        case BCI_WALL:
+          rec->recordMethodSample(buf, tid, 0, correlation_id,
+                                  (ExecutionEvent *)event);
+          break;
+        default:
+          // Delegation is only wired for CPU/wall samples in v1.
+          return;
+      }
+      rec->flushIfNeeded(buf);
```
jbachorik: Remove the explicit `rec->flushIfNeeded(buf)` call from `recordEventDelegated()`; the flush inside `recordExecutionSample`/`recordMethodSample` is sufficient.

Also, do we want to `addThread` if the event recording was not delegated (eg. not CPU/Wall?)

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -1059,6 +1118,10 @@ Error Profiler::start(Arguments &args, bool reset) {
     return error;
   }
 
+  if (args._jvmtistacks && !VM::canRequestStackTrace()) {
```
jbachorik: The retry condition `if (args._jvmtistacks && !VM::canRequestStackTrace())` calls `initializeRequestStackTrace()` only when the pointer is already null (because `canRequestStackTrace()` is false), but `initializeRequestStackTrace()` immediately returns false when `_request_stack_trace` is null, making this an unconditional no-op retry.

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -950,6 +990,20 @@ Engine *Profiler::selectCpuEngine(Arguments &args) {
       }
       TEST_LOG("J9[cpu]=asgct");
     }
+    // Prefer the JVMTI JFR-delegated engine when the HotSpot extension is
+    // available and the user opted into jvmtistacks.  On Linux, CTimerJvmti
+    // uses per-thread CPU timers.  On other platforms (e.g. macOS) it is not
+    // supported, so fall back to ITimerJvmti which uses setitimer(ITIMER_PROF).
+    if (args._jvmtistacks) {
```
jbachorik, line 997: Log a warning before falling through to the ASGCT engine when `jvmtistacks` was requested but both JVMTI CPU engines failed their capability checks.

jbachorik (approved):
Looks good - pending the failing CI runs

PR #511, opened 2026-05-06
fix: own line number table memory to avoid UAF on class unload (PROF-14547)

jbachorik:
**What does this PR do?**:

Fixes SIGSEGV in `Profiler::processCallTraces` (production stack: `processCallTraces` → `Recording::writeStackTraces` → `Recording::writeCpool` → `Recording::finishChunk` → `FlightRecorder::dump` → `Profiler::dump`). Detaches `MethodInfo::_line_number_table->_ptr` lifetime from JVMTI by copying the line number table into a malloc'd buffer at capture time inside `Lookup::fillJavaMethodInfo` and immediately calling `jvmti->Deallocate` on the JVMTI memory. `~SharedLineNumberTable` now calls `free()` on the owned buffer.

**Motivation**:

Production SIGSEGV in `processCallTraces+0xaa` during JFR dump. Root cause: `MethodInfo::_line_number_table->_ptr` held a raw pointer to JVMTI-allocated memory returned by `GetLineNumberTable`. When the underlying class was unloaded, the JVM freed that memory; the next dump's `writeStackTraces` lambda called `MethodInfo::getLineNumber(bci)` on a dangling `_ptr` and crashed.

This is a peer manifestation of (PR #510), which crashes at the symmetric *deallocation* site (`~SharedLineNumberTable` → `Deallocate`). Both stem from the same dangling pointer; this fix addresses the lifetime invariant once and subsumes both crashes. PR #510 stays correct (becomes belt-and-suspenders) but is no longer required for the SIGSEGV fix.

**Additional Notes**:

- Hot-path impact is once per first-time-encountered method per dump cycle (gated by `!mi->_mark` and `mi->_key == 0`). The new work is one `malloc` + `memcpy` of the line number table (typically 16–80 bytes) plus an immediate `Deallocate` — replaces the implicit JVMTI-side allocation that already happened. Not on the signal handler path (`fillJavaMethodInfo` runs on the JFR-writer thread under `_state_lock` + `lockAll()`).
- Memory footprint is unchanged: same logical bytes per method, same lifetime, gated by the existing `AGE_THRESHOLD=3` eviction.
- `malloc` failure is handled silently on the read side (`getLineNumber` returns 0 when `_line_number_table` is null) and logged via `TEST_LOG`.
- `jvmti->Deallocate` failure now logged via `TEST_LOG` (preserves the diagnostic that the destructor previously emitted).
- Reviewed by 4 internal reviewer voices (concurrency, perf, error-paths, tests). Concerns addressed inline; deferred items: JMH benchmark for chunk-write latency on method-churn workloads (follow-up).

**How to test the change?**:

Added regression test `WriteStackTracesAfterClassUnloadTest`:
1. Generate a class with line-numbered methods via ASM.
2. Hammer its methods for ~300ms while CPU profiling at 1ms so traces land in `call_trace_storage` and a `MethodInfo` with `_line_number_table` is created.
3. Drop strong refs and `System.gc()` until the `WeakReference` clears (class actually unloaded); `assumeTrue`-skip if the JVM declined to unload within the deadline.
4. Dump 4 times in tight succession. Without the fix, `getLineNumber` dereferences the freed JVMTI memory → SIGSEGV. With the fix, `_ptr` is a malloc'd copy → safe.

Test set deliberately uses `mcleanup=false` to isolate the failure to the PROF-14547 read path; the PROF-14545 cleanup path is covered separately by `CleanupAfterClassUnloadTest` (PR #510). Best-effort under ASan/UBSan, which amplify the signal even when the freed region happens to read sensible bytes — limitations documented in the test javadoc.

Build verified locally: `:ddprof-lib:assembleRelease` and `:ddprof-test:compileTestJava` both pass on macOS arm64.

**For Datadog employees**:
- [x] JIRA: [PROF-14547](https://datadoghq.atlassian.net/browse/PROF-14547)

[PROF-14547]: https://datadoghq.atlassian.net/browse/PROF-14547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -45,23 +45,12 @@ static const char *const SETTING_RING[] = {NULL, "kernel", "user", "any"};
 static const char *const SETTING_CSTACK[] = {NULL, "no", "fp", "dwarf", "lbr"};
 
 SharedLineNumberTable::~SharedLineNumberTable() {
-  // Always attempt to deallocate if we have a valid pointer
-  // JVMTI spec requires that memory allocated by GetLineNumberTable
-  // must be freed with Deallocate
+  // _ptr is a malloc'd copy of the JVMTI line number table (see
+  // Lookup::fillJavaMethodInfo). Freeing here is independent of class
+  // unload — fixes PROF-14545 (~SharedLineNumberTable Deallocate UAF)
+  // and PROF-14547 (getLineNumber UAF during writeStackTraces).
   if (_ptr != nullptr) {
-    jvmtiEnv *jvmti = VM::jvmti();
-    if (jvmti != nullptr) {
-      jvmtiError err = jvmti->Deallocate((unsigned char *)_ptr);
-      // If Deallocate fails, log it for debugging (this could indicate a JVM bug)
-      // JVMTI_ERROR_ILLEGAL_ARGUMENT means the memory wasn't allocated by JVMTI
-      // which would be a serious bug in GetLineNumberTable
-      if (err != JVMTI_ERROR_NONE) {
-        TEST_LOG("Unexpected error while deallocating linenumber table: %d", err);
-      }
-    } else {
-      TEST_LOG("WARNING: Cannot deallocate line number table - JVMTI is null");
-    }
-    // Decrement counter whenever destructor runs (symmetric with increment at creation)
+    free(_ptr);
     Counters::decrement(LINE_NUMBER_TABLES);
   }
 }
@@ -308,10 +297,30 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
     mi->_type = FRAME_INTERPRETED;
     mi->_is_entry = entry;
     if (line_number_table != nullptr) {
-      mi->_line_number_table = std::make_shared<SharedLineNumberTable>(
-          line_number_table_size, line_number_table);
-      // Increment counter for tracking live line number tables
-      Counters::increment(LINE_NUMBER_TABLES);
+      // Detach from JVMTI lifetime: copy into our own buffer and deallocate
+      // the JVMTI-allocated memory immediately. This keeps _ptr valid even
+      // after the underlying class is unloaded — see PROF-14547.
+      void *owned_table = nullptr;
+      if (line_number_table_size > 0) {
+        size_t bytes = (size_t)line_number_table_size * sizeof(jvmtiLineNumberEntry);
+        owned_table = malloc(bytes);
+        if (owned_table != nullptr) {
+          memcpy(owned_table, line_number_table, bytes);
+        } else {
+          TEST_LOG("Failed to allocate %zu bytes for line number table copy", bytes);
+        }
+      }
+      jvmtiError dealloc_err = jvmti->Deallocate((unsigned char *)line_number_table);
+      if (dealloc_err != JVMTI_ERROR_NONE) {
+        TEST_LOG("Unexpected error while deallocating linenumber table: %d", dealloc_err);
+      }
+      line_number_table = nullptr;
+      if (owned_table != nullptr) {
+        mi->_line_number_table = std::make_shared<SharedLineNumberTable>(
+            line_number_table_size, owned_table);
+        // Increment counter for tracking live line number tables
+        Counters::increment(LINE_NUMBER_TABLES);
+      }
     }
 
     // strings are null or came from JVMTI
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -313,7 +313,6 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
       if (dealloc_err != JVMTI_ERROR_NONE) {
         TEST_LOG("Unexpected error while deallocating linenumber table: %d", dealloc_err);
       }
-      line_number_table = nullptr;
       if (owned_table != nullptr) {
         mi->_line_number_table = std::make_shared<SharedLineNumberTable>(
             line_number_table_size, owned_table);
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -47,8 +47,7 @@ static const char *const SETTING_CSTACK[] = {NULL, "no", "fp", "dwarf", "lbr"};
 SharedLineNumberTable::~SharedLineNumberTable() {
   // _ptr is a malloc'd copy of the JVMTI line number table (see
   // Lookup::fillJavaMethodInfo). Freeing here is independent of class
-  // unload — fixes PROF-14545 (~SharedLineNumberTable Deallocate UAF)
-  // and PROF-14547 (getLineNumber UAF during writeStackTraces).
+  // unload, preventing use-after-free in ~SharedLineNumberTable and getLineNumber.
   if (_ptr != nullptr) {
     free(_ptr);
     Counters::decrement(LINE_NUMBER_TABLES);
@@ -299,7 +298,7 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
     if (line_number_table != nullptr) {
       // Detach from JVMTI lifetime: copy into our own buffer and deallocate
       // the JVMTI-allocated memory immediately. This keeps _ptr valid even
-      // after the underlying class is unloaded — see PROF-14547.
+      // after the underlying class is unloaded.
       void *owned_table = nullptr;
       if (line_number_table_size > 0) {
         size_t bytes = (size_t)line_number_table_size * sizeof(jvmtiLineNumberEntry);
@@ -938,8 +937,8 @@ void Recording::writeSettings(Buffer *buf, Arguments &args) {
   writeBoolSetting(buf, T_MALLOC, "enabled", args._nativemem >= 0);
   if (args._nativemem >= 0) {
     writeIntSetting(buf, T_MALLOC, "nativemem", args._nativemem);
-    // samplingInterval=-1 means every allocation is recorded (nativemem=0).
-    writeIntSetting(buf, T_MALLOC, "samplingInterval", args._nativemem == 0 ? -1 : args._nativemem);
+    // samplingInterval=-1 signals "record every allocation"; mirrors shouldSample's interval<=1 threshold.
+    writeIntSetting(buf, T_MALLOC, "samplingInterval", args._nativemem <= 1 ? -1 : args._nativemem);
   }
 
   writeBoolSetting(buf, T_ACTIVE_RECORDING, "debugSymbols",
--- a/ddprof-test/src/test/java/com/datadoghq/profiler/memleak/WriteStackTracesAfterClassUnloadTest.java
+++ b/ddprof-test/src/test/java/com/datadoghq/profiler/memleak/WriteStackTracesAfterClassUnloadTest.java
@@ -93,41 +93,47 @@ public class WriteStackTracesAfterClassUnloadTest extends AbstractProfilerTest {
       // mcleanup=false isolates the test to the PROF-14547 read path; the
       // PROF-14545 cleanup path is covered separately by CleanupAfterClassUnloadTest.
       profiler.execute("start,cpu=1ms,jfr,mcleanup=false,file=" + baseFile.toAbsolutePath());
-      Thread.sleep(200); // let the profiler stabilize
+      try {
+        Thread.sleep(200); // let the profiler stabilize
 
-      // 1. Load a class, hammer its methods on this thread so the CPU profiler captures
-      //    many traces referencing it. Drop strong refs and return only a WeakReference.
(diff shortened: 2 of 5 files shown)


jbachorik:
@codex review

jbachorik:
@codex review

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -45,23 +45,12 @@ static const char *const SETTING_RING[] = {NULL, "kernel", "user", "any"};
 static const char *const SETTING_CSTACK[] = {NULL, "no", "fp", "dwarf", "lbr"};
 
 SharedLineNumberTable::~SharedLineNumberTable() {
-  // Always attempt to deallocate if we have a valid pointer
-  // JVMTI spec requires that memory allocated by GetLineNumberTable
-  // must be freed with Deallocate
+  // _ptr is a malloc'd copy of the JVMTI line number table (see
+  // Lookup::fillJavaMethodInfo). Freeing here is independent of class
+  // unload — fixes PROF-14545 (~SharedLineNumberTable Deallocate UAF)
```
jbachorik: Do not refer to closed JIRA tickets

jbachorik: Addressed — JIRA ticket references removed from all affected comments in `flightRecorder.cpp` and `flightRecorder.h`.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -906,6 +915,12 @@ void Recording::writeSettings(Buffer *buf, Arguments &args) {
 
   writeBoolSetting(buf, T_ALLOC, "enabled", args._record_allocations);
   writeBoolSetting(buf, T_HEAP_LIVE_OBJECT, "enabled", args._record_liveness);
+  writeBoolSetting(buf, T_MALLOC, "enabled", args._nativemem >= 0);
+  if (args._nativemem >= 0) {
+    writeIntSetting(buf, T_MALLOC, "nativemem", args._nativemem);
+    // samplingInterval=-1 means every allocation is recorded (nativemem=0).
+    writeIntSetting(buf, T_MALLOC, "samplingInterval", args._nativemem == 0 ? -1 : args._nativemem);
```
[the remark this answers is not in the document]
jbachorik: Addressed — `samplingInterval=-1` now also fires when `nativemem=1` (interval ≤ 1), matching `shouldSample`'s `<= 1` threshold.

gradle/libs.versions.toml
```diff
@@ -21,6 +21,11 @@ zstd = "1.5.7-7"
 # Code analysis
 asm = "9.9.1"
 
+# Datadog tracer (compileOnly for the chaos harness — runtime is provided by dd-java-agent).
+# Dynamic version: always resolve to the latest released dd-trace-api so the chaos harness
+# tracks current annotation/API surface without manual bumps.
+dd-trace-api = "latest.release"
```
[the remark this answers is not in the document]
jbachorik: Fixed — `dd-trace-api` entry has been removed from the version catalog.

.gitlab/reliability/.gitlab-ci.yml
```diff
@@ -56,8 +56,61 @@ reliability-aarch64:
   image: $BENCHMARK_IMAGE_ARM64
   variables:
     ARCH: aarch64
```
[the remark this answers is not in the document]
jbachorik: Fixed — `KUBERNETES_MEMORY_REQUEST` and `KUBERNETES_MEMORY_LIMIT` are restored for the aarch64 job.

ddprof-lib/src/main/cpp/mallocTracer.h
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    Error start(Arguments& args);
+    void stop();
+
+    static inline bool running() {
+        return __atomic_load_n(&_running, __ATOMIC_ACQUIRE);
+    }
+
+    static inline void installHooks() {
```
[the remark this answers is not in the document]
jbachorik: The hooks themselves gate on `MallocTracer::running()` before recording any data, so GOT-patching while stopped adds zero overhead. `patchLibraries()` already short-circuits when `_orig_malloc == NULL`, and `_patched_libs` is monotonic — re-patching an already-patched slot is a no-op.

ddprof-lib/src/main/cpp/mallocTracer.cpp
```diff
@@ -0,0 +1,403 @@
...
+#include <string.h>
+#include "codeCache.h"
+#include "libraries.h"
+#include "mallocTracer.h"
+#include "os.h"
+#include "pidController.h"
+#include "profiler.h"
+#include "symbols.h"
+#include "tsc.h"
+#include "vmEntry.h"
```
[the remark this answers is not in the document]
jbachorik: Will update the PR description to enumerate all changes: the `SharedLineNumberTable` UAF fix, the new `MallocTracer` nativemem engine, and the chaos-testing harness.

ddprof-lib/src/main/cpp/mallocTracer.cpp
```diff
@@ -0,0 +1,403 @@
...
+        event._address = (uintptr_t)address;
+        event._size = size;
+        // _interval == 0 means sample every allocation; weight is 1.0.
+        if (size == 0 || current_interval <= 1) {
+            event._weight = 1.0f;
+        } else {
+            event._weight = (float)(1.0 / (1.0 - exp(-(double)size / (double)current_interval)));
+        }
+
+        Profiler::instance()->recordSample(NULL, size, OS::threadId(), BCI_NATIVE_MALLOC, 0, &event);
```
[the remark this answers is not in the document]
jbachorik: Acknowledged as a known limitation of the GOT-hook approach — native frames below `malloc_hook` are not captured with `cstack=fp/dwarf` because there is no async-signal context. Tracked as a follow-up item.

ddprof-test/src/test/java/com/datadoghq/profiler/memleak/WriteStackTracesAfterClassUnloadTest.java
```diff
@@ -0,0 +1,220 @@
...
+      }
+
+      // 4. If the profiler crashed inside processCallTraces the JVM would have died and we
+      //    would never reach this line. The non-empty-output assertion is secondary — the
+      //    primary signal is reaching profiler.stop() at all.
+      profiler.stop();
+
+      assertTrue(Files.size(dumpFile) > 0,
+          "Profiler produced no output — SIGSEGV during writeStackTraces is suspected");
+
```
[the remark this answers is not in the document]
jbachorik: Fixed — `profiler.stop()` and `profiler.resetThreadContext()` are now called in a `finally` block.

ddprof-test/src/test/java/com/datadoghq/profiler/memleak/WriteStackTracesAfterClassUnloadTest.java
```diff
@@ -0,0 +1,220 @@
...
+  public void testNoSigsegvInWriteStackTracesAfterClassUnload() throws Exception {
+    stopProfiler();
+
+    Path baseFile = tempFile("prof14547-base");
+    Path dumpFile = tempFile("prof14547-dump");
+
+    try {
+      // mcleanup=false isolates the test to the PROF-14547 read path; the
+      // PROF-14545 cleanup path is covered separately by CleanupAfterClassUnloadTest.
+      profiler.execute("start,cpu=1ms,jfr,mcleanup=false,file=" + baseFile.toAbsolutePath());
```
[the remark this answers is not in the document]
jbachorik: Fixed — same `finally`-block fix applies to this call site.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -308,10 +296,30 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
...
+          memcpy(owned_table, line_number_table, bytes);
+        } else {
+          TEST_LOG("Failed to allocate %zu bytes for line number table copy", bytes);
+        }
+      }
+      jvmtiError dealloc_err = jvmti->Deallocate((unsigned char *)line_number_table);
+      if (dealloc_err != JVMTI_ERROR_NONE) {
+        TEST_LOG("Unexpected error while deallocating linenumber table: %d", dealloc_err);
+      }
+      line_number_table = nullptr;
```
rkennke: `line_number_table` is a local variable, so it has no effect on the caller and won't prevent double-free. It's harmless but slightly misleading — could be removed.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -913,8 +921,8 @@ void Recording::writeSettings(Buffer *buf, Arguments &args) {
   writeBoolSetting(buf, T_MALLOC, "enabled", args._nativemem >= 0);
   if (args._nativemem >= 0) {
     writeIntSetting(buf, T_MALLOC, "nativemem", args._nativemem);
-    // samplingInterval=-1 means every allocation is recorded (nativemem=0).
-    writeIntSetting(buf, T_MALLOC, "samplingInterval", args._nativemem == 0 ? -1 : args._nativemem);
+    // samplingInterval=-1 signals "record every allocation"; mirrors shouldSample's interval<=1 threshold.
+    writeIntSetting(buf, T_MALLOC, "samplingInterval", args._nativemem <= 1 ? -1 : args._nativemem);
```
rkennke, line 940: This is unrelated - right?

ddprof-test/src/test/java/com/datadoghq/profiler/memleak/WriteStackTracesAfterClassUnloadTest.java
```diff
@@ -0,0 +1,224 @@
...
+        }
+      }
+
+    } finally {
+      try { Files.deleteIfExists(baseFile); } catch (IOException ignored) {}
+      try { Files.deleteIfExists(dumpFile); } catch (IOException ignored) {}
+    }
+  }
+
+  private WeakReference<ClassLoader> loadAndProfileDynamicClass() throws Exception {
```
rkennke, line 134: There is a lot of duplication with CleanupAfterClassUnloadTest here (generateClassByteCode, IsolatedClassLoader, tempFile, classCounter and much of loadAndProfileDynamicClass), might be worth to use an abstract shared base class.

ddprof-test/src/test/java/com/datadoghq/profiler/memleak/WriteStackTracesAfterClassUnloadTest.java
```diff
@@ -0,0 +1,224 @@
...
+      mv.visitMaxs(0, 0);
+      mv.visitEnd();
+    }
+
+    cw.visitEnd();
+    return cw.toByteArray();
+  }
+
+  private static Path tempFile(String prefix) throws IOException {
+    Path dir = Paths.get("/tmp/recordings");
```
rkennke: Same issue as in #510, might want to use System.getProperty("java.io.tmpdir") instead.

rkennke (changes requested):
Looks good overall, I only have a few minor comments.

ddprof-lib/src/main/cpp/flightRecorder.cpp
jbachorik: Done — removed the dead assignment.

ddprof-test/src/test/java/com/datadoghq/profiler/memleak/WriteStackTracesAfterClassUnloadTest.java
jbachorik: Extracted to a new `AbstractDynamicClassTest` base class; both `GetLineNumberTableLeakTest` and `WriteStackTracesAfterClassUnloadTest` now extend it, sharing `generateClassBytecode`, `tempFile`, `IsolatedClassLoader`, and the `classCounter` field.

ddprof-test/src/test/java/com/datadoghq/profiler/memleak/WriteStackTracesAfterClassUnloadTest.java
jbachorik: `tempFile()` in the new `AbstractDynamicClassTest` base class now uses `Paths.get(System.getProperty("java.io.tmpdir"), "recordings")`.

ddprof-lib/src/main/cpp/flightRecorder.cpp
jbachorik: Yes, intentional and related. The `_nativemem <= 1 ? -1` threshold aligns the recorded `samplingInterval` setting with actual sampling behaviour: `shouldSample()` treats both `nativemem=0` and `nativemem=1` as "record every allocation" (`interval <= 1`). Using `== 0` made the emitted configuration inconsistent with what the profiler actually does at `nativemem=1`.

rkennke (approved):
Thank you! Looks good, now!

jbachorik:
Subsumed into #510 — all commits merged into muse/sigsegv-in-recording.

PR #514, opened 2026-05-07
fix: route plugin and build-logic deps through mavenRepositoryProxy

robayu:
## Summary

- The `mavenRepositoryProxy` Gradle property was only consulted in the root `dependencyResolutionManagement`. Plugin classpath resolution and the `build-logic` included build still hit `gradlePluginPortal()`/`mavenCentral()` directly, so on CI runners that restrict egress to the Depot mirror those fetches fail with `403 Forbidden`.
- This caused `:build-logic:conventions:compileKotlin` to fail downloading `kotlin-build-tools-impl-2.3.0.jar` and `kotlin-reflect-1.6.10.jar` (sample failure: [job 1664033372](https://gitlab.ddbuild.io/DataDog/java-profiler/-/jobs/1664033372)).
- Wires the proxy into three places so plugin and `build-logic` resolution also flow through Depot when the property is set:
  - root `pluginManagement` in `settings.gradle.kts`
  - `build-logic/settings.gradle` `pluginManagement` (the included build had none)
  - `repositories {}` in `build-logic/conventions/build.gradle.kts` (the actual failing config: `kotlinBuildToolsApiClasspath`)

The proxy is only added when the property is set, so local builds without `-PmavenRepositoryProxy=...` continue to behave as before. CI exports `ORG_GRADLE_PROJECT_mavenRepositoryProxy` already, so no `.gitlab-ci.yml` changes are needed.

## Test plan

- [ ] CI pipeline on this branch reaches `:build-logic:conventions:compileKotlin` and resolves `kotlin-build-tools-impl` / `kotlin-reflect` through `depot-read-api-java.us1.ddbuild.io` (no 403s in the build log).
- [ ] Local `./gradlew tasks` (without the property set) still succeeds — proxy block is no-op when the property is absent.

### DIFF
--- a/build-logic/conventions/build.gradle.kts
+++ b/build-logic/conventions/build.gradle.kts
@@ -3,6 +3,10 @@ plugins {
 }
 
 repositories {
+    val mavenRepositoryProxy = providers.gradleProperty("mavenRepositoryProxy").orNull
+    if (mavenRepositoryProxy != null) {
+        maven { url = uri(mavenRepositoryProxy) }
+    }
     gradlePluginPortal()
     mavenCentral()
 }
--- a/build-logic/settings.gradle
+++ b/build-logic/settings.gradle
@@ -1,3 +1,14 @@
+pluginManagement {
+    def mavenRepositoryProxy = providers.gradleProperty('mavenRepositoryProxy').orNull
+    repositories {
+        if (mavenRepositoryProxy != null) {
+            maven { url = uri(mavenRepositoryProxy) }
+        }
+        gradlePluginPortal()
+        mavenCentral()
+    }
+}
+
 rootProject.name = 'build-logic'
 
 include 'conventions'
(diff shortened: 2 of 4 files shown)


robayu:
Good catch — pushed [5387d26e](https://github.com/DataDog/java-profiler/pull/514/commits/5387d26e) wiring `mavenRepositoryProxy` into the root `buildscript.repositories` block ahead of `mavenCentral()`/`gradlePluginPortal()`, so `com.dipien:semantic-version-gradle-plugin` now resolves through Depot when the property is set.

Note: the `allprojects { repositories { ... } }` block and the root project's own `repositories { ... }` block further down don't actually fetch anything because `settings.gradle.kts` sets `RepositoriesMode.PREFER_SETTINGS` — under that mode project-level repositories are ignored (with a warning) and the settings-level resolution wins. So those don't need to change for the 403 issue. Happy to clean them up in a separate PR if you'd like.

jbachorik (approved):

PR #525, opened 2026-05-12
fix(profiler): TripleBufferedDictionary — fix vtable-stub class lookup and counter tracking

jbachorik:
**What does this PR do?**:

Replaces `DoubleBufferedDictionary` with `TripleBufferedDictionary` backed by a generic `TripleBufferRotator<T>` template, fixing two regressions introduced by the double-buffered class/endpoint/context maps:

1. **Vtable-stub class lookup regression** (`walkVM` `bounded_lookup`): with two buffers, class names inserted by `fillJavaMethodInfo()` only land in the dump buffer and are invisible to the active buffer between dumps. The triple-buffer adds a `_recent_ptr` (last completed dump's snapshot) that `bounded_lookup(size_limit=0)` falls back to when the active buffer misses. The fallback is guarded by `RefCountGuard` with TOCTOU revalidation (pointer-first protocol), identical to the `CallTraceStorage` pattern.

2. **`dictionary_classes_keys` counter always 0** during wall-clock profiling: fill-path inserts went to the standby buffer whose counter id was 0 (not the named slot). All three `TripleBufferedDictionary` buffers now carry the real counter id so every insert is counted correctly.

Also extracts `RefCountGuard` into a standalone `refCountGuard.h` (generalised from `CallTraceHashTable*` to `void*`) and the triple-buffer rotation logic into a generic `TripleBufferRotator<T>` template in `tripleBuffer.h`, shared between `TripleBufferedDictionary` and (existing) `CallTraceStorage`.

**Motivation**:

Follow-up fixes for PR #524. Synthetic-class churn means we cannot keep the class map growing forever; the triple-buffer bounds memory to ~2× one dump cycle's data and purges dead class names after at most one full dump cycle.

**Additional Notes**:

- `clearStandby()` uses `RefCountGuard::waitForRefCountToClear(old_recent)` instead of `lockAll()`/`unlockAll()` — the lock-free drain is sufficient and avoids stalling all signal handlers.
- `bounded_lookup(size_limit=0)` is read-only (no malloc) and therefore signal-safe; the recent-fallback path preserves this property.
- The three buffer roles cycle: active → dump → recent → (cleared). At most two non-empty buffers exist at any time.

**How to test the change?**:

- `ddprof-lib:gtestDebug_dictionary_ut` — fully rewritten for `TripleBufferedDictionary` semantics; includes `BoundedLookupReadOnlyFallsBackToRecent` which verifies keys survive exactly one dump cycle then disappear.
- `DictionaryRotationTest` (Java integration) — asserts counter resets after `clearStandby` and correct counts after fill-path inserts.

**For Datadog employees**:

- [ ] JIRA: [JIRA-XXXX]


jbachorik:
Superseded by PR #524 — the TripleBufferedDictionary commit was pushed directly to that branch.

ddprof-lib/src/main/cpp/tripleBuffer.h
```diff
@@ -0,0 +1,88 @@
...
+
+    // Last completed dump's buffer.  Stable between advanceRecent() calls.
+    T* recent() const {
+        return __atomic_load_n(const_cast<T* const volatile*>(&_recent_ptr), __ATOMIC_ACQUIRE);
+    }
+
+    // Advance _active_index by 1 mod 3.
+    // After this call the old active is accessible via dumpBuffer().
+    void rotate() {
+        int old = __atomic_load_n(&_active_index, __ATOMIC_ACQUIRE);
```
jbachorik, line 64: **Non-atomic rotate() — concurrent callers can alias the same index**

`rotate()` performs a non-atomic load-then-store on `_active_index`. If `stop()` and `dump()` race to call it concurrently, both can read the same old value and advance to the same next index, silently skipping a rotation step.

Applied as code fix — replaced with a CAS loop so each concurrent caller advances by exactly one step.

ddprof-lib/src/main/cpp/dictionary.h
```diff
@@ -84,6 +87,108 @@ class Dictionary {
...
+        // restate explicitly for clarity and belt-and-suspenders.
+        Counters::set(DICTIONARY_KEYS,       0,                 _counter_id);
+        Counters::set(DICTIONARY_KEYS_BYTES, 0,                 _counter_id);
+        Counters::set(DICTIONARY_BYTES,      sizeof(DictTable), _counter_id);
+        Counters::set(DICTIONARY_PAGES,      1,                 _counter_id);
+    }
+
+    // Clears all three buffers. Call on profiler reset (no concurrent writers).
+    void clearAll() {
+        _a.clear();
```
jbachorik, line 187: **DICTIONARY_KEYS counter zeroed incorrectly in clearStandby()**

`clearStandby()` calls `Dictionary::clear()` which resets the shared `DICTIONARY_KEYS` counter slot to zero.  Any insertions into the current active buffer that happened after the preceding `rotate()` — but before `clearStandby()` — are silently discarded from the counter.

> Fixed per user guidance: "Fix the keys tracking"

Applied as code fix — after `clear()` the counter is recalibrated to `active()->size()` so monitoring sees only live entries.

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -1300,6 +1298,12 @@ Error Profiler::stop() {
   // correct counts in the recording
   _thread_info.reportCounters();
 
+  // Promote accumulated writes to standby so that writeCpool() (called from
+  // ~Recording() inside _jfr.stop()) reads a complete, stable snapshot.
+  _class_map.rotate();
```
jbachorik, line 1303: **stop() never reclaims standby buffers — memory held until next restart**

`stop()` rotates all three maps to give JFR a stable snapshot, but never calls `clearStandby()`. The standby buffers (holding the last pre-stop entries) are not freed until `clearAll()` on the next `start()`, leaking memory for the full IDLE period.

Applied as code fix — added `clearStandby()` calls after `_jfr.stop()`, mirroring the `dump()` path.

ddprof-test/src/test/java/com/datadoghq/profiler/metadata/DictionaryRotationTest.java
```diff
@@ -0,0 +1,119 @@
...
+        String[] postDump = { "ep_post_0", "ep_post_1" };
+        int sizeLimit = 100;
+
+        for (int i = 0; i < preDump.length; i++) {
+            profiler.recordTraceRoot(i, preDump[i], null, sizeLimit);
+        }
+
+        // dump() triggers: rotate() → lockAll() → jfr.dump(snapshot) → unlockAll()
+        //                            → clearStandby() (resets counter to 0, frees pre-dump buffer)
+        Path snapshot = Files.createTempFile(Paths.get("/tmp/recordings"), "DictionaryRotation_snapshot_", ".jfr");
```
jbachorik, line 60: **Hardcoded /tmp/recordings path may not exist in CI**

`Paths.get("/tmp/recordings")` fails with `NoSuchFileException` in CI environments that do not pre-create this directory.

Applied as code fix — changed to `Files.createTempFile("DictionaryRotation_snapshot_", ".jfr")` with no directory argument, which uses the JVM default temp dir.

ddprof-test/src/test/java/com/datadoghq/profiler/metadata/DictionaryRotationTest.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.openjdk.jmc.common.item.Attribute.attr;
+import static org.openjdk.jmc.common.unit.UnitLookup.PLAIN_TEXT;
+
+/**
+ * Verifies that the DoubleBufferedDictionary rotate+clearStandby cycle correctly:
```
jbachorik, line 38: **Stale class name in Javadoc**

The class Javadoc says `DoubleBufferedDictionary` but the production class is `TripleBufferedDictionary`.

Applied as code fix.

ddprof-lib/src/main/cpp/callTraceStorage.cpp
```diff
@@ -266,10 +266,10 @@ CallTraceStorage::CallTraceStorage() : _active_storage(nullptr), _standby_storag
     _preserve_set_buffer.rehash(static_cast<size_t>(1024 / 0.75f));
 
     // Initialize triple-buffered storage
-    auto active_table = std::make_unique<CallTraceHashTable>();
-    active_table->setInstanceId(getNextInstanceId());
-    active_table->setParentStorage(this);
-    __atomic_store_n(&_active_storage, active_table.release(), __ATOMIC_RELEASE);
+    auto active_ptr = std::make_unique<CallTraceHashTable>();
```
jbachorik, line 269: **Inconsistent naming: active_ptr / standby_table in the same block**

`active_ptr` was renamed but the adjacent `standby_table` was not, creating an inconsistent naming convention in the same initialization block.

Applied as code fix — renamed `standby_table` to `standby_ptr`.

ddprof-lib/src/main/cpp/dictionary.h
```diff
@@ -84,6 +87,108 @@ class Dictionary {
...
 
   void collect(std::map<unsigned int, const char *> &map);
+
+  int  counterId() const { return _id; }
+  void setCounterId(int id) { _id = id; }
+  int  size() const { return _size; }
+};
+
+// Triple-buffered wrapper for signal-handler-safe concurrent dictionary access.
+//
```
jbachorik, line 97: **Dictionary::_id should be const**

`_id` was changed from `const int` to `int` to support `setCounterId()`, but `setCounterId()` is not called anywhere in production or test code, so the non-const field only weakens the compile-time immutability invariant.

Applied as code fix — restored `const int _id` and removed unused `setCounterId()`.

ddprof-lib/src/main/cpp/refCountGuard.h
```diff
@@ -0,0 +1,84 @@
...
+ * - Constructor: Store pointer first, then increment count
+ * - Destructor: Decrement count first, then clear pointer
+ * - Scanner: Check count first (if 0, slot is inactive)
+ *
+ * This ordering ensures no window where scanner incorrectly believes a slot
+ * is inactive when it should be protecting a resource.
+ */
+struct alignas(DEFAULT_CACHE_LINE_SIZE) RefCountSlot {
+    volatile uint32_t count;       // Reference count (0 = inactive)
+    char _padding1[4];             // Alignment padding for pointer
```
jbachorik, line 27: **Hard-coded _padding1[4] assumes 64-bit ABI**

`_padding1[4]` aligns `active_ptr` on an 8-byte boundary on 64-bit, but on a 32-bit ABI where `void*` is 4 bytes the padding is unnecessary and the `remaining[]` calculation silently over-fills.

Applied as code fix — replaced with `alignas(alignof(void*)) void* active_ptr` and a portable padding formula `DEFAULT_CACHE_LINE_SIZE - alignof(void*) - sizeof(void*)`.

jbachorik:
**RefCountGuard activation window — trace-drop, not UAF**

The `RefCountGuard` constructor stores the active pointer before incrementing the reference count. There is a theoretical window between these two operations where a scanner (calling `waitForRefCountToClear`) could see `count=0` and skip the slot, allowing `clearStandby()` to proceed two dump cycles later while the slot is in the process of being activated.

**In practice this window is never hit for signal handlers**: signal handlers complete within microseconds, while `clearStandby()` can only be called after two full dump cycles (typically seconds apart).

**Observable effect if the window were hit**: `bounded_lookup` returns `INT_MAX` (miss) — a dropped trace or a generic vtable frame, not a crash or heap corruption. This is the graceful degradation path already in use today.

> Fixed per user guidance: confirmed as observable incorrect-state / trace-drop, not a hard UAF. Updated comments in `refCountGuard.h` and `TripleBufferedDictionary` to document the activation window and its bounded, graceful consequence.

PR #526, opened 2026-05-13
fix(profiler): close SIGVTALRM race in pthread_create wrapper teardown (PROF-14603)

jbachorik:
## Why

`JavaThread::~JavaThread` / `OSThread::~OSThread` crashed (SIGSEGV) on JDK 25
when the ddprof `pthread_create` hook delivered SIGVTALRM between
`Profiler::unregisterThread()` returning and `ProfiledThread::release()`
acquiring its internal guard. The signal handler called
`currentSignalSafe()` and dereferenced the now-freed `ProfiledThread`
pointer.

## What

- **`libraryPatcher_linux.cpp`**: extract `unregister_and_release(tid)` — an
  `__attribute__((noinline))` helper that holds a `SignalBlocker` for the
  entire unregister+release sequence. Both `start_routine_wrapper` and
  `start_routine_wrapper_spec` call it from their RAII `struct Cleanup`
  destructor (from PR #510), so the race window is eliminated without
  duplicating the signal-masking logic.
- **`perfEvents_linux.cpp`**: same `SignalBlocker` wrapping for the
  `pthread_setspecific_hook` teardown path.
- **`thread.h`**: guard `clearCurrentThreadTLS()` with `#ifdef UNIT_TEST`
  so it is absent from production builds.
- **`GtestTaskBuilder.kt`**: add `-DUNIT_TEST` to gtest compiler flags so
  the guarded method compiles in test builds.
- **`thread_teardown_safety_ut.cpp`**: new acceptance-test suite
  (`ThreadTeardownSafetyTest`, T-01…T-10) covering the full teardown
  lifecycle under signal load.

## Depends on

Stacks on #510 (`muse/sigsegv-in-recording`). Should be retargeted to
`main` once #510 merges.

## Test plan

- [ ] `./gradlew :ddprof-lib:gtestRelease` passes on Linux/glibc and musl
- [ ] `./gradlew :ddprof-lib:gtestDebug` passes on Linux/glibc and musl
- [ ] `ThreadTeardownSafetyTest` T-01…T-10 all green
- [ ] No regression in `ForcedUnwindTest` suite
- [ ] CI chaos/reliability run shows no SIGSEGV in thread teardown

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/.github/actions/setup_cached_java/action.yml
+++ b/.github/actions/setup_cached_java/action.yml
@@ -27,7 +27,7 @@ runs:
           fi
       - name: Cache Build JDK [${{ inputs.arch }}]
         id: cache_build_jdk
-        uses: actions/cache/restore@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
         with:
           path: |
             jdks/${{ steps.infer_build_jdk.outputs.build_jdk }}
@@ -37,7 +37,7 @@ runs:
           enableCrossOsArchive: true
       - name: Cache JDK ${{ inputs.version }} [${{ inputs.arch }}]
         id: cache_jdk
-        uses: actions/cache/restore@0057852bfaa89a56745cba8c7296529d2fc39830 # v4.3.0
+        uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
         with:
           path: |
             jdks/jdk${{ inputs.version }}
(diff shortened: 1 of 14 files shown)


jbachorik:
Closing in favour of #529, which is a focused rewrite containing only the PROF-14603 race-fix changes (libraryPatcher_linux.cpp, perfEvents_linux.cpp, thread.h UNIT_TEST guard, thread_teardown_safety_ut.cpp, GtestTaskBuilder.kt). The chaos tests, CI pipeline changes, and objectSampler tweaks that were bundled here are dropped — they were already merged via separate PRs.

PR #528, opened 2026-05-13
fix(profiler): TripleBufferedDictionary — atomic rotate, counter tracking, stop() cleanup

jbachorik:
**What does this PR do?**:
Replaces the non-thread-safe `DoubleBufferedDictionary` with `TripleBufferedDictionary` — a three-buffer cycling design using `TripleBufferRotator<Dictionary>` that provides signal-safe concurrent access. This PR also fixes several issues found during pre-PR review:

- **Atomic rotate**: Changed `TripleBufferRotator::rotate()` from a non-atomic load-then-store to a CAS loop, preventing concurrent callers (e.g. `stop()` racing with dump) from silently aliasing the same buffer index.
- **Counter tracking accuracy**: `clearStandby()` now recalibrates the shared `DICTIONARY_KEYS` counter to `active()->size()` after clearing the scratch buffer, preventing the counter from zeroing out active buffer entry counts.
- **Stop cleanup**: `profiler.cpp::stop()` now calls `clearStandby()` on all three maps (`_class_map`, `_string_label_map`, `_context_value_map`) after `_jfr.stop()`, releasing scratch buffers held until the next `clearAll()`.
- **RefCountGuard documentation**: Documented the pointer-first activation window and its practical impossibility for signal handlers (microseconds vs. TWO full dump cycles / seconds).
- **Layout correctness**: `RefCountSlot` padding now uses `alignas(alignof(void*))` instead of a hard-coded `_padding1[4]`, which is correct on both 32-bit and 64-bit platforms.
- **Naming consistency**: Renamed `standby_table` → `standby_ptr` in `callTraceStorage.cpp` constructor.

**Motivation**:
The `DoubleBufferedDictionary` used in the class/endpoint/context maps was not signal-safe: `lookup()` could race with `rotate()`, and the counter tracking was zeroed on every standby clear, making debug counters unreliable.

**Additional Notes**:
- `TripleBufferedDictionary::lookup()` and `bounded_lookup()` acquire a per-thread `RefCountGuard` on the active buffer pointer before touching it. This makes the entire read path safe for signal handlers.
- `rotatePersistent()` pre-populates the new active buffer from the old active before rotating, preventing any gap in class-map visibility.
- All three Dictionary buffers share the same counter id so insertions are tracked regardless of which buffer is active.

**How to test the change?**:
- `DictionaryRotationTest.dumpCycleSeparatesPreAndPostDumpEntries` — verifies that rotate+clearStandby correctly separates pre-dump and post-dump entries, and that the live counter is recalibrated after clearStandby.
- `dictionary_ut.cpp` — unit tests for counter-id ownership and rotation behavior.
- CI matrix (glibc/musl, amd64/aarch64, JDK 8–25).

**For Datadog employees**:
- [ ] JIRA: [JIRA-XXXX]

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


jbachorik:
Closing: opened by mistake — work is tracked in PR #524

PR #535, opened 2026-05-18
fix(objectSampler): drop Deallocate on GetClassSignature error path

jbachorik:
**What does this PR do?**:

Removes the `jvmti->Deallocate(class_name)` call from the error branch of `ObjectSampler::recordAllocation` in `ddprof-lib/src/main/cpp/objectSampler.cpp`, and adds a TEST_F-based gtest fixture exercising the JVMTI mock for five error/success scenarios.

**Motivation**:

The error branch called `Deallocate` whenever `class_name` happened to be non-NULL after a failing `GetClassSignature` call. The JVMTI spec does not guarantee output buffers are populated on a non-`JVMTI_ERROR_NONE` return, so the pointer value is unspecified and passing it to `Deallocate` is unsafe — it crashed with SIGSEGV in `jvmti_Deallocate` reached via `JvmtiExport::post_sampled_object_alloc` (PROF-14626 / `dd-crashsig-1b6db147ca4415c7`, tracer 1.62.0*, Linux, JDK 25).

The success path Deallocate (still present at line 91) is unchanged: `class_name` is freed exactly once when `GetClassSignature` returns `JVMTI_ERROR_NONE`. The success branch's lookup-and-record flow is unaffected.

**Additional Notes**:

The test file uses a `TEST_F` fixture (`ObjectSamplerDeallocateTest`) so each test gets its own JVMTI function table and Deallocate counter, and `TearDown` restores the process-global `ObjectSampler::_active` flag — addressing review feedback from the muse chorus about shared static state, global counters, and singleton-flag leakage between tests.

**How to test the change?**:

Run the gtest suite:

```
./gradlew :ddprof-lib:gtestDebug_objectSampler_ut
```

Locally on macos-arm64: 15 tests pass (10 existing `normalizeClassSignature` cases + 5 new `ObjectSamplerDeallocateTest` cases — error+non-NULL sentinel, error+NULL name, success+NULL name, success+valid name (Deallocate called exactly once), and the !_active short-circuit).

**For Datadog employees**:
- [x] JIRA: [PROF-14626](https://datadoghq.atlassian.net/browse/PROF-14626)

[PROF-14626]: https://datadoghq.atlassian.net/browse/PROF-14626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/objectSampler.cpp
+++ b/ddprof-lib/src/main/cpp/objectSampler.cpp
@@ -77,9 +77,10 @@ void ObjectSampler::recordAllocation(jvmtiEnv *jvmti, JNIEnv *jni,
       class_name == NULL) {
     // Drop the sample: recording it under the default class id 0
     // would corrupt allocation attribution.
-    if (class_name != NULL) {
-      jvmti->Deallocate((unsigned char *)class_name);
-    }
+    // NOTE: Do NOT call Deallocate here. The JVMTI spec does not guarantee
+    // output buffers are populated on a non-JVMTI_ERROR_NONE return; the
+    // pointer value is unspecified, so passing it to Deallocate is unsafe
+    // in practice and observed to crash with SIGSEGV.
     return;
   }
   const char *name_slice = NULL;
--- a/ddprof-lib/src/main/cpp/objectSampler.h
+++ b/ddprof-lib/src/main/cpp/objectSampler.h
@@ -30,6 +30,7 @@ typedef int (*get_sampling_interval)();
 
 class ObjectSampler : public Engine {
   friend Recording;
+  friend class ObjectSamplerTestAccessor;
 
 private:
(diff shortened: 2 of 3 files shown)


jbachorik:
@codex review

rkennke (approved):
It seems weird that a JVM returns anything not-NULL in the failure case. But if that is what we observed then what else can we do? *shrugs*

jbachorik:
> It seems weird that a JVM returns anything not-NULL in the failure case. But if that is what we observed then what else can we do? _shrugs_

@rkennke indeed, it is very unexpected. But we did collect crashes at that exact point and other explanations, like broken JVMTI env are not really holding, since this happens inside a JVMTI callback 🤷

PR #536, opened 2026-05-20
fix(profiler): guard JVMTI sample path against null calltrace buffer

jbachorik:
**What does this PR do?**:

Defensive hardening of the JVMTI sampled-object-alloc path in `Profiler::recordJVMTISample`:

1. After the per-shard `tryLock` succeeds, load `_calltrace_buffer[lock_index]` into a local and drop the sample if it is `nullptr` instead of dereferencing it (`ddprof-lib/src/main/cpp/profiler.cpp:457-466`).
2. In `Profiler::start()`, serialise buffer reallocation against signal-handler readers by allocating the replacement *first*, then swapping the slot under the matching `_locks[i]`, and only freeing the old buffer once the lock is released (`ddprof-lib/src/main/cpp/profiler.cpp:1115-1133`). A `calloc` failure no longer leaves the slot pointing at freed memory.

**Motivation**:

Single observed SIGSEGV null-deref crash reported — `Profiler::recordJVMTISample` called from `JvmtiExport::post_sampled_object_alloc`, fault address `0x0`, JDK 25.0.3 / amd64 / Linux, 1 event / 4 days / 1 service.

The data is too thin to confirm a single root cause — these are best-effort defensive measures, not a verified fix. They close the cheapest plausible window without changing hot-path behaviour or adding any allocations.

Triage notes posted on the JIRA ticket.

### Triage summary

Walking the function for derefs reachable from the top frame, the closest match to a fault at `0x0` is `_calltrace_buffer[lock_index]->_asgct_frames` at `profiler.cpp:461-462` (now `465-466`): `_asgct_frames` is a union member at offset 0 of `CallTraceBuffer`, so a null `_calltrace_buffer[lock_index]` produces `frames == nullptr`, and the subsequent `GetStackTrace(..., jvmti_frames, ...)` write traps at `0x0`.

For the slot to be observed null while `ObjectSampler::_active = true`, the most plausible window is `Profiler::start()` reallocation at `profiler.cpp:1106-1114`: `free(_calltrace_buffer[i])` followed by `calloc`, with no per-shard lock held by `start` and no nulling between the two calls. A `tryLock`-based reader (the JVMTI sample path) does not synchronise against this writer, and a `calloc` failure path additionally leaves the slot pointing at freed memory.

`VM::jvmti()` and `Profiler::instance()` are statics that cannot become null at runtime; `FlightRecorder::recordEvent` already null-checks `_rec` under a shared lock. Inlining of `Recording::recordAllocation` writes into the same top frame remains a theoretical possibility but no obvious unguarded null was found there on inspection.

**Additional Notes**:

* Both changes are local and allocation-free on the hot path. The reader path adds one extra load and one branch (predictable: the slot is non-null in steady state). The writer adds a per-shard lock acquisition during `Profiler::start()` only.
* The reader uses a plain load: the writer publishes the pointer with the spin-lock's release barrier, and the reader's `tryLock` provides the matching acquire barrier.
* The change does not touch the other two consumers of `_calltrace_buffer[lock_index]` (`recordSample` at line ~559, `recordExternalSample` at line ~642). The writer-side serialisation under `_locks[i]` benefits them as well, but their explicit null-check is intentionally out of scope here.
* No new tests: the failure mode is a narrow start/stop/realloc race that cannot be exercised reliably from the existing gtests without introducing test-only hooks. Existing `:ddprof-lib:gtestDebug` suite still passes locally.

**How to test the change?**:

- [x] `:ddprof-lib:compileRelease` succeeds locally — verified
- [x] `:ddprof-lib:gtestDebug` passes locally — verified (8/8 in `UtilsTest`)
- [x] CI green on Linux x86_64 / aarch64 and musl variants
- [x] No regression on existing `ObjectSampler` JVMTI tests

**For Datadog employees**:

- [x] JIRA: [PROF-14679](https://datadoghq.atlassian.net/browse/PROF-14679)

[PROF-14679]: https://datadoghq.atlassian.net/browse/PROF-14679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PROF-14679]: https://datadoghq.atlassian.net/browse/PROF-14679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/profiler.cpp
+++ b/ddprof-lib/src/main/cpp/profiler.cpp
@@ -455,11 +455,20 @@ u64 Profiler::recordJVMTISample(u64 counter, int tid, jthread thread, jint event
   }
   u64 call_trace_id = 0;
   if (!_omit_stacktraces) {
+    // Defensive: the buffer slot can be observed as nullptr in pathological
+    // start sequences (e.g. a calloc failure path in Profiler::start before
+    // engines are enabled). Drop the sample rather than dereferencing.
+    CallTraceBuffer *buf = _calltrace_buffer[lock_index];
+    if (buf == nullptr) {
+      atomicIncRelaxed(_failures[-ticks_skipped]);
+      _locks[lock_index].unlock();
+      return 0;
+    }
 #ifdef COUNTERS
     u64 startTime = TSC::ticks();
 #endif // COUNTERS
-    ASGCT_CallFrame *frames = _calltrace_buffer[lock_index]->_asgct_frames;
-    jvmtiFrameInfo *jvmti_frames = _calltrace_buffer[lock_index]->_jvmti_frames;
+    ASGCT_CallFrame *frames = buf->_asgct_frames;
+    jvmtiFrameInfo *jvmti_frames = buf->_jvmti_frames;
 
     int num_frames = 0;
 
@@ -1104,13 +1113,24 @@ Error Profiler::start(Arguments &args, bool reset) {
     size_t nelem = _max_stack_depth + RESERVED_FRAMES;
 
     for (int i = 0; i < CONCURRENCY_LEVEL; i++) {
-      free(_calltrace_buffer[i]);
-      _calltrace_buffer[i] = (CallTraceBuffer*)calloc(nelem, sizeof(CallTraceBuffer));
-      if (_calltrace_buffer[i] == NULL) {
+      // Allocate the replacement before touching the slot so a calloc failure
+      // does not leave the slot pointing at freed memory.
+      CallTraceBuffer *fresh =
+          (CallTraceBuffer*)calloc(nelem, sizeof(CallTraceBuffer));
+      if (fresh == NULL) {
         _max_stack_depth = 0;
         return Error("Not enough memory to allocate stack trace buffers (try "
                      "smaller jstackdepth)");
       }
+      // Swap under the per-shard lock so a reader holding the same lock via
+      // tryLock cannot observe a freed pointer mid-replacement. The previous
+      // buffer is freed only after the lock is released, by which point no
+      // reader can be using it.
+      _locks[i].lock();
+      CallTraceBuffer *prev = _calltrace_buffer[i];
+      _calltrace_buffer[i] = fresh;
+      _locks[i].unlock();
+      free(prev);
     }
   }
 
--- a/ddprof-lib/src/main/cpp/profiler.cpp
+++ b/ddprof-lib/src/main/cpp/profiler.cpp
@@ -1122,10 +1122,10 @@ Error Profiler::start(Arguments &args, bool reset) {
         return Error("Not enough memory to allocate stack trace buffers (try "
                      "smaller jstackdepth)");
       }
-      // Swap under the per-shard lock so a reader holding the same lock via
-      // tryLock cannot observe a freed pointer mid-replacement. The previous
-      // buffer is freed only after the lock is released, by which point no
-      // reader can be using it.
+      // Swap under the per-shard lock: tryLock-based readers (e.g. recordJVMTISample)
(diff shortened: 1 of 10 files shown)


jbachorik:
@codex review

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -1104,13 +1113,24 @@ Error Profiler::start(Arguments &args, bool reset) {
...
+          (CallTraceBuffer*)calloc(nelem, sizeof(CallTraceBuffer));
+      if (fresh == NULL) {
         _max_stack_depth = 0;
         return Error("Not enough memory to allocate stack trace buffers (try "
                      "smaller jstackdepth)");
       }
+      // Swap under the per-shard lock so a reader holding the same lock via
+      // tryLock cannot observe a freed pointer mid-replacement. The previous
+      // buffer is freed only after the lock is released, by which point no
+      // reader can be using it.
```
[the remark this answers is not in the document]
jbachorik: Addressed — replaced the misleading comment with a two-path explanation: tryLock-based readers (e.g. `recordJVMTISample`) are protected by the shard lock, and lock-free readers (e.g. `recordExternalSample`) are safe because the swap completes before `enableEngines()` is called — no sampling engine is active yet.

jbachorik:
@Copilot review

kaahos (approved):
looks good to me!

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -1166,13 +1175,24 @@ Error Profiler::start(Arguments &args, bool reset) {
...
+      CallTraceBuffer *fresh =
+          (CallTraceBuffer*)calloc(nelem, sizeof(CallTraceBuffer));
+      if (fresh == NULL) {
         _max_stack_depth = 0;
         return Error("Not enough memory to allocate stack trace buffers (try "
                      "smaller jstackdepth)");
       }
+      // Swap under the per-shard lock: tryLock-based readers (e.g. recordJVMTISample)
+      // cannot observe a freed pointer mid-replacement. Lock-free readers
+      // (e.g. recordExternalSample) are safe because this swap completes before
```
jbachorik: Wait, should not recordExternalSample also do try-lock????

kaahos: I think you are right, I missed it. `recordExternalSample` suffers from the same concurrency issue. try-lock seems necessary.

jbachorik: Addressed — restructured `recordExternalSample` to acquire the tryLock before reading `_calltrace_buffer`, matching the pattern from `recordJVMTISample`. Also added the same null-guard for the buffer slot and updated the misleading 'lock-free readers' comment in `Profiler::start()`.

ddprof-lib/src/main/cpp/profiler.cpp
kaahos, line 571: after a more careful check, here is the last place where there is an unguarded access to `_calltrace_buffer`. It is not directly related to the PR purpose but it looks like a positive defensive check to prevent future crashes, even though it looks less critical here compared to the previous places.

jbachorik: This one should be ok - we are locking at line [L547](https://github.com/DataDog/java-profiler/pull/536/changes/BASE..908435dd42d7ea7ba31da14af99d3378e4297b21#diff-2da44c131e16a432746c7c1e09102d1d3b7b6717a36c087f746d2fa412fad9eaR547)

PR #537, opened 2026-05-20
fix(profiler): probe JVMTI strings before strncmp in fillJavaMethodInfo

jbachorik:
**What does this PR do?**:

Adds a page-range readability probe (`SafeAccess::isReadableRange(ptr, 256)`) for the `class_name`, `method_name`, and `method_sig` pointers returned by JVMTI inside `Lookup::fillJavaMethodInfo`. If any probe fails, the method routes to the existing `JMETHODID_SKIPPED` placeholder instead of crashing in the subsequent `strncmp` / `has_prefix` calls.

**Motivation**:

We have reports of SIGSEGV crashes whose top frame is `strncmp` directly under `Lookup::fillJavaMethodInfo` on JDK 25 + tracer 1.62.0 (4 events / 2 orgs in 3 days). The JVMTI 1.2 spec guarantees these pointers are non-null on success, and JDK 25 source review (`jvmtiEnv.cpp:3214` `GetMethodName`, `:2644` `GetClassSignature`) confirms no path produces `JVMTI_ERROR_NONE` with a NULL output. The crash is therefore either a HotSpot bug not visible from code review or external memory corruption — in either case a page-range probe converts the crash into observable degraded behavior via the existing `JMETHODID_SKIPPED` counter.

The structural fix is the in-flight jmethodID-removal plan (`doc/plans/2026-04-01-remove-jmethodid-*.md`); this is an interim guard.

**Additional Notes**:

- Probe range: 256 bytes per pointer — covers the longest prefix compared in the function (`"Ljdk/internal/reflect/GeneratedConstructorAccessor"`, ~50 bytes) plus headroom for `strlen(class_name)` at line 282. Page-level check, so typically one page per pointer; two only when the string straddles a page boundary.
- Catches: pointer to an unmapped page anywhere in the 256-byte window — the observed crash signature.
- Does NOT catch: freed-but-still-mapped buffers in the libc arena (would not produce the observed SIGSEGV), or a string whose body extends past 256 bytes into an unmapped page (rare; typical class signatures are well under 256).
- Hot path cost: three `isReadableRange` calls per `fillJavaMethodInfo` — each is one or two signal-trapped page loads.
- Cleanup at `flightRecorder.cpp:325-333` is unchanged — the existing `if (ptr) Deallocate(ptr)` guards run unconditionally after the success/fallback branch, so probe rejections do not leak JVMTI strings.
- TOCTOU: a concurrent `munmap` between probe and use can still fault. Inherent to probe-before-use; the existing SIGSEGV handler is the second line of defence. Acknowledged in the inline comment.

**How to test the change?**:

Without a reproducer for the JDK 25 crash, the guard is only exercised on the production crash path that we cannot trigger synthetically — `JMETHODID_SKIPPED` increments in JFR counter telemetry on affected hosts would confirm it fires. No unit test added because the production JVMTI path has no test-only hook to inject an unmapped pointer, and adding one would carry more risk than the guard itself. `buildDebug` passes locally.

**For Datadog employees**:

- [x] JIRA: [PROF-14623](https://datadoghq.atlassian.net/browse/PROF-14623)

[PROF-14623]: https://datadoghq.atlassian.net/browse/PROF-14623?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -190,7 +190,7 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
       // so a single bad pointer does not leak its siblings. Best-effort only:
       // a concurrent munmap between probe and use can still fault; the SIGSEGV
       // handler is the second line of defence.
-      auto probe = [&](const char*& ptr) -> bool {
+      auto probe = [&](char*& ptr) -> bool {
         if (ptr == nullptr || !SafeAccess::isReadableRange(ptr, probe_len)) {
           ptr = nullptr;
           return false;
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -164,6 +164,7 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
   if ((phase & (JVMTI_PHASE_START | JVMTI_PHASE_LIVE)) != 0) {
     bool entry = false;
     bool readable = false;
+    const size_t probe_len = 256;
     if (VMMethod::check_jmethodID(method) &&
         jvmti->GetMethodDeclaringClass(method, &method_class) == 0 &&
         // GetMethodDeclaringClass may return a jclass wrapping a stale/garbage oop when the class was
@@ -189,20 +190,14 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
       // so a single bad pointer does not leak its siblings. Best-effort only:
       // a concurrent munmap between probe and use can still fault; the SIGSEGV
       // handler is the second line of defence.
-      const uintptr_t probe_len = 256;
-      bool class_name_ok = class_name != nullptr &&
-                           reinterpret_cast<uintptr_t>(class_name) <= UINTPTR_MAX - (probe_len - 1) &&
-                           SafeAccess::isReadableRange(class_name, probe_len);
-      bool method_name_ok = method_name != nullptr &&
-                            reinterpret_cast<uintptr_t>(method_name) <= UINTPTR_MAX - (probe_len - 1) &&
-                            SafeAccess::isReadableRange(method_name, probe_len);
-      bool method_sig_ok = method_sig != nullptr &&
-                           reinterpret_cast<uintptr_t>(method_sig) <= UINTPTR_MAX - (probe_len - 1) &&
-                           SafeAccess::isReadableRange(method_sig, probe_len);
-      if (!class_name_ok) class_name = nullptr;
-      if (!method_name_ok) method_name = nullptr;
-      if (!method_sig_ok) method_sig = nullptr;
-      readable = class_name_ok && method_name_ok && method_sig_ok;
+      auto probe = [&](const char*& ptr) -> bool {
+        if (ptr == nullptr || !SafeAccess::isReadableRange(ptr, probe_len)) {
+          ptr = nullptr;
+          return false;
+        }
+        return true;
+      };
+      readable = probe(class_name) & probe(method_name) & probe(method_sig);
     }
     if (readable) {
 
@@ -301,13 +296,13 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
         } else {
           // don't recognise the suffix, so don't normalise
           class_name_id =
-              _classes->lookup(class_name + 1, strlen(class_name) - 2);
+              _classes->lookup(class_name + 1, strnlen(class_name, 65536) - 2);
         }
         method_name_id = _symbols.lookup(method_name);
         method_sig_id = _symbols.lookup(method_sig);
       } else {
         class_name_id =
-            _classes->lookup(class_name + 1, strlen(class_name) - 2);
+            _classes->lookup(class_name + 1, strnlen(class_name, 65536) - 2);
         method_name_id = _symbols.lookup(method_name);
         method_sig_id = _symbols.lookup(method_sig);
       }
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -20,6 +20,7 @@
 #include "os.h"
 #include "profiler.h"
 #include "rustDemangler.h"
+#include "safeAccess.h"
 #include "spinLock.h"
 #include "unwindStats.h"
 #include "symbols.h"
@@ -176,7 +177,13 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
         // The check below mitigates these crashes on J9.
         (!VM::isOpenJ9() || method_class != reinterpret_cast<jclass>(-1)) &&
         jvmti->GetClassSignature(method_class, &class_name, NULL) == 0 &&
-        jvmti->GetMethodName(method, &method_name, &method_sig, NULL) == 0) {
+        jvmti->GetMethodName(method, &method_name, &method_sig, NULL) == 0 &&
+        // PROF-14623: the JVMTI strings should be non-null and mapped, but crash
+        // telemetry shows strncmp dereferences on unmapped memory anyway. Page-level
+        // probe of each pointer; on failure fall through to JMETHODID_SKIPPED.
+        class_name != nullptr && SafeAccess::isReadable(class_name) &&
+        method_name != nullptr && SafeAccess::isReadable(method_name) &&
+        method_sig != nullptr && SafeAccess::isReadable(method_sig)) {
 
       if (first_time) {
         jvmtiError line_table_error = jvmti->GetLineNumberTable(method, &line_number_table_size,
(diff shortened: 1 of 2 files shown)


jbachorik:
@codex review

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -176,7 +177,13 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
...
         (!VM::isOpenJ9() || method_class != reinterpret_cast<jclass>(-1)) &&
         jvmti->GetClassSignature(method_class, &class_name, NULL) == 0 &&
-        jvmti->GetMethodName(method, &method_name, &method_sig, NULL) == 0) {
+        jvmti->GetMethodName(method, &method_name, &method_sig, NULL) == 0 &&
+        // PROF-14623: the JVMTI strings should be non-null and mapped, but crash
+        // telemetry shows strncmp dereferences on unmapped memory anyway. Page-level
+        // probe of each pointer; on failure fall through to JMETHODID_SKIPPED.
+        class_name != nullptr && SafeAccess::isReadable(class_name) &&
+        method_name != nullptr && SafeAccess::isReadable(method_name) &&
+        method_sig != nullptr && SafeAccess::isReadable(method_sig)) {
```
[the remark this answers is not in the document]
jbachorik: Addressed in edf9dcb. You're correct — the original ticket already documented a `jvmti_Deallocate` variant of the crash with identical frames #1–7, which I missed in my first pass. The refactored guard now probes each JVMTI string independently and `NULL`s any pointer that fails, so the unconditional `Deallocate` block at the end of the function skips it. Accepts a small leak on the corruption path — better than a second-order SIGSEGV inside `os::free`.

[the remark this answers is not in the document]
jbachorik: Addressed in 789b9ff. Swapped `isReadable(ptr)` for `isReadableRange(ptr, 256)` on each of the three pointers. 256 covers the longest prefix compared (`Ljdk/internal/reflect/GeneratedConstructorAccessor`, ~50 bytes) plus headroom for `strlen(class_name)` at the lookup site. Page-level check, so still one probe per pointer in the typical case, two only when the string straddles a page boundary.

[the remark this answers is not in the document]
jbachorik: Addressed in edf9dcb. The probes are now independent and NULL the offending pointer on failure, so the `Deallocate` block skips unreadable pointers. Independent (not short-circuited) so a failure on one string doesn't leak the others. Accepts the small leak on the corruption path as the right tradeoff vs. a second-order SIGSEGV inside `free`.

jbachorik:
@codex review

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -177,6 +179,25 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
...
+      // compared below (~50 bytes) plus headroom for strlen, and NULL any that
+      // fails so the unconditional Deallocate block at end of this function
+      // skips it (os::free faults on an unmapped pointer just like strncmp).
+      // Accept a small leak on the corruption path. Probes run independently
+      // so a single bad pointer does not leak its siblings. Best-effort only:
+      // a concurrent munmap between probe and use can still fault; the SIGSEGV
+      // handler is the second line of defence.
+      bool class_name_ok = class_name != nullptr && SafeAccess::isReadableRange(class_name, 256);
+      bool method_name_ok = method_name != nullptr && SafeAccess::isReadableRange(method_name, 256);
+      bool method_sig_ok = method_sig != nullptr && SafeAccess::isReadableRange(method_sig, 256);
```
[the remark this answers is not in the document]
jbachorik: Addressed — added an overflow guard in isReadableRange before computing end_page, and restructured the loop to use page != end_page with a separate return isReadable(end_page) — the loop increment is now only evaluated when page < end_page, eliminating wrap risk on the last valid page.

jbachorik:
@codex review

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -177,6 +179,32 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
...
+      // compared below (~50 bytes) plus headroom for strlen, and NULL any that
+      // fails so the unconditional Deallocate block at end of this function
+      // skips it (os::free faults on an unmapped pointer just like strncmp).
+      // Accept a small leak on the corruption path. Probes run independently
+      // so a single bad pointer does not leak its siblings. Best-effort only:
+      // a concurrent munmap between probe and use can still fault; the SIGSEGV
+      // handler is the second line of defence.
+      const uintptr_t probe_len = 256;
+      bool class_name_ok = class_name != nullptr &&
+                           reinterpret_cast<uintptr_t>(class_name) <= UINTPTR_MAX - (probe_len - 1) &&
```
rkennke: wowzers, what is that check for? A valid pointer would never even get close to UINTPTR_MAX (no platform uses 64bit pointers, afaik). *scratches head*

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -177,6 +179,32 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
...
+                           SafeAccess::isReadableRange(class_name, probe_len);
+      bool method_name_ok = method_name != nullptr &&
+                            reinterpret_cast<uintptr_t>(method_name) <= UINTPTR_MAX - (probe_len - 1) &&
+                            SafeAccess::isReadableRange(method_name, probe_len);
+      bool method_sig_ok = method_sig != nullptr &&
+                           reinterpret_cast<uintptr_t>(method_sig) <= UINTPTR_MAX - (probe_len - 1) &&
+                           SafeAccess::isReadableRange(method_sig, probe_len);
+      if (!class_name_ok) class_name = nullptr;
```
rkennke: err hmm, maybe worth to extract the whole thing in a small helper function and call that once on each of the strings? Might even come in handy for other similar cases elsewhere...

ddprof-lib/src/main/cpp/flightRecorder.cpp
rkennke: Especially since you already seem to handle that case in isReadableRange().

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -177,6 +179,32 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
...
+      // telemetry shows both `strncmp` and `jvmti_Deallocate` faulting on them.
+      // Probe each pointer over a range covering the longest prefix
+      // compared below (~50 bytes) plus headroom for strlen, and NULL any that
+      // fails so the unconditional Deallocate block at end of this function
+      // skips it (os::free faults on an unmapped pointer just like strncmp).
+      // Accept a small leak on the corruption path. Probes run independently
+      // so a single bad pointer does not leak its siblings. Best-effort only:
+      // a concurrent munmap between probe and use can still fault; the SIGSEGV
+      // handler is the second line of defence.
+      const uintptr_t probe_len = 256;
```
rkennke: make this size_t to match isReadableRange argument? It is also semantically better.

rkennke (dismissed):
Good defensive change, I have some small concerns.
One thing that I don't feel good about is the unbounded strlen() calls, especially if you assume >256 is rare. class signatures are specified as u2 which allows 65535 bytes, and I would think that packages/.../ClassName can quickly exceed 256 bytes in enterprisey applications, especially when code generators are involved. I've seen horrors... So better use strnlen with a meaningful limit?

ddprof-lib/src/main/cpp/flightRecorder.cpp
jbachorik: Addressed — changed to `const size_t probe_len = 256;` (matching `isReadableRange`'s `size_t` parameter) and moved to the enclosing scope so it is visible at both the probe site and the `strnlen` call sites below.

ddprof-lib/src/main/cpp/flightRecorder.cpp
jbachorik: The `reinterpret_cast<uintptr_t>(X) <= UINTPTR_MAX - (probe_len - 1)` guard was added to protect against pointer arithmetic overflow before `isReadableRange` computed `start + size - 1` internally. Since `isReadableRange` now performs this check itself (the guard was added there in this same PR), the outer redundant guards have been removed in the lambda refactor.

ddprof-lib/src/main/cpp/flightRecorder.cpp
jbachorik: Addressed — extracted into a `[&](const char*\& ptr) -> bool` lambda that nulls the pointer if `isReadableRange` fails. All three probes are combined with bitwise `&` so each runs independently (no short-circuit), preserving the 'single bad pointer does not leak its siblings' invariant from the comment.

jbachorik:
Addressed — replaced both `strlen(class_name)` calls with `strnlen(class_name, 65536)`, using the JVM u2 maximum (65535 bytes) as the bound rather than the smaller 256-byte probe window. This way, valid long class names aren't silently truncated in the profile data.

kaahos (approved):
looks good to me. Every highlighted concern has been addressed.

PR #550, opened 2026-05-27
Native malloc fixes and cleanup

jbachorik:
**What does this PR do?**:
Fixes reentrancy double-counting in the native malloc tracer, cleans up hook dispatch for musl, and removes dead code.

- Add `CriticalSection` guard in `maybeRecord` to prevent profiler-internal allocations (e.g. JFR buffer growth) from re-entering the hook and being double-counted
- Replace `calloc_hook_dummy` / `posix_memalign_hook_dummy` pass-through hooks with a simple skip: on musl where `calloc` delegates to `malloc` and `posix_memalign` delegates to `aligned_alloc` internally, we leave those GOT entries unpatched and let the inner hook record the allocation
- Remove `_orig_free` / `SAVE_IMPORT(free)` — free is not tracked for liveness; the resolved pointer was only used for a probe cleanup where plain `free()` is correct
- Remove the `_calloc_hook_fn` / `_posix_memalign_hook_fn` pre-computed pointer members and the `NO_OPTIMIZE` macro that existed solely for the dummy hooks
- Gate `MallocTracer::installHooks()` in `Libraries::refresh()` on `MallocTracer::running()` to avoid unnecessary GOT patching work between profiling sessions

**Motivation**:
Profiler-internal allocations triggered inside `recordMalloc` (e.g. sample buffer allocation) re-enter the patched GOT and were being silently double-counted. The dummy hook approach for musl added function-call overhead on every `calloc`/`posix_memalign` without benefit.

**Additional Notes**:
`detectNestedMalloc()` is kept — it still drives the decision of whether to skip patching `calloc`/`posix_memalign` on musl.

**How to test the change?**:
Covered by existing integration tests:
- `NativememProfilerTest` — verifies `profiler.Malloc` JFR events have positive size, non-zero address, and Java stack trace
- `NativememSampledProfilerTest` — verifies weight and sample count from repeated allocations

**For Datadog employees**:
- [x] JIRA: [PROF-14807](https://datadoghq.atlassian.net/browse/PROF-14807)

[PROF-14807]: https://datadoghq.atlassian.net/browse/PROF-14807?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/libraries.cpp
+++ b/ddprof-lib/src/main/cpp/libraries.cpp
@@ -60,7 +60,9 @@ void Libraries::refresh() {
   _dirty.store(false, std::memory_order_release);
   updateSymbols(false);
   LibraryPatcher::patch_sigaction();
-  MallocTracer::installHooks();
+  if (MallocTracer::running()) {
+    MallocTracer::installHooks();
+  }
   if (_remote_symbolication) {
     updateBuildIds();
   }
--- a/ddprof-lib/src/main/cpp/mallocTracer.cpp
+++ b/ddprof-lib/src/main/cpp/mallocTracer.cpp
@@ -20,12 +20,6 @@
 #include "tsc.h"
 #include "vmEntry.h"
 
-#ifdef __clang__
-#  define NO_OPTIMIZE __attribute__((optnone))
-#else
-#  define NO_OPTIMIZE __attribute__((optimize("-fno-omit-frame-pointer,-fno-optimize-sibling-calls")))
-#endif
-
 #define SAVE_IMPORT(FUNC) \
     do { \
         void** _entry = lib->findImport(im_##FUNC); \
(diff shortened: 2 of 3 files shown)


rkennke (approved):
Looks good. I assume the added overhead of the CS in maybeRecord() is acceptable? E.g. malloc-heavy workloads?

jbachorik:
> I assume the added overhead of the CS in maybeRecord() is acceptable? E.g. malloc-heavy workloads?

I would lean towards 'yes' - but I don't have any benchmarks for that :(
For now, the feature is a kind of experimental and opt-in so we can easily reevaluate later, before making it a publicly supported feature, based on proper benchmarks.

PR #552, opened 2026-05-28
fix(profiler): close onThreadEnd teardown race

jbachorik:
**What does this PR do?**:
Closes a signal-safety gap in `Profiler::onThreadEnd()` where a `SIGPROF` or `SIGVTALRM`
delivered between `ProfiledThread::release()` and `_wall_engine->unregisterThread(tid)` could
call `currentSignalSafe()` and dereference a freed `ProfiledThread` (SIGSEGV in
`JavaThread::thread_main_inner` on JDK 25.0.3).

Also upgrades T-05 and T-09 in `thread_teardown_safety_ut.cpp` from crash-only guards
(signals discarded via `SIG_IGN`) to active guards: a real handler now calls
`currentSignalSafe()` and asserts the handler was actually invoked.

**Motivation**:
SIGSEGV crash reproduced on JDK 25.0.3. The teardown sequence in
`onThreadEnd()` had a window between `ProfiledThread::release()` freeing TLS and
`unregisterThread()` removing the thread from the wall-clock filter; a profiling signal
in that window could sample the freed thread.

**Additional Notes**:
- `updateThreadName()` is kept *before* the `SignalBlocker` scope because JVMTI
  `GetThreadInfo` allocates memory. Under ASAN, JVMTI allocation acquires the same
  allocator lock a signal handler would try to acquire, causing deadlock. `ProfiledThread`
  is still live at that point, so concurrent signals are safe.
- Pattern mirrors the prior art in `libraryPatcher_linux.cpp` (`unregister_and_release`,
  PROF-14603).

**How to test the change?**:
- `thread_teardown_safety_ut` T-05 and T-09 now exercise the handler path and will fail
  (rather than silently pass) if signals are discarded or the handler is never invoked.
- Existing T-01 through T-10 cover the broader teardown contract.
- On debug builds: `./gradlew :ddprof-lib:testDebug` (Linux).

**For Datadog employees**:

- [x] JIRA: [PROF-14674](https://datadoghq.atlassian.net/browse/PROF-14674)

[PROF-14674]: https://datadoghq.atlassian.net/browse/PROF-14674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/profiler.cpp
+++ b/ddprof-lib/src/main/cpp/profiler.cpp
@@ -103,19 +103,28 @@ void Profiler::onThreadEnd(jvmtiEnv *jvmti, JNIEnv *jni, jthread thread) {
       _thread_filter.unregisterThread(slot_id);
       current->setFilterSlotId(-1);
     }
-    
-    ProfiledThread::release();
-  } else {
-    // ProfiledThread already cleaned up - try to get tid from JVMTI as fallback
-    tid = JVMThread::nativeThreadId(jni, thread);
-    if (tid < 0) {
-      // No ProfiledThread AND can't get tid from JVMTI - nothing we can do
-      return;
+
+    updateThreadName(jvmti, jni, thread, false);
+    // Block profiling signals around engine unregistration + TLS release to
+    // close the window where a wall-clock/CPU signal could sample a
+    // partially-torn-down thread (PROF-14674).
+    {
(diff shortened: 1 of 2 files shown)


zhengyu123 (approved):
Looks fine.

PR #558, opened 2026-05-29
feat(chaos): production-pattern antagonists + Java-25 thread-churn crash simulation

jbachorik:
**What does this PR do?**:

Extends the chaos reliability harness in two areas:

**1. Six new production-pattern antagonists** targeting failure patterns observed when profiling gRPC/Netty/Kafka-style workloads:
- `BoundedThreadPoolAntagonist` (`bounded-pool`): 4 scheduled thread pools cross-submitting tasks; one pool torn down and recreated every 5 s — targets signal-vs-thread-end races during pool shutdown.
- `ContextHopAntagonist` (`context-hop`): 8 self-renewing `CompletableFuture` chains hopping A→B→C→A across 3 pools, each stage setting/clearing a `ThreadLocal` — targets RefCountGuard slot contention under cross-pool handoff.
- `ConsumerGroupAntagonist` (`consumer-group`): 16 consumer threads with 4 replaced in a burst every 3 s, simulating Kafka rebalance — targets ProfiledThread recycling and signal-vs-thread-end race window.
- `HiddenClassChurnAntagonist` (`hidden-class-churn`): generates hidden classes via `MethodHandles.Lookup.defineHiddenClass` (Java 15+, reflective no-op on older JDKs) — targets StringDictionary concurrent eviction racing hidden-class GC.
- `DirectMemoryAntagonist` (`direct-memory`): ring-buffer + burst `ByteBuffer.allocateDirect()` allocation — targets liveness-table overflow under high off-heap churn.
- `WeakRefWaveAntagonist` (`weakref-wave`): alternating fill/drop phases with 10k `WeakReference<byte[]>` and a concurrent reader — targets jweak ref leak and liveness-table `clearAll()` race.

**2. Crash simulation for the confirmed Java-25 thread-churn × recording-dump UAF** (PROF quality escape from logs-backend internal services):
- `DumpStormAntagonist` (`dump-storm`): spawns 96 short-lived, uniquely-named threads each with a distinct stack shape, maximising churn in the thread-name table and call-trace storage concurrent with recording dumps — targets `Recording::switchChunk/writeCpool`, `updateJavaThreadNames → ThreadInfo::set`, `Dictionary::clear`.
- `stress_threadLifecycle_ut.cpp` (Layer-1 gtest): drives `ProfiledThread`, `ThreadFilter`, `CallTraceStorage`, and `Dictionary` concurrently without a JVM so ASan/TSan can locate the UAF at its origin. Runs under the existing `buildGtestAsan`/`buildGtestTsan` targets.
- Chaos CI matrix extended with `CHAOS_JDK` dimension (`21.0.3-tem` + `25.0.3-tem`) — all crashes confirmed on Java 25 in production.

**CI fixes** (introduced with the JDK matrix):
- `REASON` dotenv key now includes `CHAOS_JDK` slug to prevent key collisions when two JDK legs of the same job both fail.
- `tee` replaced with direct redirect so `chaos_check.sh` exit status is not swallowed.
- JDK version verified after `sdk use`; emits `FAIL:wrong JDK` if install silently fell back.

**Motivation**:

Running the profiler against logs-backend production services (gRPC servers, Kafka consumers, Netty pipelines) reveals two classes of crashes our existing tests miss:

1. Framework-driven thread lifecycle patterns not present in synthetic benchmarks (covered by the 6 production-pattern antagonists).
2. A confirmed profiler UAF on Java 25 under high thread churn × concurrent recording dump — ~200 crashes/7 days on Datadog's own staging services. Disabling the profiler removes the crashes; the `crash_datadog` attribution tag undercounted the real impact by ~8× because most crashes surface in JVM/JFR code rather than profiler frames. `DumpStormAntagonist` + the JDK-25 chaos cell reproduce this scenario end-to-end.

**Additional Notes**:

All antagonists compile under `--release 8` and use only JDK APIs. `HiddenClassChurnAntagonist` uses reflection for the Java 15+ `defineHiddenClass` API and prints a skip message on older JDKs.

Investigation and design: `doc/plans/2026-05-29-logs-backend-crash-simulation-design.md`. A follow-up plan for JFR-seeded antagonist calibration is at `doc/plans/2026-05-29-jfr-seeded-antagonist-calibration.md`.

**How to test the change?**:

The chaos cell is the primary test. Antagonists run under all three allocators × both archs × both configs × both JDKs in the scheduled reliability pipeline.

Local smoke test (no profiler attached — verifies no immediate crash or hang):
```bash
./gradlew :ddprof-stresstest:chaosJar
java -jar ddprof-stresstest/build/libs/chaos.jar \
  --duration 30s \
  --antagonists bounded-pool,context-hop,consumer-group,hidden-class-churn,direct-memory,weakref-wave,dump-storm
```
Expected: exits 0, last line `[chaos] completed cleanly`.

ASan gtest (Linux, requires Docker):
```bash
./utils/run-docker-tests.sh --config=asan --gtest --jdk=25 --mount
```
Expected: `stress_threadLifecycle_ut` compiles and all three cases pass under ASan.

**For Datadog employees**:
- [ ] JIRA: [JIRA-XXXX]


ddprof-stresstest/src/chaos/java/com/datadoghq/profiler/chaos/HiddenClassChurnAntagonist.java
```diff
@@ -0,0 +1,130 @@
...
+            } catch (Throwable t) {
+                // transient; JVM crash is the signal we watch for
+            }
+        }
+    }
+
+    private static byte[] generateClass(long uid) {
+        // Unique internal name per class so each one gets a distinct entry
+        // in the profiler's class map / StringDictionary.
+        String internalName = "chaos/hidden/Gen" + uid;
```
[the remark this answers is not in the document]
jbachorik: Fixed in 2c24c6075.

docs/sphinx/specs/2026-05-29-generate-hidden-classes-in-the-lookup.md
```diff
@@ -0,0 +1,20 @@
+# generate-hidden-classes-in-the-lookup-packag
```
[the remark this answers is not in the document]
jbachorik: Fixed in 05c8fcb4c.

ddprof-stresstest/src/chaos/java/com/datadoghq/profiler/chaos/BoundedThreadPoolAntagonist.java
```diff
@@ -0,0 +1,148 @@
...
+            recycler.join(timeout.toMillis());
+        } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+        }
+        for (ScheduledExecutorService pool : pools) {
+            if (pool != null) {
+                pool.shutdownNow();
+            }
+        }
+    }
```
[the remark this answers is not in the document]
jbachorik: Fixed in 05c8fcb4c.

ddprof-stresstest/src/chaos/java/com/datadoghq/profiler/chaos/ConsumerGroupAntagonist.java
```diff
@@ -0,0 +1,132 @@
...
+        }
+        for (Thread t : consumers) {
+            if (t != null) {
+                t.interrupt();
+                try { t.join(500L); } catch (InterruptedException e) {
+                    Thread.currentThread().interrupt();
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
jbachorik: Fixed in 05c8fcb4c.

.gitlab/reliability/.gitlab-ci.yml
```diff
@@ -78,16 +78,17 @@ reliability-aarch64:
...
+        CHAOS_JDK: ["21.0.3-tem", "25.0.3-tem"]
   script:
     - set +e
-    - echo "runtime=${RUNTIME}, config=${CONFIG}, allocator=${ALLOCATOR}, arch=${ARCH}"
-    - .gitlab/reliability/chaos_check.sh "$RUNTIME" "$CONFIG" "$ALLOCATOR" 2>err.log | tee out.log
+    - echo "runtime=${RUNTIME}, config=${CONFIG}, allocator=${ALLOCATOR}, arch=${ARCH}, jdk=${CHAOS_JDK}"
+    - CHAOS_JDK="${CHAOS_JDK}" .gitlab/reliability/chaos_check.sh "$RUNTIME" "$CONFIG" "$ALLOCATOR" 2>err.log 1>out.log
     - REASON=$(cat err.log | grep "FAIL:" | cut -f2 -d':') || true
-    - if [ -n "${REASON}" ]; then echo "REASON_${CONFIG}_${ALLOCATOR}_${ARCH}Xchaos=${REASON}" | tr '+' '_' >> build.env; exit 1; fi
+    - if [ -n "${REASON}" ]; then echo "REASON_${CONFIG}_${ALLOCATOR}_${ARCH}_${CHAOS_JDK//./_}Xchaos=${REASON}" | tr '+' '_' >> build.env; exit 1; fi
```
rkennke: My Claude review says that GitLab would reject the newly introduced '-' (from "25.0.3-tem"), and suggests to also replace '-' with '_'.

ddprof-lib/src/test/cpp/stress_threadLifecycle_ut.cpp
```diff
@@ -0,0 +1,206 @@
...
+// Thread-name / ThreadInfo coverage: this native reproducer does not call
+// ThreadInfo::set() or updateJavaThreadNames(). That path is covered by the
+// JVM-level DumpStormAntagonist antagonist (Layer 2). A clean ASan/TSan run
+// here is not conclusive for the thread-name path.
+static void churn_worker(ThreadFilter* filter, bool with_dump) {
+  while (!g_run.load(std::memory_order_acquire)) { }
+  for (int i = 0; i < kChurnIterations && g_run.load(std::memory_order_relaxed); i++) {
+    ProfiledThread::initCurrentThread();
+    ProfiledThread* self = ProfiledThread::current();
+    ASSERT_NE(nullptr, self);
```
rkennke: Fatal assertions (ASSERT_*) are only safe on the main thread in GoogleTest — in a spawned thread they don't reliably register the failure and aren't thread-safe. Notably the neighboring thread_teardown_safety_ut.cpp already uses EXPECT_* inside its thread bodies for exactly this reason. self should never be null here so it won't fire in practice, but for correctness prefer EXPECT_NE(...) followed by an explicit if (!self) return;, matching the established convention in the file's own reference neighbors.

.gitlab/reliability/.gitlab-ci.yml
```diff
@@ -78,16 +78,17 @@ reliability-aarch64:
     matrix:
       - CONFIG: ["profiler", "profiler+tracer"]
         ALLOCATOR: ["gmalloc", "jemalloc", "tcmalloc"]
+        CHAOS_JDK: ["21.0.3-tem", "25.0.3-tem"]
```
rkennke, line 81: Are we ok with doubling the load?

ddprof-stresstest/src/chaos/java/com/datadoghq/profiler/chaos/ConsumerGroupAntagonist.java
```diff
@@ -0,0 +1,135 @@
...
+ * signal-vs-thread-end race window widened by simultaneous stops;
+ * calltrace storage put() racing thread destruction.
+ */
+public final class ConsumerGroupAntagonist implements Antagonist {
+
+    private static final int GROUP_SIZE = 16;
+    private static final int BURST_SIZE = 4;
+    private static final long REBALANCE_INTERVAL_MS = 3_000L;
+
+    private final Thread[] consumers = new Thread[GROUP_SIZE];
```
rkennke, line 32: consumers[] is read in stopGracefully and written in rebalanceLoop. If the rebalancer.join(timeout/2) times out (e.g. it's mid-Thread.sleep and the interrupt is slightly delayed), stopGracefully then iterates consumers[] concurrently with the rebalancer mutating it. Benign for a stress harness, but a volatile/snapshot would make it clean.

ddprof-lib/src/test/cpp/stress_threadLifecycle_ut.cpp
```diff
@@ -0,0 +1,206 @@
...
+static constexpr const char STRESS_TEST_NAME[] = "StressThreadLifecycle";
+
+// Number of churn workers and iterations per worker.
+static constexpr int kChurnWorkers = 16;
+static constexpr int kChurnIterations = 2000;
+
+// Shared dump-side storage. Churn workers write it through the same shard-lock
+// protocol as profiler sample paths; the dump thread processes it under
+// lock_all(), matching Profiler::rotateDictsAndRun().
+static CallTraceStorage g_storage;
```
rkennke, line 41: global g_storage is never reset between tests and grows to ~32k traces over a run. Harmless, but a clear() in setup would keep tests independent

rkennke (changes requested):
Good change, I have a few minor-ish requests.

.gitlab/reliability/.gitlab-ci.yml
jbachorik: I think it's fine. This is run nightly, so the number of executions is limited.

.gitlab/reliability/.gitlab-ci.yml
jbachorik: Done — fixed by replacing `${CHAOS_JDK//./_}` with `${CHAOS_JDK//[.-]/_}` in both `script` and `after_script`.

ddprof-lib/src/test/cpp/stress_threadLifecycle_ut.cpp
jbachorik: Fixed: replaced `ASSERT_NE(nullptr, self)` with `EXPECT_NE(nullptr, self)` followed by `if (!self) return;`, matching the convention in `thread_teardown_safety_ut.cpp`. Also added `g_storage.clear()` at the start of `ChurnOnly` and `ChurnDuringDump` to keep tests independent.

ddprof-stresstest/src/chaos/java/com/datadoghq/profiler/chaos/ConsumerGroupAntagonist.java
jbachorik: Fixed: take a snapshot via `consumers.clone()` after setting `running = false` so `stopGracefully` iterates a stable copy, avoiding the concurrent read/write race window if `rebalancer.join` times out.

ddprof-lib/src/test/cpp/stress_threadLifecycle_ut.cpp
jbachorik: Fixed: added `g_storage.clear()` at the start of both `ChurnOnly` and `ChurnDuringDump` (the two tests that drive `churn_worker` which writes to `g_storage`) to keep them independent.

jbachorik:
@rkennke Fixed the finding. Thanks for the review!

rkennke (approved):
Ok great, looks good now!

PR #560, opened 2026-05-29
feat(profiler): wall-clock signal suppression for idle threads

kaahos:
**What does this PR do?**:

Before sending `SIGVTALRM`, the wall-clock timer checks whether the target thread is already in a skippable OS state and has already produced a sample during this blocking run. If so, the signal is suppressed. The first sample of every blocking run is always collected.
 
**Motivation**:

High-frequency wall-clock sampling on threads spending most of their time idle generates many redundant samples of identical stacks at high CPU cost. The precheck lets us emit one representative sample per blocking run and skip the rest, recovering signal throughput for threads that are actually running.

This is the first of two PRs splitting `paul.fournillon/wallclock_precheck`. This PR ships the suppression infrastructure. The follow-up PR adds `datadog.TaskBlock` event recording.

**Additional Notes**:

**How to test the change?**:

New unit tests:
- `ddprof-lib/src/test/cpp/park_state_ut.cpp`
- `ddprof-lib/src/test/cpp/wallprecheck_args_ut.cpp`

New integration tests:
- `PrecheckTest`: verifies suppression fires and reduces signal count.
- `PrecheckEfficiencyTest`: measures overhead reduction under sustained park.
- `WallclockMitigationsCombinedTest`: end-to-end combined mitigations scenario.

Run locally with:
`./gradlew :ddprof-test:test --tests '*Precheck*' --tests '*WallclockMitigations*' --tests '*WallclockTimer*'`

**For Datadog employees**:
- [x] JIRA: [JIRA-14354]


kaahos:
@copilot+gpt-5.2-codex review this PR

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -314,6 +317,87 @@ Java_com_datadoghq_profiler_JavaProfiler_recordQueueEnd0(
...
+  }
+  ThreadFilter *tf = Profiler::instance()->threadFilter();
+  if (!tf->enabled()) {
+    return 0;
+  }
+  ThreadFilter::SlotID slot_id = current->filterSlotId();
+  if (slot_id < 0) {
+    return 0;
+  }
+  tf->enterBlockedRun(slot_id, decoded);
```
jbachorik: **[Sphinx Review — HIGH]** `blockEnter0` discards the return value of `enterBlockedRun` and always returns `slot_id+1`, hiding the case where the block was not actually armed (slot already owned by `parkEnter` or another lifecycle path). `blockExit0` then calls the unconditional `exitBlockedRun` overload, which clears whatever block run is active regardless of owner — destroying `parkEnter`'s `CONDVAR_WAIT` state.

**Suggestion:** Return 0 from `blockEnter0` when `enterBlockedRun` returns 0 (failed to arm). In `blockExit0`, use the generation-checked `exitBlockedRun(slot_id, generation)` overload by decoding the generation from the stored token.

> Confirmed by adversary. Corroborated by specialist at line 378.

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -314,6 +317,87 @@ Java_com_datadoghq_profiler_JavaProfiler_recordQueueEnd0(
...
+    JNIEnv *env, jclass unused, jlong blocker, jlong unblockingSpanId) {
+  ProfiledThread *current = ProfiledThread::current();
+  if (current == nullptr) {
+    return;
+  }
+  u64 start_ticks = 0;
+  Context park_context = {};
+  current->parkExit(start_ticks, park_context);
+  ThreadFilter *tf = Profiler::instance()->threadFilter();
+  if (tf->enabled()) {
```
jbachorik, line 349: **[Sphinx Review — HIGH]** `parkExit0` ignores the return value of `current->parkExit()` and always calls `tf->exitBlockedRun(slot_id)` unconditionally — even when the thread was not parked (FLAG_PARKED was not set) or when the slot is owned by `blockEnter0`. This clears an active block run it does not own, disabling suppression for the remaining sleep interval.

**Suggestion:** `bool was_parked = current->parkExit(start_ticks, park_context); if (was_parked) tf->exitBlockedRun(slot_id);` Alternatively store the `enterBlockedRun` token during `parkEnter0` and use the generation-checked overload.

> Confirmed by adversary.

ddprof-lib/src/main/cpp/threadFilter.cpp
```diff
@@ -284,6 +287,109 @@ void ThreadFilter::collect(std::vector<int>& tids) const {
...
+
+void ThreadFilter::clearActive() {
+    int num_chunks = _num_chunks.load(std::memory_order_acquire);
+    for (int chunk_idx = 0; chunk_idx < num_chunks; ++chunk_idx) {
+        ChunkStorage* chunk = _chunks[chunk_idx].load(std::memory_order_acquire);
+        if (chunk == nullptr) {
+            continue;
+        }
+
+        for (auto& slot : chunk->slots) {
```
jbachorik, line 318: **[Sphinx Review — MEDIUM]** `clearActive()` calls both `slot.resetSampledRun(UNKNOWN)` and `slot.clearActiveBlockRun(UNKNOWN)` per slot. `clearActiveBlockRun` internally calls `resetSampledRun`, so it executes twice per slot on every profiler start — redundant atomic stores.

**Suggestion:** Remove the explicit `slot.resetSampledRun()` call from `clearActive()`; `clearActiveBlockRun()` already performs this reset.

ddprof-lib/src/main/cpp/wallClock.cpp
```diff
@@ -191,29 +281,43 @@ void WallClockASGCT::timerLoop() {
...
+      // thread cannot prove run boundaries for the target thread.
+      if (_precheck && entry.slot != nullptr) {
+        OSThreadState block_state = entry.slot->activeBlockState();
+        if (entry.slot->activeBlockOwner() != BlockRunOwner::NONE &&
+            isPrecheckSuppressionState(block_state) &&
+            entry.slot->sampledThisRun() &&
+            block_state == entry.slot->lastSampledState()) {
+          Counters::increment(WC_SIGNAL_SUPPRESSED_SAMPLED_RUN);
+          WallClockCounters::incrementSuppressedSampledRun();
+          const_cast<ThreadFilter::Slot*>(entry.slot)->incrementSuppressedSampleCount();
```
jbachorik: **[Sphinx Review — MEDIUM · CONSENSUS]** `ThreadEntry::slot` is declared `const Slot*` but the timer thread mutates it via `const_cast` to call `incrementSuppressedSampleCount()`. The `const` qualifier incorrectly signals to readers that the timer only reads slot state.

**Suggestion:** Change `ThreadEntry::slot` from `const ThreadFilter::Slot*` to `ThreadFilter::Slot*` in `threadFilter.h`. The slot objects in `ChunkStorage` are non-const; the `const` qualifier was incorrect.

ddprof-lib/src/main/cpp/threadFilter.h
```diff
@@ -48,17 +203,45 @@ class ThreadFilter {
...
+    // Clears per-recording membership and suppression state while keeping
+    // process-lifetime slot ownership intact. Threads must opt in again with add().
+    void clearActive();
+    void resetSlotRunState(SlotID slot_id);
+    u64 enterBlockedRun(SlotID slot_id, OSThreadState state,
+                        BlockRunOwner owner = BlockRunOwner::JAVA);
+    // Unconditional cleanup only. Normal block lifecycles must use the
+    // generation-checked overload so they cannot clear another owner.
+    void exitBlockedRun(SlotID slot_id);
+    bool exitBlockedRun(SlotID slot_id, u32 generation);
```
jbachorik, line 199: **[Sphinx Review — MEDIUM · CONSENSUS]** Generation-checked overloads — `exitBlockedRun(SlotID, u32)`, `snapshotAndExitBlockedRun`, `snapshotBlockedRun`, `encodeBlockRunToken`, `tokenSlotId`, `tokenGeneration`, `BlockRunSnapshot`, `Slot::snapshotBlockRun` — have zero production callers. The comment at line 213 says *"Normal block lifecycles must use the generation-checked overload so they cannot clear another owner"* but `blockEnter0`/`blockExit0` never do.

**Suggestion:** Wire the generation-checked path through `blockEnter0`/`blockExit0` (fixing the HIGH correctness finding), which activates this dead API with correct semantics. Alternatively defer all generation-checked API to PR2.

ddprof-lib/src/main/cpp/wallClockCounters.h
```diff
@@ -0,0 +1,32 @@
...
+private:
+  inline static std::atomic<long long> _suppressed_sampled_run{0};
+
+public:
+  static void incrementSuppressedSampledRun() {
+    _suppressed_sampled_run.fetch_add(1, std::memory_order_relaxed);
+  }
+
+  static u64 drainSuppressedSampledRun() {
+    return (u64)_suppressed_sampled_run.exchange(0, std::memory_order_acquire);
```
jbachorik: **[Sphinx Review — MEDIUM]** `drainSuppressedSampledRun` uses `memory_order_acquire` on the exchange. On ARMv8, acquire on an exchange issues `LDAR` but not `STLR`, so concurrent `relaxed` `fetch_add` increments from signal handlers on other cores can be architecturally reordered past the drain, attributing them to the next epoch rather than the current one.

**Suggestion:** Change to `memory_order_acq_rel`: `_suppressed_sampled_run.exchange(0, std::memory_order_acq_rel)` — this issues the `STLR`+`LDAR` pair on ARMv8 required to prevent in-flight relaxed increments from being reordered past the drain.

ddprof-lib/src/main/cpp/wallClockCounters.h
```diff
@@ -0,0 +1,32 @@
...
+
+static_assert(std::atomic<long long>::is_always_lock_free,
+              "WallClockCounters fields must be lock-free for signal-handler safety");
+
+// Holds signal-handler-safe atomic counters for wall-clock sampling statistics.
+// Increments use relaxed lock-free atomics; drains use atomic exchange, so each
+// increment is counted in either the current drain or a later one.
+class WallClockCounters {
+private:
+  inline static std::atomic<long long> _suppressed_sampled_run{0};
```
jbachorik, line 20: **[Sphinx Review — MEDIUM]** `_suppressed_sampled_run` is an `inline static` never reset by profiler stop or restart. After `BaseWallClock::stop()` joins the timer thread without a final drain, any `SIGVTALRM` signals handled after the last drain increment this counter. The next session's first `WallClockEpochEvent` drains these stale counts, inflating `_num_suppressed_sampled_run`.

**Suggestion:** Add `static void reset() { _suppressed_sampled_run.exchange(0, memory_order_relaxed); }` and call it from `WallClockASGCT::initialize()` or `BaseWallClock::start()` before the timer loop begins.

ddprof-test/src/test/java/com/datadoghq/profiler/wallclock/WallclockMitigationsCombinedTest.java
```diff
@@ -0,0 +1,113 @@
...
+        sleeping.join(1000);
+        parkedBusy.join(1000);
+        runnable.join(1000);
+
+        stopProfiler();
+
+        // Runnable thread must still produce samples (precheck doesn't suppress RUNNABLE).
+        long sampleCount = verifyEvents("datadog.MethodSample", false)
+                .getAggregate(Aggregators.count()).longValue();
+        assertTrue(sampleCount > 0, "Expected MethodSample events from runnable thread");
```
jbachorik: **[Sphinx Review — LOW]** `precheckAndParkSuppressionWorkTogether` asserts only `assertTrue(sampleCount > 0)` — a stub that never suppresses any signal would pass this. The test javadoc says *"TaskBlock assertions are in PR2"* with no issue tracker reference.

**Suggestion:** Add an upper-bound assertion on the sleeping thread's sample count (analogous to `PrecheckTest.testSleepingThreadIsNotSampled`), and add a Jira/GitHub issue reference for the deferred PR2 `TaskBlock` assertions.

jbachorik:
**[Sphinx Review — HIGH · CONSENSUS]** `wallClock.cpp` · `WallClockJvmti::timerLoop` · completeness

`WallClockJvmti::timerLoop()` and `signalHandler()` have no `wallprecheck` logic. `WallClockJvmti::initialize()` never reads `args._wall_precheck` and the class has no `_precheck` member. When `jvmtistacks=true` is combined with `wallprecheck=true` (valid on HotSpot JDK 11+ where `canRequestStackTrace()` returns true), the precheck suppression feature is silently a no-op — no timer-thread fast-path, no signal-handler suppression, no counter increments.

**Suggestion:** Either (a) add the same once-per-run precheck logic to `WallClockJvmti::timerLoop`/`signalHandler` mirroring the `WallClockASGCT` path (switching to `timerLoopCommon<ThreadEntry>` with `collect(entries)`), or (b) document the incompatibility and emit a warning/error in `Profiler::start()` when both `jvmtistacks` and `wallprecheck` are enabled.

> Confirmed by adversary.

jbachorik:
**[Sphinx Review — MEDIUM]** `thread.h` · `ProfiledThread` park infrastructure · necessity

`_park_start_ticks` and `_park_context` are captured on every `LockSupport.park` call but `parkExit0` immediately discards both output values. `FLAG_PARKED` is set/cleared but never consulted in any production path. `setJavaThread` was made more expensive (CAS loop) to preserve `FLAG_PARKED` that nothing reads. The PR description says `TaskBlock` event recording is in PR2.

**Suggestion:** Either strip `_park_start_ticks`, `_park_context`, and `FLAG_PARKED` from `ProfiledThread` until they are consumed in PR2, or add a Jira ticket reference linking this scaffolding to the follow-on work. If kept, revert `setJavaThread` to the simpler non-CAS form until there is a real reader for `FLAG_PARKED`.

jbachorik:
**[Sphinx Review — LOW · CONSENSUS]** `thread.h` · `setJavaThread` comment · consistency

The comment justifying the CAS RMW states `parkEnter`/`parkExit` run *"in signal-handler context"* but they are JNI calls from Java application threads, not signal handlers. The concurrent access that justifies the CAS is between the owner thread calling `parkEnter`/`parkExit` (JNI) and a different thread calling `setJavaThread` (JVMTI `onThreadStart` or `vmStructs`).

**Suggestion:** Replace with: *"CAS RMW to update only TYPE_MASK bits without clobbering FLAG_PARKED, which may be written concurrently by `parkEnter`/`parkExit` JNI calls on the owner thread while a JVMTI callback modifies thread type on a different thread."*

ddprof-lib/src/main/cpp/javaApi.cpp
kaahos: addressed in 6ece88da2.

ddprof-lib/src/main/cpp/threadFilter.cpp
kaahos: addressed in 110abaa41.

ddprof-lib/src/main/cpp/javaApi.cpp
kaahos: addressed in 6ece88da2.

ddprof-lib/src/main/cpp/wallClock.cpp
kaahos: addressed in 110abaa41.

ddprof-lib/src/main/cpp/threadFilter.h
kaahos: addressed in 6ece88da2.

ddprof-lib/src/main/cpp/wallClockCounters.h
kaahos: addressed in e2d60da21.

ddprof-lib/src/main/cpp/wallClockCounters.h
kaahos: I am not really sure about this one, I have addressed it anyway in 43471ee65 as it looks an harmless change to me.

ddprof-test/src/test/java/com/datadoghq/profiler/wallclock/WallclockMitigationsCombinedTest.java
kaahos: addressed in 37df0a587.

kaahos:
@copilot review this PR.

kaahos:
most of the previous comments have been addressed @jbachorik! The branch now also includes the new features coming from main.

ddprof-lib/src/main/cpp/threadFilter.h
```diff
@@ -48,17 +203,46 @@ class ThreadFilter {
...
+    u64 enterBlockedRun(SlotID slot_id, OSThreadState state,
+                        BlockRunOwner owner = BlockRunOwner::JAVA);
+    // Unconditional cleanup for reset/unregister paths only. Normal block
+    // lifecycles must use the generation-checked overload so they cannot clear
+    // another owner.
+    void exitBlockedRun(SlotID slot_id);
+    bool exitBlockedRun(SlotID slot_id, u32 generation);
+    bool snapshotAndExitBlockedRun(SlotID slot_id, u32 generation,
+                                   BlockRunSnapshot* snapshot);
+    BlockRunSnapshot snapshotBlockedRun(SlotID slot_id) const;
```
rkennke: Looks like this whole snapshot subsystem is unused in this PR, is that correct? Is it meant for staging later PRs? If so, would it make more sense to include that subsystem in the PR that actually needs it? Or is it cleaner/easier to just have it now?

ddprof-lib/src/main/cpp/thread.h
rkennke: You changed other accesses on _misc_flags to use _atomic* reads and writes, and should to it here, too, for consistency.

ddprof-lib/src/main/cpp/livenessTracker.cpp
```diff
@@ -216,12 +216,10 @@ Error LivenessTracker::initialize(Arguments &args) {
     return Error::OK;
   }
 
-  // _record_heap_usage controls per-session JFR event emission only, not the
```
rkennke, line 219: Are the changes in this file even related? Otherwise please remove them and propose them in a separate PR.

ddprof-lib/src/main/cpp/hotspot/vmStructs.inline.h
```diff
@@ -9,12 +9,21 @@
 
 #include "hotspot/vmStructs.h"
 #include "jvmThread.h"
+#include "vmEntry.h"
 
 VMThread* VMThread::current() {
+    // JVMThread::current() is the native thread self pointer. On OpenJ9/Zing it
+    // is not a HotSpot JavaThread*; only HotSpot may reinterpret it as VMThread*.
```
rkennke: I am unsure about this change. It is certainly good to protect against wrong casts here. However, it seems like this introduces some implicit knowledge. Maybe, if this can only safely be done on HotSpot, then maybe assert that, and make the calling code explicitly guarding on HotSpot-only? Not sure...

ddprof-lib/src/main/cpp/hotspot/vmStructs.inline.h
```diff
@@ -9,12 +9,21 @@
...
+    if (!VM::isHotspot() || JVMThread::current() == nullptr) {
```
rkennke: Calling current() twice and checking it for nullptr here is unnecessary - simply cast the returned nullptr, that should work fine, I think.

rkennke (changes requested):
This is a nice improvement! I have some questions...

ddprof-lib/src/main/cpp/threadFilter.h
kaahos: I agree, this belongs to the next PR not this one. I have removed it in 3df9af61.

ddprof-lib/src/main/cpp/thread.h
kaahos: agreed, addressed in 726eb80b.

ddprof-lib/src/main/cpp/livenessTracker.cpp
kaahos: they were not, the issue has already been addressed on main so I removed these changes in 726eb80b.

ddprof-lib/src/main/cpp/hotspot/vmStructs.inline.h
kaahos: I agree. I have added an assert so it does not silently fail anymore and have also added a guard for HotSpot-only. it has been addressed in 726eb80b.

ddprof-lib/src/main/cpp/hotspot/vmStructs.inline.h
kaahos: addressed in 726eb80b.

ddprof-lib/src/main/cpp/wallClock.cpp
```diff
@@ -27,6 +29,126 @@
...
+  if (isPrecheckSuppressionState(result.observed_state)) {
+    if (!slot->shouldRecordUnownedBlockedSample()) {
+      Counters::increment(WC_UNOWNED_BLOCKED_SUPPRESSED);
+      result.suppress = true;
+      return result;
+    }
+    result.unowned_weight_slot = slot;
+    result.unowned_weight = slot->consumeUnownedBlockedWeight();
+  } else {
+    slot->resetUnownedBlockedSampling();
```
rkennke: **Unowned path: the accumulated tail weight is discarded on a blocked→runnable transition.**

`shouldRecordUnownedBlockedSample()` accumulates `unowned_blocked_pending_weight` for each suppressed signal, but that pending weight is only flushed into a real sample when the *next* record-decision fires (`decision % kUnownedBlockedFallbackRatio == 1`). When the thread goes runnable, this `else` branch calls `resetUnownedBlockedSampling()`, which zeroes the pending weight — so the suppressed intervals since the last recorded sample are dropped rather than attributed.

Concretely, at `wall=1ms` a ~9ms unowned block that ends before the 11th signal emits exactly one sample of weight 1, undercounting it ~9×. The "first sample of every run" guarantee still holds, but total blocked-time accounting is systematically under-reported for short/medium unowned blocks (the only production-active path, since precheck is opt-in).

Is this approximation intended? Would it be more accurate to flush the pending weight into the last sample (or emit a final weighted sample) on the runnable transition instead of discarding it?

ddprof-lib/src/main/cpp/wallClock.cpp
```diff
@@ -27,6 +29,126 @@
...
+  OSThreadState active_block_state = slot->activeBlockState();
+  BlockRunOwner active_block_owner = slot->activeBlockOwner();
+  bool has_owned_block =
+      active_block_owner != BlockRunOwner::NONE &&
+      isPrecheckSuppressionState(active_block_state);
+  if (has_owned_block) {
+    if (slot->sampledThisRun() &&
+        active_block_state == slot->lastSampledState()) {
+      incrementSuppressedSampledRun();
+      result.suppress = true;
```
rkennke, line 116: **Owned path: no weight compensation for suppressed signals.**

Unlike the unowned path, the owned-block branch emits one sample of weight 1 and then fully suppresses every subsequent signal for the run with no accumulated weight. So a long instrumented sleep/park is attributed only a single interval of wall time, and the rest of the blocked duration is lost from the wall-clock view — presumably to be recovered by PR2's `datadog.TaskBlock` event.

This is safe in *this* PR because the only callers of `parkEnter/blockEnter` are tests. Could you confirm the owned hooks won't be enabled on any production/tracer path before PR2 lands? Otherwise instrumented blocks would undercount wall time in the interim.

rkennke (commented):
Two follow-up accuracy/design questions on the weighted-sampling logic (not blockers). Thanks for addressing all my earlier points thoroughly.

ddprof-lib/src/main/cpp/wallClock.cpp
kaahos: that's a good catch, thanks. I fixed this by preserving the last recorded unowned blocked stack and flushing the accumulated pending tail weight when the thread transitions back to runnable, instead of dropping it in `resetUnownedBlockedSampling()`. addressed in 5e484d0d8.

ddprof-lib/src/main/cpp/wallClock.cpp
kaahos: I confirm, in this PR the owned hooks are not wired to production/tracer instrumentation; the only current callers are tests.

To make that harder to misuse before PR2, I made the Java wrappers package-private and moved test access behind a test-only ProfilerOwnedBlockHooks bridge. The native entry points remain private, and PR2 will be where production TaskBlock instrumentation is introduced. addressed in 3bfa844b.

rkennke (approved):
Thanks for making the changes. The CI failures appear to be infra issues (failure to upload artifacts). Good to go! Thanks!

PR #568, opened 2026-06-01
fix: bounds-check ELF section and symbol-table parsing

edznux-dd:
**What does this PR do?**:

Add bound checks to the elf section and symbol table parsing.
This adds both unit test and minimized fuzz seeds to the repo (full reproducers are a bit large to be in Git IMO)

**Motivation**:
Fix overflow / out of bound reads.

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:
- [ ] JIRA: [JIRA-XXXX]


ddprof-lib/src/main/cpp/symbols_linux.cpp
```diff
@@ -559,7 +626,10 @@ void ElfParser::parseDynamicSection() {
         }
 
         if (!_cc->hasDebugSymbols() && nsyms > 0) {
-            loadSymbolTable(symtab, syment * nsyms, syment, strtab);
+            // strsz (DT_STRSZ) bounds string-table reads. The dynamic symbol
+            // region is virtual-address-relative live memory, so it is iterated
+            // as before rather than clamped to the file image.
+            loadSymbolTable(symtab, syment * nsyms, syment, strtab, strsz);
```
jbachorik, line 664: **[MEDIUM · correctness]** When `DT_STRSZ` is absent from the dynamic section, `strsz` stays `0`. `strAt(strings, 0, off)` rejects every offset ≥ 1, so **all dynamic symbols are silently dropped** — a functional regression for any DSO that loads correctly without `DT_STRSZ`.

Suggestion: when `strsz == 0`, either fall back to the pre-hardening path for dynamic symbols, or emit a warning so the loss is observable rather than silent.

ddprof-lib/src/test/cpp/elfparser_ut.cpp
```diff
@@ -359,4 +362,117 @@ INSTANTIATE_TEST_SUITE_P(
...
+    ElfParser::parseFile(&cc, nullptr, path, /*use_debug=*/false);
+    unlink(path);
+}
+
+}  // namespace
+
+// e_shoff pointing far outside the image made findSection() dereference a wild
+// section-header pointer (ElfParser::at). 16 TB is reliably unmapped.
+TEST_F(ElfTest, sectionHeaderOffsetOutOfBounds) {
+    Elf64_Ehdr e = validEhdr();
```
jbachorik, line 449: **[LOW · robustness]** `parseElfBytes` leaks the `mkstemp` fd and temp file if this `ASSERT_EQ` fires: `close(fd)` and `unlink(path)` are only reached after the assertion, so a write failure (e.g. full `/tmp`) leaves both open.

Suggestion: close and unlink **before** asserting on the write result, or wrap `fd`/`path` in an RAII guard so cleanup runs regardless of which assertion fires.

jbachorik (commented):
Sphinx multi-stream review (generalist + specialist + adversary).

**Overall:** The core hardening is correct — offset-based iteration, division-based index bounds, `inImage`/`sectionAt`/`strAt` layering, and NUL-termination enforcement are all sound. Two additional gaps in the same file are noted below as they are directly related to the PR's stated goal.

---

**⚠ HIGH (out-of-diff, same function as the fix)** — `symbols_linux.cpp` ~line 642

`parseDynamicSection` captures `strsz` from `DT_STRSZ` but never uses it to guard the two `addImport` calls in the `.rela.plt`/`.rela.dyn` relocation loops:

```cpp
_cc->addImport((void**)(base + r->r_offset), strtab + sym->st_name);  // unbounded
```

Every other name lookup in the file now routes through `strAt()`; these two are the only survivors — an internally inconsistent gap given the PR's own discipline.

**Fix:** Replace `strtab + sym->st_name` with `strAt(strtab, strsz, sym->st_name)` and skip the import when it returns `NULL`, mirroring `loadSymbolTable`.

---

**MEDIUM (out-of-diff)** — `symbols_linux.cpp` ~line 497

`findProgramHeader` and `calcVirtualLoadAddress` iterate the program-header table with raw `e_phoff`/`e_phnum`/`e_phentsize` arithmetic and **no `inImage()` check** — asymmetric with the now-hardened section-header path. The only guard in `parseProgramHeaders` validates the start of the first entry but not the full table.

**Fix:** Add a `phdrAt()` accessor analogous to `sectionAt()`: validate `e_phentsize >= sizeof(ElfProgramHeader)`, `index < e_phnum`, and `inImage(pheader, sizeof(ElfProgramHeader))` before each dereference.

jbachorik:
**[HIGH · security]**  captures `strsz` from `DT_STRSZ` but never uses it to guard the two `addImport` calls in the `.rela.plt`/`.rela.dyn` relocation loops (~line 642):

```cpp
if (sym->st_name != 0) {
    _cc->addImport((void**)(base + r->r_offset), strtab + sym->st_name);  // unbounded
}
```

Every other name lookup in the file now routes through `strAt()`; these two are the only survivors — an internally inconsistent gap given the PR's own discipline. A corrupted or malicious symbol table entry with `st_name >= strsz` triggers an out-of-bounds read of process memory.

**Fix:** Replace `strtab + sym->st_name` with `strAt(strtab, strsz, sym->st_name)` and skip the import when it returns `NULL`, mirroring `loadSymbolTable`.

jbachorik:
**[MEDIUM · security]** `findProgramHeader` and `calcVirtualLoadAddress` iterate the program-header table with raw `e_phoff`/`e_phnum`/`e_phentsize` arithmetic and no `inImage()` bounds check (~line 497):

```cpp
const char* pheaders = (const char*)_header + _header->e_phoff;
for (int i = 0; i < _header->e_phnum; i++) {
    ElfProgramHeader* pheader = (ElfProgramHeader*)(pheaders + i * _header->e_phentsize);
    if (pheader->p_type == type) { ... }  // no inImage() guard
}
```

This is asymmetric with the now-hardened section-header path (`sectionAt()` validates bounds on every access). The only guard in `parseProgramHeaders` validates the start of the first entry but not the full table nor subsequent entries.

**Fix:** Add a `phdrAt()` accessor analogous to `sectionAt()`: validate `e_phentsize >= sizeof(ElfProgramHeader)`, `index < e_phnum`, and `inImage(pheader, sizeof(ElfProgramHeader))` before each dereference in both `findProgramHeader` and `calcVirtualLoadAddress`.

ddprof-lib/src/main/cpp/symbols_linux.cpp
```diff
@@ -348,29 +348,91 @@ class ElfParser {
...
+    }
+
+    // Program-header entry at `index`, or NULL when the index or entry is out of bounds.
+    ElfProgramHeader* phdrAt(int index) {
+        if (index < 0 || index >= _header->e_phnum
+                || _header->e_phentsize < sizeof(ElfProgramHeader)) {
+            return NULL;
+        }
+        ElfProgramHeader* ph = (ElfProgramHeader*)(
+            (const char*)_header + _header->e_phoff + (size_t)index * _header->e_phentsize);
```
jbachorik: **[CONSENSUS · MEDIUM · security]** `phdrAt()` / `contentAt()` — pointer-arithmetic UB before `inImage()` guard

Both `phdrAt()` and `contentAt()` form the file-offset pointer **before** `inImage()` validates it:

```cpp
// phdrAt (this line):
ElfProgramHeader* ph = (ElfProgramHeader*)((const char*)_header + _header->e_phoff + (size_t)index * _header->e_phentsize);
return inImage(ph, sizeof(ElfProgramHeader)) ? ph : NULL;

// contentAt (line ~411):
const char* p = (const char*)_header + s->sh_offset;
return inImage(p, want) ? p : NULL;
```

With an attacker-controlled `e_phoff` / `sh_offset` near `UINT64_MAX`, the addition overflows (UB). An optimising compiler may assume the addition does not overflow and elide the `>= _header` branch inside `inImage()`, letting the wrapped pointer pass the guard.

The constructor already applies the correct integer-first pattern for `_sections` / `e_shoff`:
```cpp
_sections = (image_size >= sizeof(ElfHeader) && _header->e_shoff < image_size)
    ? (const char*)addr + _header->e_shoff : NULL;
```

Suggested fix — apply the same pattern in both accessors:
```cpp
// contentAt — before forming the pointer:
if (s->sh_offset > (size_t)(_image_end - (const char*)_header)) return NULL;
const char* p = (const char*)_header + s->sh_offset;

// phdrAt — before forming the pointer:
if (_header->e_phoff > (size_t)(_image_end - (const char*)_header)) return NULL;
ElfProgramHeader* ph = (ElfProgramHeader*)((const char*)_header + _header->e_phoff + (size_t)index * _header->e_phentsize);
```

ddprof-lib/src/test/cpp/elfparser_ut.cpp
```diff
@@ -413,4 +443,69 @@ TEST(ElfBuildId, noteOffsetOverflow) {
...
+    sym.st_value = 0x1000;
+    const uint64_t str_off = sym_off + sizeof(sym);
+    const char strtab[] = "\0main";
+
+    Elf64_Ehdr e = validEhdr();
+    e.e_shoff = shoff;
+    e.e_shentsize = sizeof(Elf64_Shdr);
+    e.e_shnum = NSEC;
+    e.e_shstrndx = 3;
+
```
jbachorik, line 480: **[SPECIALIST · MEDIUM · test-adequacy]** All new regression tests use `use_debug=false` — `addRelocationSymbols` hardening is not CI-gated

```cpp
ElfParser::parseFile(&cc, nullptr, path, /*use_debug=*/false);  // every new test
```

`loadSymbols(use_debug)` only enters the `if (use_debug)` branch when `use_debug=true`, so `addRelocationSymbols` — which contains the most complex new hardening (`max_sym_index` bound, offset-based relocation iteration, bounded `strAt` lookup) — is never reached by a CI unit test. The only coverage is `fuzz_elf.cpp`, gated behind `FUZZING_BUILD_MODE_UNSAFE_FOR_PRODUCTION` and not run on every PR.

Suggested addition: a test that parses a minimal ELF with `.plt` + `.rela.plt` + `.symtab` + `.strtab` containing an out-of-bounds `r_info` symbol index, with `use_debug=true`, asserting no crash.

ddprof-lib/src/main/cpp/symbols_linux.cpp
```diff
@@ -569,7 +657,10 @@ void ElfParser::parseDynamicSection() {
                 ElfRelocation* r = (ElfRelocation*)(jmprel + offs);
                 ElfSymbol* sym = (ElfSymbol*)(symtab + ELF_R_SYM(r->r_info) * syment);
```
jbachorik, line 672: **[SPECIALIST · LOW · security]** Dynamic-section relocation loops have no symbol-index bound (unlike `addRelocationSymbols`)

```cpp
ElfSymbol* sym = (ElfSymbol*)(symtab + ELF_R_SYM(r->r_info) * syment);
if (sym->st_name != 0) { ... }   // dereferenced with no index check
```

`addRelocationSymbols` (file-section path) was hardened with `max_sym_index` in this PR; the parallel dynamic-section path here was not. A corrupt dynamic section with a large `r_info` symbol index or large `syment` yields an out-of-range `ElfSymbol*` that is then dereferenced.

Lower severity because this path operates on linker-validated live memory rather than an attacker-supplied file. Consider applying the same `max_sym_index` bound, or adding a comment that explicitly acknowledges the asymmetry.

jbachorik (commented):
**Sphinx Review** — `fix/elf-parser-oob` — 2026-06-02

3 findings: 2 × MEDIUM, 1 × LOW. The bounds-hardening is solid overall; the issues below are incomplete coverage of the same pattern the PR itself introduced.

Streams: generalist (Opus) · specialist (Haiku lens-match · Sonnet scout · Opus review, 0 lenses applied) · adversary (Opus, 0 refutations)
Total reviewed: 7 → 4 valid, 3 structurally invalid (mutation findings outside diff)

ddprof-lib/src/test/cpp/elfparser_ut.cpp
edznux-dd: I'll do that, but for context however, I'll later on setup the fuzzing infrastructure to run these test on a regular basis.

ddprof-lib/src/main/cpp/symbols_linux.cpp
```diff
@@ -348,29 +348,91 @@ class ElfParser {
...
     }
 
-    const char* at(ElfSection* section) {
-        return (const char*)_header + section->sh_offset;
+    // Start of a section's first `want` content bytes, or NULL if not fully mapped.
+    const char* contentAt(ElfSection* s, size_t want) {
+        if (s == NULL) {
+            return NULL;
+        }
+        const char* p = (const char*)_header + s->sh_offset;
```
jbachorik: **[CONSENSUS · HIGH · correctness · ADVERSARY-TESTED]** `contentAt()` / `phdrAt()` — pointer-arithmetic overflow UB fires under project's own sanitizer config

Both `contentAt()` (this line) and `phdrAt()` (lines 433-434) form `_header + attacker-controlled uint64` **before** `inImage()` validates the result:

```cpp
// contentAt:
const char* p = (const char*)_header + s->sh_offset;  // ← UB if sh_offset near UINT64_MAX
return inImage(p, want) ? p : NULL;

// phdrAt:
(const char*)_header + _header->e_phoff + (size_t)index * _header->e_phentsize
```

The project builds with `-fsanitize=pointer-overflow -fno-sanitize-recover=all` (`ConfigurationPresets.kt:171-176`) and the fuzz target with `-fsanitize=fuzzer,address,undefined` (`FuzzTargetsPlugin.kt:203`). Forming `_header + huge_uint64` is a **non-recoverable abort** under the project's own CI config when the fuzz corpus exercises a large-offset section. A wrap-around value (e.g. `UINT64_MAX - delta`) can also land back inside `[_header, _image_end)`, passing `inImage()` and returning a bogus in-image pointer.

The constructor already applies the safe integer-first pattern for `_sections`/`e_shoff`:
```cpp
_sections = (image_size >= sizeof(ElfHeader) && _header->e_shoff < image_size)
    ? (const char*)addr + _header->e_shoff : NULL;
```
`extractBuildIdFromMemory` uses it too. `contentAt`/`phdrAt` should match:
```cpp
// contentAt — validate offset before pointer formation:
size_t img_size = (size_t)(_image_end - (const char*)_header);
if (s->sh_offset > img_size || want > img_size - s->sh_offset) return NULL;
const char* p = (const char*)_header + s->sh_offset;

// phdrAt — similarly in integer space before casting:
size_t phoff = _header->e_phoff;
size_t stride = (size_t)index * _header->e_phentsize;
if (phoff > img_size || stride > img_size - phoff) return NULL;
ElfProgramHeader* ph = (ElfProgramHeader*)((const char*)_header + phoff + stride);
```

ddprof-lib/src/main/cpp/symbols_linux.cpp
```diff
@@ -468,7 +542,7 @@ bool ElfParser::parseFile(CodeCache* cc, const char* base, const char* file_name
 }
 
 void ElfParser::parseProgramHeaders(CodeCache* cc, const char* base, const char* end, bool relocate_dyn) {
-    ElfParser elf(cc, base, base, NULL, relocate_dyn);
+    ElfParser elf(cc, base, base, (size_t)(end - base), NULL, relocate_dyn);
     if (elf.validHeader() && base + elf._header->e_phoff < end) {
```
jbachorik, line 560: **[SPECIALIST · MEDIUM · correctness]** `parseProgramHeaders` outer guard — same pointer-overflow UB on `e_phoff`

```cpp
if (elf.validHeader() && base + elf._header->e_phoff < end) {
```

`e_phoff` is `Elf64_Off` (uint64_t) and fully attacker-controlled. The addition `base + e_phoff` is pointer arithmetic on an unbounded uint64 — same UBSan abort risk as `contentAt`/`phdrAt` above. Additionally a crafted `e_phoff` designed to overflow back to a small value satisfies `base + e_phoff < end` even when `e_phoff` clearly exceeds the image size, allowing the body to execute when it should be skipped.

The constructor checks `e_shoff < image_size` (safe integer form) a few lines above. Apply the same here:
```cpp
if (elf.validHeader() && elf._header->e_phoff < (size_t)(end - base)) {
```
This avoids forming the overflowed pointer entirely.

ddprof-lib/src/main/cpp/symbols_linux.cpp
```diff
@@ -558,8 +635,19 @@ void ElfParser::parseDynamicSection() {
...
 
+        // DT_STRSZ is required by the ELF spec whenever DT_STRTAB is present.
+        // When it is absent (strsz == 0) all string lookups via strAt() would
+        // be rejected, silently dropping every symbol. Fall back to unbounded
+        // scans instead — the dynamic section is already in live linker-validated
+        // memory, so NUL-terminated strings are guaranteed.
+        if (strsz == 0) {
+            Log::warn("DT_STRSZ absent from dynamic section in %s; string lookups will be unbounded",
+                      _file_name != NULL ? _file_name : "unknown");
+            strsz = (size_t)-1;
```
jbachorik: **[SPECIALIST · MEDIUM · robustness]** `strsz = (size_t)-1` fallback — unbounded `memchr` over live memory

```cpp
if (strsz == 0) {
    Log::warn("No DT_STRSZ in dynamic section of %s", _file_name);
    strsz = (size_t)-1;  // ← SIZE_MAX
}
```

Every subsequent `strAt(strtab, strsz, sym->st_name)` call resolves to `memchr(strtab + off, '\0', SIZE_MAX - off)`. `memchr` stops at the first NUL, so for a well-formed library this is fast. However:

1. **Fault**: if `DT_STRTAB` points to a memory region that has no NUL byte before an unmapped page (corrupt or hand-crafted library), `memchr` walks off the end of mapped memory and the profiler thread faults.
2. **DoS**: a pathological-but-valid library with very long symbol names causes `memchr` to scan megabytes per symbol per library load.

Cap the fallback to a sane upper bound — the fuzz target already uses 4 MB as its limit:
```cpp
strsz = 4 * 1024 * 1024;  // bound worst-case scan; matches fuzz target cap
```
This still handles all real-world libraries while preventing both failure modes.

jbachorik (commented):
**Sphinx Review (run 2)** — `fix/elf-parser-oob` — 2026-06-02

This run surfaced 1 × HIGH and 2 × MEDIUM not in the previous review. Posting only those here.

Streams: generalist (Opus) · specialist (Haiku + Sonnet + Opus) · adversary (Opus, HIGH confirmed)
Total: 10 reviewed → 9 valid

ddprof-lib/src/test/cpp/elfparser_ut.cpp
```diff
@@ -379,10 +391,28 @@ Elf64_Ehdr validEhdr() {
...
     e.e_ehsize = sizeof(Elf64_Ehdr);
-    e.e_shstrndx = 1;
+    e.e_shstrndx = 1;  // non-zero so validHeader() accepts the image
     return e;
 }
 
+// Write the bytes to a unique temp file and run ElfParser::parseFile over it,
+// mirroring how Symbols::parseLibraries() parses an on-disk library.
+void parseElfBytes(const std::vector<char>& bytes) {
+    char path[] = "/tmp/elf_regress_XXXXXX";
```
jbachorik, line 401: Shouldn't this file be deleted after/before the test? To make sure we are not transmitting any unwanted context?

jbachorik: Ah, the `unlink()` seems to be the convention everywhere. Plz, disregard.

jbachorik (approved):
Looks ok - not an ELF format expert but nothing seems to be obviously off.

ddprof-lib/src/main/cpp/symbols_linux.cpp
edznux-dd: fixed in e02b443c.

PR #571, opened 2026-06-01
fix: bound DWARF .eh_frame_hdr parsing to the section

edznux-dd:
**What does this PR do?**:

Fix bound checking on untrusted section size, avoid infinite loops, and fix small fuzz-harness only memory leak.

**Motivation**:

Make the dwarf parser more reliable.

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:
- [ ] JIRA: [JIRA-XXXX]

### DIFF
--- a/ddprof-lib/src/main/cpp/dwarf.cpp
+++ b/ddprof-lib/src/main/cpp/dwarf.cpp
@@ -135,7 +135,7 @@ static constexpr u8 omit_sign_bit_mask_low(u8 value) {
 
 void DwarfParser::parse(const char *eh_frame_hdr, size_t size, const char *image_end) {
   // Fixed .eh_frame_hdr header: version (1) + 3 encoding bytes + eh_frame_ptr (4)
-  // + fde_count at offset 8 (4), binary-search table starting at offset 16.
+  // + fde_count at offset 8 (4), binary-search table starting at offset 12.
   // Refuse anything too small.
   if (eh_frame_hdr == NULL || size < 16) {
     return;
@@ -168,19 +168,20 @@ void DwarfParser::parse(const char *eh_frame_hdr, size_t size, const char *image
   }
 
   u32 fde_count = *(u32 *)(eh_frame_hdr + 8);
-  u32 *table = (u32 *)(eh_frame_hdr + 16);
+  // Table starts at offset 12 (4-byte header + 4-byte eh_frame_ptr + 4-byte fde_count).
   // Each entry is a (initial_loc, fde_ptr) pair of 4-byte section-relative
   // offsets (DW_EH_PE_datarel | DW_EH_PE_udata4). Reject a count that would
   // make the table walk read past the section.
-  if (fde_count > (size - 16) / 8) {
+  u32 *table = (u32 *)(eh_frame_hdr + 12);
+  if (fde_count > (size - 12) / 8) {
     Log::warn("Truncated or invalid .eh_frame_hdr (fde_count=%u, size=%lu) in %s",
               fde_count, (unsigned long)size, _name);
     return;
   }
   for (u32 i = 0; i < fde_count; i++) {
-    // table[i*2+1] is the FDE pointer (datarel sdata4); table[i*2] is initial_loc.
-    // Cast to int32_t to correctly handle negative offsets (FDE before the header).
-    _ptr = eh_frame_hdr + (int32_t)table[i * 2 + 1];
+    // table[i*2] is initial_loc; table[i*2+1] is the FDE pointer (datarel sdata4).
+    // Cast to int to correctly handle negative offsets (FDE before the header).
+    _ptr = eh_frame_hdr + (int)table[i * 2 + 1];
     parseFde();
   }
 }
--- a/ddprof-lib/src/main/cpp/dwarf.cpp
+++ b/ddprof-lib/src/main/cpp/dwarf.cpp
@@ -96,6 +96,8 @@ FrameDesc FrameDesc::no_dwarf_frame = {0, DW_REG_INVALID, DW_REG_INVALID, DW_REG
 void DwarfParser::init(const char *name, const char *image_base) {
   _name = name;
   _image_base = image_base;
+  _section_start = NULL;
+  _section_end = reinterpret_cast<const char*>(~(size_t)0);
 
   _capacity = 128;
   _count = 0;
@@ -109,9 +111,10 @@ void DwarfParser::init(const char *name, const char *image_base) {
 }
 
 DwarfParser::DwarfParser(const char *name, const char *image_base,
-                         const char *eh_frame_hdr) {
+                         const char *eh_frame_hdr, size_t eh_frame_hdr_size,
+                         EhFrameHdrTag) {
   init(name, image_base);
-  parse(eh_frame_hdr);
+  parse(eh_frame_hdr, eh_frame_hdr_size);
 }
 
 DwarfParser::DwarfParser(const char *name, const char *image_base,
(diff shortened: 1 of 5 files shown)


ddprof-lib/src/main/cpp/dwarf.cpp
```diff
@@ -149,8 +163,23 @@ void DwarfParser::parse(const char *eh_frame_hdr) {
...
+  // would make the index walk read past the section. A negative count cannot
+  // be valid either.
+  if (fde_count < 0 || (size_t)fde_count > (size - 16) / 8) {
+    Log::warn("Truncated or invalid .eh_frame_hdr (fde_count=%d, size=%lu) in %s",
+              fde_count, (unsigned long)size, _name);
+    return;
+  }
   for (int i = 0; i < fde_count; i++) {
-    _ptr = eh_frame_hdr + table[i * 2];
+    const char *fde = eh_frame_hdr + table[i * 2];
```
jbachorik: **Pre-existing correctness bug — worth fixing while touching this code**

`table[i * 2]` is `initial_loc` (the PC coverage start address), not `fde_ptr` (the FDE location). The actual data-relative FDE pointer is at `table[i * 2 + 1]`. This predates the PR: before the bounds check was added, `parseFde()` was called with `_ptr` pointing into the code section (initial_loc values are code addresses), producing garbage unwind records. The new bounds check now correctly rejects all initial_loc-derived addresses as outside `.eh_frame_hdr`, turning silent garbage into a guaranteed total miss.

Fixing the index here is low-effort and would prevent this regression from landing.

There is also a secondary safety issue on this exact line: forming the pointer from a signed `int` before the guard fires is out-of-bounds pointer arithmetic (UB) when `table[i*2]` is negative or very large. Safer approach:
```cpp
int off = table[i * 2 + 1];   // fde_ptr, data-relative to eh_frame_hdr
if (off < 0 || (size_t)off >= size) continue;
const char *fde = eh_frame_hdr + off;
```

**Bounds window note**: even after fixing the index, `fde_ptr` values point into `.eh_frame` (a separate ELF section) and will also fall outside `[_section_start, _section_end)` as currently computed. Either expand the window using the `eh_frame_ptr` field at offset 4 of `.eh_frame_hdr` to also cover `.eh_frame`, or remove the per-entry in-section guard for `fde_ptr` — the `table_enc` validation at parse entry already guarantees valid `sdata4` data-relative encoding.

**Test feasibility**: feasible with existing infrastructure. Lay out a contiguous buffer: 16-byte `.eh_frame_hdr` header + 8-byte table entry (`fde_ptr` pointing at offset 24) + minimal CIE at offset 24 + minimal FDE immediately after. Assert `count() > 0`. Non-trivial but doable in this PR or a follow-up.

_Raised by Sphinx review._

ddprof-lib/src/main/cpp/dwarf.cpp
```diff
@@ -250,7 +282,7 @@ void DwarfParser::parseCie() {
 
   const char *cie_start = _ptr;
   _ptr += 5;
-  while (*_ptr++) {
+  while (_ptr < _section_end && *_ptr++) {
```
jbachorik, line 287: **Unconditional `_ptr += 5` above inconsistent with the bound-before-advance pattern this PR establishes**

The line immediately above (`_ptr += 5;`) advances without a `canRead(5)` guard, while every other fixed-size advance in this PR (`get8`, `get16`, `get32`, `getPtr`, `skipLeb`) now goes through `canRead` first. Latent: downstream `canRead` calls prevent any out-of-bounds read today, but the unconditional advance can form a pointer well past one-past-end (UB) and breaks the uniform invariant.

Suggested fix for the line above:
```cpp
if (!canRead(5)) { _ptr = _section_end; return; }
_ptr += 5;
```

_Raised by Sphinx review._

ddprof-lib/src/main/cpp/dwarf.h
```diff
@@ -118,39 +125,56 @@ class DwarfParser {
...
+            result |= (u32)(b & 0x7f) << shift;
             if ((b & 0x80) == 0) {
-                return result;
+                break;
             }
         }
+        return result;
     }
 
     u32 getLeb(const char* end) {
```
jbachorik, line 179: **`getLeb(end)` / `getSLeb(end)` trust caller not to exceed `_section_end`**

The bounded overloads loop on `_ptr < end` but do not clamp `end` against `_section_end`, unlike the no-arg siblings that go through `canRead()`. All current callers derive `end` from a validated record length (`≤ _section_end`), so this is safe today — but a future caller with an unvalidated `end` would silently over-read. Same issue on `getSLeb(const char* end)` below.

Suggested — at entry to both overloads:
```cpp
if (end > _section_end) end = _section_end;
```
or at minimum `assert(end <= _section_end)` to make the contract explicit.

_Raised by Sphinx review._

ddprof-lib/src/test/cpp/dwarf_ut.cpp
```diff
@@ -190,4 +190,26 @@ TEST(DwarfEhFrame, FdeAugDataOverrun) {
...
+    hdr[1] = 0x03;  // eh_frame_ptr_enc = DW_EH_PE_udata4
+    hdr[2] = 0x03;  // fde_count_enc    = DW_EH_PE_udata4
+    hdr[3] = 0x33;  // table_enc        = DW_EH_PE_datarel | DW_EH_PE_udata4
+    // fde_count at offset 8 (little-endian): claim 1024 entries that aren't there.
+    hdr[8] = 0x00;
+    hdr[9] = 0x04;
+
+    const char* base = reinterpret_cast<const char*>(hdr.data());
+    DwarfParser dwarf("test", base, base, hdr.size(), DwarfParser::EhFrameHdrTag{});
+    EXPECT_EQ(dwarf.count(), 0);  // rejected: no records, no crash
```
jbachorik, line 211: **Two coverage gaps on the new `EhFrameHdrTag` path**

1. **Non-discriminating assertion**: `EXPECT_EQ(dwarf.count(), 0)` holds for *any* early exit, not specifically the `fde_count` guard. The test exercises "no crash under ASan" but cannot distinguish the guard from a stub that always returns. Consider adding: a case with `fde_count=1` + one valid 8-byte table entry (guard allows it, should produce records once the index fix is in place), and a case with `fde_count=2` + only 8 bytes of table data (guard rejects).

2. **Missing path**: `parseFde()`'s `_ptr = fde_start - cie_offset` walk is not exercised through this constructor at all — that's the highest-risk pointer arithmetic in the diff (see review body for line 301). A test with a valid header + table entry + FDE where `cie_offset > (fde_start - _section_start)` asserting no crash (best run under ASan) would cover the fix for that finding.

Additionally, six boundary mutations have no killing tests (low priority, can be deferred): `canRead()` at `_section_end-1`; `getLeb()`/`getSLeb()` `shift < 32` with a 5-byte value; `addRecord()` at `cfa_reg = 255/256` and `cfa_off = ±8388607/8388608`.

_Raised by Sphinx review._

jbachorik (commented):
## Sphinx review — issues in unchanged `parseFde()` body

Two findings that couldn't be anchored as inline comments because they fall in the unchanged part of `parseFde()`:

### `dwarf.cpp:301` — unguarded `cie_offset` underflow (HIGH)

`cie_offset` is a `u32` read directly from the untrusted, memory-mapped ELF image. When `cie_offset > (fde_start - _section_start)`, the subtraction `fde_start - cie_offset` underflows past the start of the mapped allocation — undefined behaviour. A crafted small value also redirects CIE parsing to an attacker-controlled position inside the section, defeating the bounds model this PR establishes.

```cpp
if (cie_offset > (size_t)(fde_start - _section_start)) {
    return;
}
_ptr = fde_start - cie_offset;
```

### `dwarf.cpp:308` — unclamped `_ptr` advance after untrusted LEB (MEDIUM)

`_ptr += getLeb()` advances by a decoded value from untrusted input without clamping to `_section_end`. A large value pushes `_ptr` past the section boundary, silently truncating the rest of the FDE. Same issue in `parseInstructions()` at the `DW_CFA_def_cfa_expression` and `DW_CFA_expression` branches (lines ~404, ~409).

After each unclamped advance: `if (_ptr > _section_end) _ptr = _section_end;`

_Raised by Sphinx review._

jbachorik (approved):
LGTM

PR #572, opened 2026-06-02
fix(spinlock): bound OptionalSharedLockGuard spin in signal handler paths

jbachorik:
**What does this PR do?**:
`OptionalSharedLockGuard` previously used an unbounded `tryLockShared` spin loop that could stall indefinitely inside signal handlers under heavy reader CAS contention. This PR replaces the unbounded variant with a bounded one (default budget: 256 CAS-retry iterations) and makes it the sole implementation.

**Motivation**:
Signal handlers must return promptly. Under heavy concurrent sampling (many threads recording events simultaneously), CAS contention on `_rec_lock` could cause signal handlers reaching `FlightRecorder::recordEvent*` to spin for an unbounded number of iterations. When the budget is exhausted, the sample is dropped and a new `JVMTI_STACKS_DROPPED_REC_LOCK` counter is incremented for observability.

**Additional Notes**:
- `OptionalSharedLockGuard` now takes an optional `max_spins` parameter (default `SpinLock::DEFAULT_SHARED_SPIN_BUDGET = 256`). The budget bounds CAS-retry iterations, not wall-clock latency.
- `unlock()` and `unlockShared()` gain debug-mode `assert`s to catch mismatched-variant misuse as the class grows.
- All 8 existing `OptionalSharedLockGuard` usages in `flightRecorder.cpp` pick up the bounded behaviour automatically — no call-site changes needed.
- `BoundedOptionalSharedLockGuard` was an intermediate name used during development; it is not present in this PR.

**How to test the change?**:
New gtest suite: `ddprof-lib/src/test/cpp/spinLock_ut.cpp`
- `OptionalGuard_UncontendedAcquire` — acquires and releases correctly
- `OptionalGuard_ExclusiveHeld_ImmediateReturn` — exclusive lock causes immediate false return (no spin)
- `OptionalGuard_SpinBudgetBound` — tiny budget under reader CAS contention; guard returns without hanging
- `OptionalGuard_BudgetEnforced_ExclusivePath` — deterministic: exclusive lock always returns false for any budget
- `OptionalGuard_SharedReentrancy` — multiple shared guards coexist; exclusive try fails
- `TryLockShared_ExclusiveHeld_ReturnsFalse` / `TryLockShared_Free_ReturnsTrue` — unbounded overload sanity

Run with:
```
./gradlew :ddprof-lib:gtestDebug
```

**For Datadog employees**:


ddprof-lib/src/main/cpp/spinLock.h
```diff
@@ -89,12 +118,17 @@ class SharedLockGuard {
...
   SpinLock* _lock;
 public:
-  explicit OptionalSharedLockGuard(SpinLock* lock) : _lock(lock) {
-    if (!_lock->tryLockShared()) {
-      // Exclusive lock is held; no unlock needed. Only fails when an exclusive lock is observed.
+  explicit OptionalSharedLockGuard(
+      SpinLock* lock, int max_spins = SpinLock::DEFAULT_SHARED_SPIN_BUDGET)
+      : _lock(lock) {
+    if (!_lock->tryLockShared(max_spins)) {
       _lock = nullptr;
```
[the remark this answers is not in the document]
jbachorik: Fixed in 444462b66: added `SAMPLES_DROPPED_REC_LOCK` counter and an `else` branch in `recordEvent()` mirroring the existing `JVMTI_STACKS_DROPPED_REC_LOCK` in `recordEventDelegated()`.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -2111,6 +2113,8 @@ void FlightRecorder::recordEventDelegated(int lock_index, int tid,
       rec->flushIfNeeded(buf);
       rec->addThread(lock_index, tid);
     }
+  } else {
+    Counters::increment(JVMTI_STACKS_DROPPED_REC_LOCK);
```
kaahos: nit: in [profiler.cpp:1485](https://github.com/DataDog/java-profiler/blob/444462b66aafaf46f69c3aa3adcd0b4efbb0bd37/ddprof-lib/src/main/cpp/profiler.cpp#L1485), I think that `JVMTI_STACKS_DROPPED_LOCK` and `JVMTI_STACKS_DROPPED_REC_LOCK` should be summed up for the error message to convey relevant info.

kaahos (approved):
It looks good to me! I've just suggested a small non-blocking modification in [profiler.cpp:1485](https://github.com/DataDog/java-profiler/blob/444462b66aafaf46f69c3aa3adcd0b4efbb0bd37/ddprof-lib/src/main/cpp/profiler.cpp#L1485).

ddprof-lib/src/main/cpp/flightRecorder.cpp
jbachorik: removed JVMTI_STACKS_DROPPED_REC_LOCK entirely — both recordEvent() and recordEventDelegated() now report rec-lock drops via SAMPLES_DROPPED_REC_LOCK, regardless of the call path.

PR #576, opened 2026-06-04
[PROF-14883] Publish thread context attribute keys at process start

ivoanjo:
**What does this PR do?**

This PR unifies the `registerAttributeKeys` mechanism used for supporting the "OTel thread context" with `setProcessContext` so that they get both published together at process start.

**Motivation:**

Beyond the code simplification (most of this PR is deletes), the big advantage of this approach is that `registerAttributeKeys` happened at profiler start, and now this is all moved to process start.

This was meh because profiler start is by default delayed (afaik up to 70 seconds in practice) which means that the "thread context" information would be missing for the same period of time, which was super confusing, and would mean an outside reader would be missing this data for that period.

**Additional Notes:**

This PR will pair with one on the dd-trace-java side to provide the needed info when setting the process context.

**How to test the change?**

This change includes test coverage + on the dd-trace-java coverage we'll add a few tests too.

**For Datadog employees**:
- [x] JIRA: PROF-14883

### DIFF
--- a/ddprof-lib/src/main/cpp/javaApi.cpp
+++ b/ddprof-lib/src/main/cpp/javaApi.cpp
@@ -473,6 +473,14 @@ Java_com_datadoghq_profiler_OTelContext_setProcessCtx0(JNIEnv *env,
       return;
     }
     jni_keys[built] = new JniString(env, jstr);
+    if (jni_keys[built]->c_str() == nullptr) {
+      // GetStringUTFChars failed (e.g. OOM); a NULL key pointer would truncate
+      // the published map mid-array, so abort the publish.
+      delete jni_keys[built];
+      for (int j = 0; j < built; j++) delete jni_keys[j];
+      Log::warn("setProcessContext: failed to read attribute key at index %d; skipping publish", i);
+      return;
+    }
     key_ptrs[i + 1] = jni_keys[built]->c_str();
     built++;
   }
--- a/ddprof-lib/src/main/cpp/javaApi.cpp
+++ b/ddprof-lib/src/main/cpp/javaApi.cpp
@@ -438,7 +438,8 @@ Java_com_datadoghq_profiler_OTelContext_setProcessCtx0(JNIEnv *env,
                                                          jstring runtime_id,
                                                          jstring service,
                                                          jstring version,
-                                                         jstring tracer_version
+                                                         jstring tracer_version,
+                                                         jobjectArray attribute_keys
                                                         ) {
   JniString env_str(env, env_data);
   JniString hostname_str(env, hostname);
@@ -448,6 +449,39 @@ Java_com_datadoghq_profiler_OTelContext_setProcessCtx0(JNIEnv *env,
   JniString tracer_version_str(env, tracer_version);
 
   const char *host_name_attrs[] = {"host.name", hostname_str.c_str(), NULL};
+
+  // Build the thread context attribute_key_map published alongside the process
+  // context: index 0 is the reserved datadog.local_root_span_id slot, followed by
+  // the caller-provided keys (clipped to DD_TAGS_CAPACITY)
+  int count = (attribute_keys != nullptr) ? env->GetArrayLength(attribute_keys) : 0;
+  int n = count < (int)DD_TAGS_CAPACITY ? count : (int)DD_TAGS_CAPACITY;
+  if (count > n) {
+    LOG_WARN("setProcessContext: %d attribute keys requested but capacity is %d; extra keys will be ignored",
+             count, (int)DD_TAGS_CAPACITY);
+  }
+
+  const char *key_ptrs[DD_TAGS_CAPACITY + 2]; // +1 reserved slot, +1 NULL terminator
+  JniString *jni_keys[DD_TAGS_CAPACITY];
+  int built = 0;
+  key_ptrs[0] = "datadog.local_root_span_id";
+  for (int i = 0; i < n; i++) {
+    jstring jstr = (jstring)env->GetObjectArrayElement(attribute_keys, i);
+    if (jstr == nullptr) {
+      // A null key would corrupt the index mapping; abort the publish.
+      for (int j = 0; j < built; j++) delete jni_keys[j];
+      Log::warn("setProcessContext: null attribute key at index %d; skipping publish", i);
+      return;
+    }
+    jni_keys[built] = new JniString(env, jstr);
+    key_ptrs[i + 1] = jni_keys[built]->c_str();
+    built++;
+  }
+  key_ptrs[n + 1] = nullptr;
+
+  otel_thread_ctx_config_data thread_ctx_config = {
+    .schema_version = "tlsdesc_v1_dev",
+    .attribute_key_map = key_ptrs,
+  };
 
   otel_process_ctx_data data = {
     .deployment_environment_name = env_str.c_str(),
@@ -459,13 +493,15 @@ Java_com_datadoghq_profiler_OTelContext_setProcessCtx0(JNIEnv *env,
     .telemetry_sdk_name = "dd-trace-java",
     .resource_attributes = host_name_attrs,
     .extra_attributes = NULL,
-    .thread_ctx_config = NULL  // Set later by ContextApi::registerAttributeKeys() when keys are known
+    .thread_ctx_config = &thread_ctx_config
   };
 
   otel_process_ctx_result result = otel_process_ctx_publish(&data);
   if (!result.success) {
     Log::warn("Failed to publish process context: %s", result.error_message);
   }
+
+  for (int i = 0; i < built; i++) delete jni_keys[i];
 }
 
 extern "C" DLLEXPORT jobject JNICALL
@@ -597,35 +633,6 @@ Java_com_datadoghq_profiler_ThreadContext_registerConstant0(JNIEnv* env, jclass 
   u32 encoding = Profiler::instance()->contextValueMap()->bounded_lookup(
       value_str.c_str(), value_str.length(), 1 << 16);
   return encoding == 0 ? -1 : (jint)encoding;
-}
-
-extern "C" DLLEXPORT void JNICALL
-Java_com_datadoghq_profiler_OTelContext_registerAttributeKeys0(JNIEnv* env, jclass unused, jobjectArray keys) {
-  int count = (keys != nullptr) ? env->GetArrayLength(keys) : 0;
-  int n = count < (int)DD_TAGS_CAPACITY ? count : (int)DD_TAGS_CAPACITY;
-  if (count > n) {
-    LOG_WARN("registerAttributeKeys: %d keys requested but capacity is %d; extra keys will be ignored",
-             count, (int)DD_TAGS_CAPACITY);
-  }
-
-  const char* key_ptrs[DD_TAGS_CAPACITY];
-  JniString* jni_strings[DD_TAGS_CAPACITY];
-
-  for (int i = 0; i < n; i++) {
-    jstring jstr = (jstring)env->GetObjectArrayElement(keys, i);
-    if (jstr == nullptr) {
-      for (int j = 0; j < i; j++) delete jni_strings[j];
-      return;
-    }
-    jni_strings[i] = new JniString(env, jstr);
-    key_ptrs[i] = jni_strings[i]->c_str();
-  }
-
-  // Always call registerAttributeKeys even with n==0 so the reserved
-  // datadog.local_root_span_id key (index 0) is published in the process context.
-  ContextApi::registerAttributeKeys(key_ptrs, n);
-
-  for (int i = 0; i < n; i++) delete jni_strings[i];
 }
 
 // ---- test and debug utilities
--- a/ddprof-lib/src/main/cpp/javaApi.cpp
+++ b/ddprof-lib/src/main/cpp/javaApi.cpp
@@ -456,8 +456,8 @@ Java_com_datadoghq_profiler_OTelContext_setProcessCtx0(JNIEnv *env,
   int count = (attribute_keys != nullptr) ? env->GetArrayLength(attribute_keys) : 0;
   int n = count < (int)DD_TAGS_CAPACITY ? count : (int)DD_TAGS_CAPACITY;
   if (count > n) {
-    LOG_WARN("setProcessContext: %d attribute keys requested but capacity is %d; extra keys will be ignored",
-             count, (int)DD_TAGS_CAPACITY);
+    Log::warn("setProcessContext: %d attribute keys requested but capacity is %d; extra keys will be ignored",
+              count, (int)DD_TAGS_CAPACITY);
   }
 
   const char *key_ptrs[DD_TAGS_CAPACITY + 2]; // +1 reserved slot, +1 NULL terminator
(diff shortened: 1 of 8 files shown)


ddprof-lib/src/main/java/com/datadoghq/profiler/OTelContext.java
```diff
@@ -215,58 +216,30 @@ public ProcessContext readProcessContext() {
...
+     *                      reserved datadog.local_root_span_id slot. Must not be null
+     *                      (may be empty); keys beyond capacity are clipped. Order must
+     *                      match the indices used with
+     *                      {@link ThreadContext#setContextAttribute(int, String)}.
+     *
      * @see <a href="https://opentelemetry.io/docs/specs/semconv/registry/attributes/service/">OpenTelemetry Service Attributes</a>
      * @see <a href="https://opentelemetry.io/docs/specs/semconv/registry/attributes/deployment/">OpenTelemetry Deployment Attributes</a>
      */
-    public void setProcessContext(String env, String hostname, String runtimeId, String service, String version, String tracerVersion) {
+    public void setProcessContext(String env, String hostname, String runtimeId, String service, String version, String tracerVersion, String[] attributeKeys) {
```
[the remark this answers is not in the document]
ivoanjo: This is not a really public API we want to maintain -- it's only consumer is dd-trace-java which will be updated in https://github.com/DataDog/dd-trace-java/pull/11558

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -1355,9 +1354,6 @@ Error Profiler::start(Arguments &args, bool reset) {
   JfrMetadata::reset();
   JfrMetadata::initialize(args._context_attributes);
   _num_context_attributes = args._context_attributes.size();
-  // Initialize the OTel thread context so external profilers can decode
-  // the per-thread context, including custom attributes
-  ContextApi::registerAttributeKeys(args._context_attributes);
   error = _jfr.start(args, reset);
```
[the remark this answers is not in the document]
ivoanjo: The only consumer of this code is dd-trace-java and we control it so that none of these theoretical things happen in practice, so no need for extra boilerplate to maintain non-existing just-in-cases.

jbachorik (commented):
Sphinx automated review — PR #576 · *ivoanjo/prof-14883-early-thread-context*

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -1355,9 +1354,6 @@ Error Profiler::start(Arguments &args, bool reset) {
   JfrMetadata::reset();
   JfrMetadata::initialize(args._context_attributes);
   _num_context_attributes = args._context_attributes.size();
```
jbachorik, line 1358: 🟡 **MEDIUM · completeness** `[CONSENSUS]`

Removing `ContextApi::registerAttributeKeys(args._context_attributes)` from this location means starting the profiler with `attributes=...` no longer publishes the OTEP `attribute_key_map` to the process context. The auto-publish path (added in commit `5c28a04d4`) is gone; only `setProcessContext(..., attributeKeys)` now publishes the key map. External consumers that discover the OTEL mapping but find no `thread_ctx_config` cannot map `key_index` bytes back to attribute names.

**Suggestion:** Confirm the consolidation is intentional and ensure the production caller always invokes `setProcessContext` with `attributeKeys` matching `args._context_attributes` (same order). If the start-only path must remain decodable without an explicit `setProcessContext` call, retain a publish step here.

_Also identified by the specialist reviewer (lens: `otel-context-registration-missing`)._

ddprof-test/src/test/java/com/datadoghq/profiler/context/ProcessContextTest.java
```diff
@@ -93,91 +92,24 @@ public void testNativeReadBackFunctionality() {
         String tracerVersion = "3.5.0";
```
jbachorik, line 99: 🟡 **MEDIUM · completeness** `[SPECIALIST]`

`testStartAttributesAutoRegistersKeys` — the only test asserting that `start,attributes=...` auto-publishes the OTEP `attribute_key_map` without a separate `setProcessContext` call — is deleted with no replacement. The behaviour change in `profiler.cpp` is now untested; a regression would go undetected by the surviving test suite.

**Suggestion:** Add a test asserting the intended new behaviour of the `attributes=...` start path — either explicitly verifying the key map is absent (if the two-step contract is now required), or providing an equivalent end-to-end test that covers the same scenario.

_Matched lens: `Tests missing or dropped for new code`._

ddprof-lib/src/main/java/com/datadoghq/profiler/OTelContext.java
```diff
@@ -215,58 +216,30 @@ public ProcessContext readProcessContext() {
      * @param tracerVersion the version of the tracer as defined by OpenTelemetry
      *                      semantic conventions (telemetry.sdk.version). Must not be null.
      *                      Examples: "3.5.0", "4.2.0"
-     *     *
+     * @param attributeKeys the thread-context attribute key names whose per-thread
+     *                      values are recorded in the OTEP thread-local record (e.g.
+     *                      "http.route", "db.system"). Published in the process context's
+     *                      thread_ctx_config as the attribute_key_map, preceded by the
+     *                      reserved datadog.local_root_span_id slot. Must not be null
```
jbachorik, line 233: 🔵 **LOW · consistency** `[CONSENSUS]`

The Javadoc states `attributeKeys` "Must not be null" but neither the Java wrapper nor the JNI layer enforces this contract. The JNI code silently treats `null` as an empty array (publishes only the reserved `datadog.local_root_span_id` slot with no warning or exception).

**Suggestion:** Either relax the Javadoc to document that `null` is treated as an empty key list (i.e. it is permitted), or add `Objects.requireNonNull(attributeKeys, "attributeKeys")` before the native call to enforce the stated contract.

_Also identified by the specialist reviewer._

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -449,6 +450,39 @@ Java_com_datadoghq_profiler_OTelContext_setProcessCtx0(JNIEnv *env,
...
+             count, (int)DD_TAGS_CAPACITY);
+  }
+
+  const char *key_ptrs[DD_TAGS_CAPACITY + 2]; // +1 reserved slot, +1 NULL terminator
+  JniString *jni_keys[DD_TAGS_CAPACITY];
+  int built = 0;
+  key_ptrs[0] = "datadog.local_root_span_id";
+  for (int i = 0; i < n; i++) {
+    jstring jstr = (jstring)env->GetObjectArrayElement(attribute_keys, i);
+    if (jstr == nullptr) {
```
jbachorik, line 469: 🔵 **LOW · robustness** `[GENERALIST]`

A single `null` element in `attributeKeys` aborts the **entire** `setProcessContext` publish — the deployment environment, service name, runtime ID, and all other process-context fields are dropped, not just the bad key. Only a `Log::warn` is emitted. The Javadoc covers the case where the array itself is null but does not document that a null element aborts the whole publish.

**Suggestion:** Either skip `null` elements rather than aborting the entire publish (e.g. treat them as a no-op slot), or document the abort-on-null-element contract explicitly in `OTelContext.setProcessContext` and add a test asserting the behaviour.

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -449,6 +450,39 @@ Java_com_datadoghq_profiler_OTelContext_setProcessCtx0(JNIEnv *env,
...
+      // A null key would corrupt the index mapping; abort the publish.
+      for (int j = 0; j < built; j++) delete jni_keys[j];
+      Log::warn("setProcessContext: null attribute key at index %d; skipping publish", i);
```
jbachorik, line 472: 🔵 **LOW · consistency** `[SPECIALIST]`

Two different logging APIs are used for `WARN`-level messages within the same function: `LOG_WARN` macro (line 459, capacity clip) vs `Log::warn` direct call (here, null key). The inconsistency suggests a copy-paste from two different source conventions.

**Suggestion:** Use one convention throughout `setProcessCtx0` — check what the rest of `javaApi.cpp` uses and align to it.

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -449,6 +450,39 @@ Java_com_datadoghq_profiler_OTelContext_setProcessCtx0(JNIEnv *env,
...
+      return;
+    }
+    jni_keys[built] = new JniString(env, jstr);
```
jbachorik, line 475: 🔵 **LOW · robustness** `[SPECIALIST]`

If `GetStringUTFChars` fails on OOM inside the `JniString` constructor, `c_str()` returns `NULL`. That `NULL` is then written into `key_ptrs[i+1]` mid-array. The consumer loop in `otel_process_ctx_publish` iterates `attribute_key_map` until it finds a `NULL` sentinel and stops early, silently truncating the map and corrupting the `index → key` mapping for all keys after the failed slot. This is consistent with the existing codebase convention, but the array-mapping semantics make the impact more significant here than in the older single-string JNI sites.

**Suggestion:** After `new JniString(env, jstr)`, check `jni_keys[built]->c_str() != nullptr`; on failure, clean up and abort the publish (mirroring the null-element path at lines 469–473).

ddprof-test/src/test/java/com/datadoghq/profiler/context/ProcessContextTest.java
```diff
@@ -29,7 +28,7 @@ public void testProcessContextMappingCreation() throws IOException {
         String version = "1.0.0";
         String tracerVersion = "3.5.0";
 
-        OTelContext.getInstance().setProcessContext(env, hostname, runtimeId, service, version, tracerVersion);
+        OTelContext.getInstance().setProcessContext(env, hostname, runtimeId, service, version, tracerVersion, new String[0]);
```
jbachorik: 🔵 **LOW · completeness** `[SPECIALIST]`

`testProcessContextMappingCreation` calls `setProcessContext` with `new String[0]` (empty keys) but only asserts an OTEL mapping exists. It never reads back `attributeKeyMap` to verify it contains exactly `["datadog.local_root_span_id"]` — the expected published map when no user keys are supplied.

**Suggestion:** Add a `readProcessContext()` call and assert `assertArrayEquals(new String[]{"datadog.local_root_span_id"}, readContext.attributeKeyMap)` to pin the empty-array behaviour.

ddprof-test/src/test/java/com/datadoghq/profiler/context/ProcessContextTest.java
```diff
@@ -93,91 +92,24 @@ public void testNativeReadBackFunctionality() {
         String tracerVersion = "3.5.0";
 
         OTelContext context = OTelContext.getInstance();
-        context.setProcessContext(env, hostname, runtimeId, service, version, tracerVersion);
+        context.setProcessContext(env, hostname, runtimeId, service, version, tracerVersion,
```
jbachorik: 🔵 **LOW · completeness** `[GENERALIST]`

No test exercises the capacity-clip path (`count > DD_TAGS_CAPACITY = 10`). The `LOG_WARN` branch and truncation logic are covered with at most 2 keys in the current suite, so an off-by-one at the clip boundary would go undetected.

**Suggestion:** Add a test that passes ≥ 11 keys to `setProcessContext` and asserts the published `attributeKeyMap` is `{"datadog.local_root_span_id", key0, …, key9}` (exactly `DD_TAGS_CAPACITY` user keys, `DD_TAGS_CAPACITY + 1` total).

ddprof-lib/src/main/cpp/profiler.cpp
ivoanjo: This is by design -- https://github.com/DataDog/dd-trace-java/pull/11558 has the matching change and the assumption is that they are hand-in-hand

ddprof-test/src/test/java/com/datadoghq/profiler/context/ProcessContextTest.java
ivoanjo: I don't agree with the AI -- it's suggesting a negative test (Y doesn't happen when X) and I don't see the value of such a negative tests. (E.g. I also don't assert that "free nachos every friday" doesn't happen when X happens -- negative tests are weird)

jbachorik: This is related to https://github.com/DataDog/java-profiler/pull/576#discussion_r3372069220 - if the context attributes are no longer autpublished and one needs to call `setProcessContext` to actually publish them, then the test, as described here, is not needed.

But - since `setProcessContext` is now setting the OTEL process context as well as making sure the custom attribute names are actually published, I wonder if it should not be renamed to something better reflecting the span of operations it does.

ddprof-test/src/test/java/com/datadoghq/profiler/context/ProcessContextTest.java
ivoanjo: Fixed in cf8b977d3ab24b8100413c566f62b321106865d0

ddprof-test/src/test/java/com/datadoghq/profiler/context/ProcessContextTest.java
ivoanjo: Fixed in 9c6811f5cf32070c662593d76b660c085b14fe14

ddprof-lib/src/main/cpp/javaApi.cpp
ivoanjo: Fixed in 34835e94aaa5a5ee6ccffddc438fd1bf8483a192

ddprof-lib/src/main/cpp/javaApi.cpp
ivoanjo: Fixed in 696ab91de58e15202a5dfa21c9962a095ee43f4d to use `Log::warn` always

ddprof-lib/src/main/cpp/javaApi.cpp
ivoanjo: In 7cafcf5e59d7df2ea94e7ef38adfc2b0aeabc22a the "abort-on-null" was documented + with a test for it

ddprof-lib/src/main/java/com/datadoghq/profiler/OTelContext.java
ivoanjo: Fixed in 02ed55719d365b9f396d949ced9a60f27f0c1386

ddprof-test/src/test/java/com/datadoghq/profiler/context/ProcessContextTest.java
ivoanjo: Hmmm, that's a good point. Perhaps `setProcessAndThreadContext`? Any suggestions?

jbachorik: Maybe `initializeAllContext()` and add a small javadoc mention of the custom attributes setup being done as part of this? Also, that one *has* to call this method to make sure the custom attributes are setup properly.

ivoanjo: Renamed and added big comment in 79cc3cc2e995078e2d9438c7cc48fe6df785a09f; also updated the downstream https://github.com/DataDog/dd-trace-java/pull/11558 to match.

jbachorik (approved):

ivoanjo:
I did a trivial rebase on top of main + force push; to get over the "commits were not signed" above (I had authored this in a workspace, need to check what's up there with missing signatures)

PR #578, opened 2026-06-05
Potentially infinite loop and store trace in unowned slot in CallTraceHashTable::putWithExistingId()

zhengyu123:
**What does this PR do?**:
This PR fixes a corner case, that may result in infinite loop and/or store trace in unowned slot.

**Motivation**:
Improve stability.

**Additional Notes**:

**How to test the change?**:

- Regular CI tests
- New test cases for this corner cases.
 
**For Datadog employees**:
- [X] JIRA: [PROF-14915](https://datadoghq.atlassian.net/browse/PROF-14915)

[PROF-14915]: https://datadoghq.atlassian.net/browse/PROF-14915?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddprof-lib/src/main/cpp/callTraceHashTable.cpp
```diff
@@ -455,37 +455,36 @@ void CallTraceHashTable::putWithExistingId(CallTrace* source_trace, u64 weight)
...
-      } else {
-        // Allocation failure - clear the key we claimed
-        __atomic_store_n(&keys[slot], 0, __ATOMIC_RELEASE);
-      }
+    }
+    if (probe.hasNext()) {
+      slot = probe.next();
+    } else {
+      // No more slot, break out
       break;
```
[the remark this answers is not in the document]
zhengyu123: @jbachorik Any reason it does not expand table here?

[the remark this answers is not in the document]
jbachorik: An omission in the initial impl :(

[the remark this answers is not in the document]
zhengyu123: so, we should expand table here as well?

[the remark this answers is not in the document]
jbachorik: Yes. That would be good to do.

ddprof-lib/src/test/cpp/test_callTraceStorage.cpp
```diff
@@ -664,3 +664,144 @@ TEST_F(CallTraceStorageTest, UseAfterFreeInProcessTraces) {
...
+        });
+
+        auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(10);
+        while (!done.load() && std::chrono::steady_clock::now() < deadline) {
+            std::this_thread::sleep_for(std::chrono::milliseconds(50));
+        }
+
+        bool ok = done.load();
+        if (!ok) {
+            t.detach();
```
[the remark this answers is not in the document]
zhengyu123: Fixed:

- std::atomic<bool> completed{false} → auto completed = std::make_shared<std::atomic<bool>>(false) — moves the atomic off the test stack and onto the heap.
- The lambda captures completed by value, copying the shared_ptr. Now both the test body and the worker thread hold a reference count, so the atomic's lifetime extends to whichever of the two outlives the other.
- All accesses become completed->store(true) / completed->load() — semantically identical, just through a pointer.

jbachorik:

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
```diff
@@ -256,6 +256,21 @@ CallTrace *CallTraceHashTable::findCallTrace(LongHashTable *table, u64 hash) {
...
 }
 
+void CallTraceHashTable::expandTableIfNeeded(LongHashTable* table) {
+    u32 size = table->size();
+    u32 capacity = table->capacity();
+
+    // EXPANSION LOGIC: Check if 75% capacity reached after incrementing size
+    if (size >= capacity * 3 / 4) {
+      // Allocate new table with double capacity using LinearAllocator
+      LongHashTable* new_table = LongHashTable::allocate(table, capacity * 2, &_allocator);
```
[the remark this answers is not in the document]
zhengyu123: Recheck

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
```diff
@@ -274,6 +274,23 @@ CallTrace *CallTraceHashTable::findCallTrace(LongHashTable *table, u64 hash) {
...
+void CallTraceHashTable::expandTableIfNeeded(LongHashTable* table) {
+
+    u32 size = table->size();
+    u32 capacity = table->capacity();
+    // No need to expand
+    if (size < capacity * 3 / 4) {
+      return;
+    }
+
+    ExclusiveLockGuard locker(&_expansionLock);
```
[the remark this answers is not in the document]
zhengyu123: Okay, reverted.

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
```diff
@@ -274,6 +274,23 @@ CallTrace *CallTraceHashTable::findCallTrace(LongHashTable *table, u64 hash) {
...
+    if (table == _table) {
+      // Allocate new table with double capacity using LinearAllocator
+      LongHashTable* new_table = LongHashTable::allocate(table, capacity * 2, &_allocator);
+      __atomic_store_n(&_table, new_table, __ATOMIC_RELEASE);
```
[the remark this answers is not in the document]
zhengyu123: Reverted

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
```diff
@@ -274,6 +274,23 @@ CallTrace *CallTraceHashTable::findCallTrace(LongHashTable *table, u64 hash) {
   return nullptr;
 }
 
+void CallTraceHashTable::expandTableIfNeeded(LongHashTable* table) {
+
+    u32 size = table->size();
+    u32 capacity = table->capacity();
+
+    // EXPANSION LOGIC: Check if 75% capacity reached after incrementing size
+    if (size >= capacity * 3 / 4 &&
```
rkennke: I would put the load-factor as constant somewhere.

ddprof-lib/src/main/cpp/callTraceHashTable.h
```diff
@@ -8,6 +8,7 @@
 
 #include "arch.h"
 #include "linearAllocator.h"
+#include "spinLock.h"
```
rkennke: Not used anymore?

ddprof-lib/src/main/cpp/callTraceHashTable.h
```diff
@@ -53,11 +54,13 @@ struct CallTraceSample {
 // Forward declaration for circular dependency
 class CallTraceStorage;
 
-class CallTraceHashTable {
+class alignas(alignof(SpinLock)) CallTraceHashTable {
```
rkennke: Also not used?

ddprof-lib/src/main/cpp/callTraceHashTable.h
```diff
@@ -53,11 +54,13 @@ struct CallTraceSample {
...
 public:
   static CallTrace _overflow_trace;
 
 private:
+  SpinLock _expansionLock;
```
rkennke: Also not used?

ddprof-lib/src/test/cpp/test_callTraceStorage.cpp
```diff
@@ -665,6 +665,145 @@ TEST_F(CallTraceStorageTest, UseAfterFreeInProcessTraces) {
...
+ *
+ * Before the fix: when all INITIAL_CAPACITY (65536) slots were occupied the
+ * probe loop called probe.next() without checking probe.hasNext(), so the
+ * prime-probe step cycled forever through already-occupied slots.
+ *
+ * After the fix: the hasNext() guard breaks the loop and the call returns,
+ * silently dropping the trace that could not be inserted.
+ *
+ */
+TEST_F(CallTraceStorageTest, PutWithExistingIdNoInfiniteLoopWhenFull) {
```
rkennke, line 679: It's good to have that test, but ... it doesn't actually exercise what it says it exercises, because now you're also growing the table, which means the table never gets full. Except for an OOM scenario. Maybe add a way to not grow the table, only for testing purpuses?

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
rkennke: The new expansion trigger >= vs the old == leaves a window where bursty concurrent put attempts would all allocate a new table, but only one wins, exposing us to potentially significant allocation waste. It's probably acceptable, but may be worth pointing out in a comment as a known fact.
(and conversely, leaving the old condition == in place leads to concurrent threads potentially pushing the size past the actual trigger value, thus potentially leading to no expansion at all, and therefore OOM situation, which is of course worse)

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
```diff
@@ -477,37 +485,39 @@ void CallTraceHashTable::putWithExistingId(CallTrace* source_trace, u64 weight)
...
       // Found empty slot - claim it atomically
       u64 expected = 0;
-      if (!__atomic_compare_exchange_n(&keys[slot], &expected, hash, false, __ATOMIC_ACQ_REL, __ATOMIC_RELAXED)) {
-        // Another thread claimed it, try next slot
-        if (probe.hasNext()) {
-          slot = probe.next();
-          continue;
+      if (__atomic_compare_exchange_n(&keys[slot], &expected, hash, false, __ATOMIC_ACQ_REL, __ATOMIC_RELAXED)) {
+        // Successfully claimed the slot
+
```
rkennke: trailing whitespace

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
```diff
@@ -477,37 +485,39 @@ void CallTraceHashTable::putWithExistingId(CallTrace* source_trace, u64 weight)
...
+          memcpy(copied_trace->frames, source_trace->frames, source_trace->num_frames * sizeof(ASGCT_CallFrame));
+          table->values()[slot].setTrace(copied_trace);
+          Counters::increment(CALLTRACE_STORAGE_BYTES, total_size);
+          Counters::increment(CALLTRACE_STORAGE_TRACES);
+
+          // Increment table size
+          table->incSize();
+
+          expandTableIfNeeded(table);
+
```
rkennke: whitespace

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
```diff
@@ -274,6 +274,23 @@ CallTrace *CallTraceHashTable::findCallTrace(LongHashTable *table, u64 hash) {
   return nullptr;
 }
 
+void CallTraceHashTable::expandTableIfNeeded(LongHashTable* table) {
+
+    u32 size = table->size();
```
rkennke: the indentation seems way off. all over the place.

rkennke (changes requested):
This is a nice fix(es)! Some minor-ish comments...

ddprof-lib/src/main/cpp/callTraceHashTable.h
zhengyu123: My bad, I did not revert `callTraceHashTable.h`

ddprof-lib/src/main/cpp/callTraceHashTable.h
zhengyu123: Reverted

ddprof-lib/src/test/cpp/test_callTraceStorage.cpp
zhengyu123: There is still a small chance, because it inserts before checking the load ratio. Also in OOM scenario.
Let's keep the test and figure out how to inject fault.

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
zhengyu123: Yes, that's why I explored locking. I think we can rollback the allocations, but I would prefer do it separately.

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
zhengyu123: Fixed

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
zhengyu123: Fixed

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
zhengyu123: Fixed

ddprof-lib/src/main/cpp/callTraceHashTable.h
zhengyu123: Noop, reverted

jbachorik (commented):
Sphinx review failed: 'str' object has no attribute 'get'

zhengyu123:
> Sphinx review failed: 'str' object has no attribute 'get'

@jbachorik Context please!

jbachorik:
> > Sphinx review failed: 'str' object has no attribute 'get'
> 
> @jbachorik Context please!

@zhengyu123 sorry, tool crashed locally and posted nonsense :/

jbachorik (commented):
Sphinx review failed: 'TimeoutExpired' object has no attribute 'process'

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
```diff
@@ -274,6 +274,21 @@ CallTrace *CallTraceHashTable::findCallTrace(LongHashTable *table, u64 hash) {
   return nullptr;
 }
 
+void CallTraceHashTable::expandTableIfNeeded(LongHashTable* table, u32 size) {
+  u32 capacity = table->capacity();
+
+  // EXPANSION LOGIC: Check if load ratio reached after incrementing size
+  if (size >= (u32) (capacity * LOAD_RATIO)  &&
```
jbachorik, line 281: This might be a premature optimization, but maybe we could precompute the threshold right when the new table is allocated with a given capacity and then save some floating point multiplications for each calltrace insertion.

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
```diff
@@ -477,37 +483,35 @@ void CallTraceHashTable::putWithExistingId(CallTrace* source_trace, u64 weight)
...
-        probe.updateCapacity(new_size);
-      } else {
-        // Allocation failure - clear the key we claimed
-        __atomic_store_n(&keys[slot], 0, __ATOMIC_RELEASE);
-      }
+    }
+    if (probe.hasNext()) {
+      slot = probe.next();
+    } else {
+      // No more slot, break out
```
jbachorik: Nit: Typo 'No more slot' -> 'No more slots'

ddprof-lib/src/test/cpp/test_callTraceStorage.cpp
```diff
@@ -4,7 +4,7 @@
  */
 
 #include "gtest/gtest.h"
-#include "callTraceStorage.h"
+#include "../../main/cpp/callTraceStorage.h"
```
jbachorik: Is this relative include necessary? `callTraceStorage.h` should work just as fine

jbachorik (changes requested):
Some nits, but LGTM assuming they are addressed

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
zhengyu123: This is more about readability.

jbachorik: I mean, we could have `table->isFull()` which would return `size >= precomputed_threshold`
But this is just a nit, feel free to ignore.

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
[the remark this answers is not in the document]
rkennke: So ... is this implemented now? Otherwise we should at the very least log it so that we don't silently drop traces?

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
zhengyu123: Fixed

ddprof-lib/src/test/cpp/test_callTraceStorage.cpp
zhengyu123: Yeap, fixed.

jbachorik (approved):
LGTM!

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
```diff
@@ -477,37 +483,35 @@ void CallTraceHashTable::putWithExistingId(CallTrace* source_trace, u64 weight)
...
-        probe.updateCapacity(new_size);
-      } else {
-        // Allocation failure - clear the key we claimed
-        __atomic_store_n(&keys[slot], 0, __ATOMIC_RELEASE);
-      }
+    }
+    if (probe.hasNext()) {
+      slot = probe.next();
+    } else {
+      // No more slots, break out
```
rkennke: The issue that has been mentioned before remains: When the scratch/standby table fills during liveness preservation, the preserved trace is dropped with no overflow sentinel and no log line. I recommend to at least log it. Or do you think this is not feasible/useful?

rkennke (changes requested):
Still one remaining issue, otherwise looks good.

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
zhengyu123: I see. Added log - @jbachorik could you confirm if it is correct?

rkennke (approved):
Thanks! Looks good now.

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
```diff
@@ -274,6 +274,21 @@ CallTrace *CallTraceHashTable::findCallTrace(LongHashTable *table, u64 hash) {
...
+  u32 capacity = table->capacity();
+
+  // EXPANSION LOGIC: Check if load ratio reached after incrementing size
+  if (size >= (u32) (capacity * LOAD_RATIO)  &&
+      table == __atomic_load_n(&_table, __ATOMIC_RELAXED)) { // quick check, if other thread already expanded the table
+    // Allocate new table with double capacity using LinearAllocator
+    LongHashTable* new_table = LongHashTable::allocate(table, capacity * 2, &_allocator);
+    if (new_table != nullptr) {
+      // Atomic table swap - only one thread succeeds
+      __atomic_compare_exchange_n(&_table, &table, new_table, false, __ATOMIC_ACQ_REL, __ATOMIC_RELAXED);
```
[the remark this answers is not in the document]
zhengyu123: This is a pre-exist bug. I believe the allocation by CAS victim, can be rollback, but would like to address it in separate PR.

ddprof-lib/src/test/cpp/test_callTraceStorage.cpp
```diff
@@ -665,6 +665,145 @@ TEST_F(CallTraceStorageTest, UseAfterFreeInProcessTraces) {
...
+        });
+
+        auto deadline = std::chrono::steady_clock::now() + std::chrono::seconds(10);
+        while (!done.load() && std::chrono::steady_clock::now() < deadline) {
+            std::this_thread::sleep_for(std::chrono::milliseconds(50));
+        }
+
+        bool ok = done.load();
+        if (!ok) {
+            t.detach();
```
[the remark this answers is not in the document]
zhengyu123: Fixed

PR #581, opened 2026-06-05
Add testProcess{Config} task to accept user options

zhengyu123:
**What does this PR do?**:
This PR adds a new task for testing, that accepts user options.

**Motivation**:
The task is created for testing new feature that avoids preloading all jmethod ids. By default, the feature is disabled and has to be enabled via `fjmethodid=false` argument, which is not possible with current `testdebug/release` tasks.

**Additional Notes**:

**How to test the change?**:
Run `./gradlew Pprofiler.options=fjmethodid=false testProcessDebug`, `fjmethodid=false` is merged into command line that print out by tests.

For example:
```
[TEST] Applied profiler.options: fjmethodid=false
===> keep_jfrs: false
com.datadoghq.profiler.wallclock.SmokeWallTest > test PASSED (789ms)
com.datadoghq.profiler.wallclock.SmokeWallTest > test STARTED
[TEST] Applied profiler.options: fjmethodid=false
===> command: start,wall=1ms,fjmethodid=false,cstack=vm,attributes=tag1;tag2;tag3,jfr,file=/tmp/recordings/fp/SmokeWallTest_test14542021396455448632.jfr
```

**For Datadog employees**:
- [X] JIRA: [PROF-14940](https://datadoghq.atlassian.net/browse/PROF-14940)

[PROF-14940]: https://datadoghq.atlassian.net/browse/PROF-14940?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbachorik (commented):
Sphinx automated review — PR #581 · *zgu/test_cmd_opts*

ddprof-test/src/test/java/com/datadoghq/profiler/AbstractProfilerTest.java
```diff
@@ -376,8 +377,64 @@ public final void registerCurrentThreadForWallClockProfiling() {
...
+    // Apply user-provided options from -Pprofiler.options (override test defaults)
+    String userOptions = System.getProperty("ddprof.test.options");
+    if (userOptions != null && !userOptions.isEmpty()) {
+      profilerCommand = mergeProfilerOptions(profilerCommand, userOptions);
+      System.out.println("[TEST] Applied profiler.options: " + userOptions);
+    }
+
     String testCstack = (String)testParams.get("cstack");
     if (testCstack != null) {
       profilerCommand += ",cstack=" + testCstack;
```
jbachorik, line 454: 🟡 **MEDIUM · correctness** `[CONSENSUS]` `[ADVERSARY-TESTED]`

The `testParams` cstack is appended **unconditionally** after `mergeProfilerOptions` has already inserted the user's `cstack` from `-Pprofiler.options`. This produces two `cstack=` entries in the final command; `arguments.cpp` parses left-to-right so the last one wins — meaning the test-parameter value silently defeats the user’s explicit override, directly contradicting the line 431 comment _“Apply user-provided options from -Pprofiler.options (override test defaults)”_.

**Suggestion:** Guard the append:
```java
if (testCstack != null && !profilerCommand.contains("cstack=")) {
    profilerCommand += ",cstack=" + testCstack;
}
```
Apply the same treatment to any other `testParams`-driven appends that run after `mergeProfilerOptions`.

_Also identified by the generalist reviewer (line 438). Adversary downgraded from HIGH to MEDIUM: test-only code, narrow trigger (cstack-parametrised test + manual `-Pprofiler.options cstack=...`). Matched lens: `Comment contradicts code behavior`._

ddprof-test/src/test/java/com/datadoghq/profiler/AbstractProfilerTest.java
```diff
@@ -376,8 +377,64 @@ public final void registerCurrentThreadForWallClockProfiling() {
...
 
+  /**
+   * Merges two profiler command strings. Options in overrides take precedence
+   * over options in base for matching keys.
+   *
+   * @param base The base profiler command (from test's getProfilerCommand())
+   * @param overrides The override options (from -Pprofiler.options)
+   * @return Merged command string with overrides taking precedence
+   */
+  private static String mergeProfilerOptions(String base, String overrides) {
```
jbachorik, line 387: 🟡 **MEDIUM · completeness** `[SPECIALIST]`

`mergeProfilerOptions` ships with **no unit tests** despite non-trivial logic: key deduplication, empty-value flag handling, override precedence, and command rebuilding. A unit test would have caught the cstack precedence inversion at line 440 before it reached review.

**Suggestion:** Add tests covering:
- Override wins for a duplicate key (`cpu=10ms` base + `cpu=100ms` override → `cpu=100ms`)
- Flag-only entry (no `=`) round-trips correctly
- `null`/empty overrides return base unchanged
- Cstack interaction: after `mergeProfilerOptions("start,cstack=vm", "cstack=dwarf")`, result contains `cstack=dwarf` exactly once

_Matched lens: `Tests missing or dropped for new code`._

ddprof-test/src/test/java/com/datadoghq/profiler/AbstractProfilerTest.java
```diff
@@ -376,8 +377,64 @@ public final void registerCurrentThreadForWallClockProfiling() {
...
+        options.put(part, "");
+      }
+    }
+
+    // Rebuild command
+    StringBuilder result = new StringBuilder();
+    for (Map.Entry<String, String> entry : options.entrySet()) {
+      if (result.length() > 0) {
+        result.append(",");
+      }
```
jbachorik, line 419: 🟡 **MEDIUM · correctness** `[SPECIALIST]`

`mergeProfilerOptions` cannot faithfully round-trip an **explicit empty-value option**: `key=` is parsed as an empty string value and re-emitted as bare `key` (the trailing `=` is silently dropped). Two consequences:
1. Option semantics change (`key=` and `key` may mean different things to the profiler).
2. The downstream guard `profilerCommand.contains("cstack=")` at line 441 fails to recognise a `cstack` that originated as `cstack=`, allowing an unexpected `,cstack=fp` append.

_Latent: no current `getProfilerCommand()` returns empty-value options._

**Suggestion:** Track whether the original token contained `=` and re-emit it even when the value is empty. Alternatively, document in the `mergeProfilerOptions` Javadoc that `key=` is treated identically to bare `key`.

ddprof-test/src/test/java/com/datadoghq/profiler/AbstractProfilerTest.java
```diff
@@ -376,8 +377,64 @@ public final void registerCurrentThreadForWallClockProfiling() {
...
+      int eq = part.indexOf('=');
+      if (eq > 0) {
+        options.put(part.substring(0, eq), part.substring(eq + 1));
+      } else if (!part.isEmpty()) {
+        options.put(part, "");
+      }
+    }
+
+    // Apply overrides
+    for (String part : overrides.split(",")) {
```
jbachorik, line 404: 🔵 **LOW · robustness** `[GENERALIST]`

A **valueless override token** (e.g. `-Pprofiler.options=cpu`) silently clears the value of a matching base option: `cpu=10ms` in the base becomes bare `cpu` in the merged result. Easy to trigger accidentally and undocumented.

**Suggestion:** Document this behavior in the `mergeProfilerOptions` Javadoc so users know that a bare key override strips any existing value. Alternatively, use `putIfAbsent` to avoid overwriting an existing valued key with an empty one:
```java
else if (!part.isEmpty()) {
    options.putIfAbsent(part, "");
}
```

ddprof-test/src/test/java/com/datadoghq/profiler/AbstractProfilerTest.java
zhengyu123: Done

ddprof-test/src/test/java/com/datadoghq/profiler/AbstractProfilerTest.java
zhengyu123: Do we really want to test of test?

ddprof-test/src/test/java/com/datadoghq/profiler/AbstractProfilerTest.java
zhengyu123: Fix: ignore key without value.

build-logic/conventions/src/main/kotlin/com/datadoghq/profiler/ProfilerTestPlugin.kt
```diff
@@ -356,6 +356,95 @@ class ProfilerTestPlugin : Plugin<Project> {
...
+                execTask.args = allArgs
+            }
+
+            // Environment variables
+            testConfig.environmentVariables.forEach { (key, value) ->
+                execTask.environment(key, value)
+            }
+
+            // Remove LD_LIBRARY_PATH to let RPATH work correctly
+            execTask.doFirst {
```
rkennke, line 501: Maybe it's cleaner do not add LD_LIBRARY_PATH in the loop just above?

rkennke (changes requested):
Good change. I found one small issue, and a bug, which I then found out that Codex already flagged it:
https://github.com/DataDog/java-profiler/pull/581/changes#r3375494440

ddprof-test/src/test/java/com/datadoghq/profiler/AbstractProfilerTest.java
zhengyu123: Fixed

zhengyu123:
Summary of what changed in `ddprof-test/src/test/java/com/datadoghq/profiler/AbstractProfilerTest.java`: the -Pprofiler.options override logic is now exposed as `public static String applyProfilerOptionOverrides(String)`, reused by g`etAmendedProfilerCommand()`. Then I wired it into every test that calls JavaProfiler.execute() with a hard-coded command instead of going through AbstractProfilerTest:

ShutdownTest — `runTest()` now builds "start," + applyProfilerOptionOverrides(options) + ",jfr,file=..."; call sites pass options only (no "start," prefix baked in).
OtelContextStorageModeTest — all 5 `execute()` calls now wrap their options portion.
NativeSocketRestartTest — the manual restart-session `execute()` call now wraps "natsock=100us".
StopWithChurningThreadsTest / StartWithChurningThreadsTest — PROFILER_CMD split into a `PROFILER_OPTIONS` constant plus an override-applied command built once before the loop.
GetLineNumberTableLeakTest, WriteStackTracesAfterClassUnloadTest, CleanupAfterClassUnloadTest (all extend AbstractProfilerTest via AbstractDynamicClassTest) — now call `applyProfilerOptionOverrides(getProfilerCommand())` instead of using `getProfilerCommand()`/a hard-coded literal directly.
One correctness detail I had to work around: the merge logic parses comma-separated `key=value` pairs, so wrapping the whole command (including the literal "start" action keyword or the trailing "jfr,file=..." segment) would corrupt them into "start=" / spurious keys. Every fix above applies the override only to the pure-options substring, matching the pattern AbstractProfilerTest itself already uses.

CarrierContextStorageTest and IdleThreadLifecycleTest needed no change — neither calls execute() with a profiler command. :ddprof-test:compileTestJava compiles cleanly with these changes (I also removed a handful of stale .o build artifacts left over from an upstream file rename that were unrelated to this fix but were blocking the native link step).

jbachorik:
@zhengyu123 ping - do you want to finish this or it's to be closed?

zhengyu123:
> @zhengyu123 ping - do you want to finish this or it's to be closed?

@jbachorik On my list, let me finish PR 639.

build-logic/conventions/src/main/kotlin/com/datadoghq/profiler/ProfilerTestPlugin.kt
zhengyu123: Could you check if it is still the case? I cannot find any loops.

rkennke: I believe that this loop (a few lines up):
```
            testConfig.environmentVariables.forEach { (key, value) ->
                execTask.environment(key, value)
            }

```

adds all environment variables to execTask. And here we are removing it again. You could just filter it in the add-loop, I think.

zhengyu123: Fixed

build-logic/conventions/src/main/kotlin/com/datadoghq/profiler/ProfilerTestPlugin.kt
```diff
@@ -399,20 +399,16 @@ class ProfilerTestPlugin : Plugin<Project> {
...
-                val currentLdLibPath = (execTask.environment["LD_LIBRARY_PATH"] as? String) ?: System.getenv("LD_LIBRARY_PATH")
-                if (!currentLdLibPath.isNullOrEmpty()) {
-                    project.logger.info("Removing LD_LIBRARY_PATH to prevent cross-JDK library conflicts (was: $currentLdLibPath)")
-                    execTask.environment.remove("LD_LIBRARY_PATH")
+            // Solution: Never propagate LD_LIBRARY_PATH so RPATH takes precedence
+            testConfig.environmentVariables.forEach { (key, value) ->
+                if (key == "LD_LIBRARY_PATH") {
+                    project.logger.info("Removing LD_LIBRARY_PATH to prevent cross-JDK library conflicts (was: $value)")
+                    return@forEach
                 }
```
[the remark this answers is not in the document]
rkennke: Right, that has been introduced by the change that I have requested. I think the previous version was actually more-correct. Sorry for that noise.

build-logic/conventions/src/main/kotlin/com/datadoghq/profiler/ProfilerTestPlugin.kt
```diff
@@ -429,6 +425,116 @@ class ProfilerTestPlugin : Plugin<Project> {
...
+                // Test filter from -Ptests property
+                val testsFilter = project.findProperty("tests") as String?
+                if (testsFilter != null) {
+                    allArgs.add("-Dtest.filter=$testsFilter")
+                }
+
+                // Profiler options from -Pprofiler.options property
+                val profilerOptions = project.findProperty("profiler.options") as String?
+                if (profilerOptions != null) {
+                    allArgs.add("-Dddprof.test.options=$profilerOptions")
```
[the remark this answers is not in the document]
rkennke: AFAICT, that is a true concern which should be addressed?

ddprof-test/src/test/java/com/datadoghq/profiler/AbstractProfilerTest.java
```diff
@@ -376,10 +376,81 @@ public final void registerCurrentThreadForWallClockProfiling() {
...
+
+    // Rebuild command, preserving keys with empty values (e.g. "filter=")
+    // so untouched base options round-trip exactly.
+    StringBuilder result = new StringBuilder();
+    for (Map.Entry<String, String> entry : options.entrySet()) {
+      if (result.length() > 0) {
+        result.append(",");
+      }
+      result.append(entry.getKey());
+      result.append("=").append(entry.getValue());
```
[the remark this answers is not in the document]
rkennke: And this also sounds like a valid concern that's not addressed?

PR #591, opened 2026-06-10
Fix crash at thread exit

zhengyu123:
**What does this PR do?**:
Fixed crashes during thread exit cleanup.

**Motivation**:
Improve stability.

**Additional Notes**:
On glibc, pthread_exit and pthread_cancel trigger _Unwind_ForcedUnwind from the dynamic libgcc_s.so.1. When the forced unwind walks a frame belonging to code built with -static-libgcc (both libjavaProfiler.so and the gtest binaries), it calls the static __gxx_personality_v0. The two libgcc versions have incompatible _Unwind_Context layouts. If the personality function finds a cleanup or handler action and calls _Unwind_SetGR, the cold/error path aborts.

Two patterns create LSDA actions that trigger _Unwind_SetGR:

- C++ RAII objects with destructors (struct Cleanup, __pthread_cleanup_class from pthread_cleanup_push in C++ mode, SigGuard)
- try/catch handler frames (catch (abi::__forced_unwind&))

**How to test the change?**:
Three failed tests now passed.
- forced_unwind_ut
- thread_teardown_safety_ut
- DynamicNativeThread

**For Datadog employees**:
- [X] JIRA: [PROF-15046](https://datadoghq.atlassian.net/browse/PROF-15046)

[PROF-15046]: https://datadoghq.atlassian.net/browse/PROF-15046?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
+++ b/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
@@ -151,6 +151,11 @@ void run_with_cleanup(func_start_routine routine, void* params,
         // should a future/variant glibc ever return from it.
         __builtin_unreachable();
     }
+    // Callers must not have a pending pthread_cancel when they enter
+    // run_with_cleanup: a cancellation arriving between __sigsetjmp returning
+    // and __pthread_register_cancel below would unwind this frame before
+    // cancel_buf is registered, silently skipping cleanup_fn.  All current
+    // callers are JVM/native start routines with no pending cancellation.
     __pthread_register_cancel(&cancel_buf);
     routine(params);
     __pthread_unregister_cancel(&cancel_buf);
@@ -344,7 +349,7 @@ static int pthread_create_hook_spec(pthread_t* thread,
 
 // Wrapper around the real start routine.
 // See comments for start_routine_wrapper_spec() for details
-__attribute__((visibility("hidden")))
+__attribute__((visibility("hidden"), no_stack_protector))
 static void* start_routine_wrapper(void* args) {
     RoutineInfo* thr = (RoutineInfo*)args;
     func_start_routine routine;
--- a/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
+++ b/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
@@ -15,6 +15,7 @@
 #include <dlfcn.h>
 #include <mutex>
 #include <limits.h>
+#include <setjmp.h>
 #include <string.h>
 #include <stdlib.h>
 
--- a/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
+++ b/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
@@ -205,6 +205,17 @@ int pthread_create_wrapped_for_test(pthread_t* thread,
     }
     return ret;
 }
+
+// Variant that passes the production cleanup_unregister as the cleanup function.
+// Exercises the full chain: start_routine_for_test → run_with_cleanup →
+// cleanup_unregister → Profiler::unregisterThread + ProfiledThread::release.
+// Profiler::unregisterThread is null-safe under UNIT_TEST (see profiler.cpp).
+int pthread_create_with_cleanup_unregister_for_test(pthread_t* thread,
+                                                    func_start_routine routine,
+                                                    void* params) {
+    return pthread_create_wrapped_for_test(thread, routine, params,
+                                           cleanup_unregister, nullptr);
+}
 #endif  // UNIT_TEST
 
 #ifdef __aarch64__
--- a/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
+++ b/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
@@ -73,6 +73,73 @@ static void unregister_and_release(int tid) {
     ProfiledThread::release();
 }
 
+// pthread_cleanup_push callback for thread wrappers.
+// Fires when the wrapped routine calls pthread_exit() or the thread is
+// canceled.  Kept noinline so its stack frame (which may hold a SignalBlocker
+// via unregister_and_release) lives outside the DEOPT-corruption zone of the
+// caller on musl/aarch64, and so that the SignalBlocker's sigset_t does not
+// appear in the caller's frame on platforms with stack-protector canaries.
+__attribute__((noinline))
+static void cleanup_unregister(void*) {
+    unregister_and_release(ProfiledThread::currentTid());
+}
+
+// Thread-cleanup wrapper that avoids the static-libgcc / forced-unwind crash.
+//
+// The crash: on glibc, pthread_cleanup_push in C++ mode expands to
+// __pthread_cleanup_class (RAII), which adds a cleanup entry to the LSDA of
+// this frame.  When libjavaProfiler.so is built with -static-libgcc, the
+// embedded __gxx_personality_v0 is called by the dynamic libgcc_s.so.1's
+// _Unwind_ForcedUnwind.  The two libgcc versions have incompatible
+// _Unwind_Context layouts; calling _Unwind_SetGR (which happens when the
+// personality finds a cleanup action) with a cross-version context triggers
+// the cold/error path, which calls abort().
+//
+// The fix: use __pthread_register_cancel / __pthread_unregister_cancel
+// directly — the same thing the C macro form of pthread_cleanup_push does.
+// This registers cleanup via a setjmp buffer in a runtime linked-list, NOT
+// via an LSDA destructor.  _Unwind_ForcedUnwind's stop function
+// (__pthread_unwind_stop) handles the cleanup without ever calling
+// __gxx_personality_v0 for this frame, so _Unwind_SetGR is never called and
+// the cross-version incompatibility is never triggered.
+//
+// On musl: pthread_cleanup_push already uses the C/setjmp form (no RAII),
+// and pthread_exit does not use _Unwind_ForcedUnwind, so there is no issue.
+// The __GLIBC__ guard keeps the musl path unchanged.
+#ifdef __GLIBC__
+// On glibc, <pthread.h> declares __pthread_register_cancel etc. only inside
+// the C (non-C++) conditional, so they're invisible in C++ code.  Redeclare
+// them with extern "C" so we can call them directly without the header guard.
+extern "C" {
+    extern void __pthread_register_cancel(__pthread_unwind_buf_t*);
+    extern void __pthread_unregister_cancel(__pthread_unwind_buf_t*);
+    [[noreturn]] extern void __pthread_unwind_next(__pthread_unwind_buf_t*);
+}
+#endif
+
+__attribute__((noinline, no_stack_protector))
+static void run_with_cleanup(func_start_routine routine, void* params) {
+#ifdef __GLIBC__
+    __pthread_unwind_buf_t cancel_buf;
+    if (__builtin_expect(
+            __sigsetjmp((struct __jmp_buf_tag*)(void*)cancel_buf.__cancel_jmp_buf, 0), 0)) {
+        // Reached via longjmp from glibc's stop function when pthread_exit
+        // (or cancellation) fires.  Run cleanup and continue unwinding.
+        cleanup_unregister(nullptr);
+        __pthread_unwind_next(&cancel_buf);
+    }
+    __pthread_register_cancel(&cancel_buf);
+    routine(params);
+    __pthread_unregister_cancel(&cancel_buf);
+    cleanup_unregister(nullptr);
+#else
+    // musl / non-glibc: pthread_cleanup_push uses the C/setjmp form, no RAII.
+    pthread_cleanup_push(cleanup_unregister, nullptr);
+    routine(params);
+    pthread_cleanup_pop(1);
+#endif
+}
+
 #ifdef __aarch64__
 // Delete RoutineInfo with profiling signals blocked to prevent ASAN
 // allocator lock reentrancy. Kept noinline so SignalBlocker's sigset_t
@@ -97,29 +164,6 @@ static void init_tls_and_register() {
         pt->startInitWindow();
     }
     Profiler::registerThread(ProfiledThread::currentTid());
-}
-
-// pthread_cleanup_push callback for start_routine_wrapper_spec.
-// Fires when the wrapped routine calls pthread_exit() or the thread is
-// canceled.  Kept noinline so its stack frame (which may hold a SignalBlocker
-// via unregister_and_release) lives outside the DEOPT-corruption zone of
-// start_routine_wrapper_spec.
-__attribute__((noinline))
-static void cleanup_unregister(void*) {
-    unregister_and_release(ProfiledThread::currentTid());
-}
-
-// pthread_cleanup_push declares `struct __ptcb` in the caller's frame.  If that
-// frame is start_routine_wrapper_spec, the structure sits inside the ~224-byte
-// DEOPT-corruption zone and pthread_cleanup_pop(1) would invoke a clobbered
-// function pointer.  This noinline + no_stack_protector helper hoists the
-// cleanup-handler frame out of the corruption zone — its own frame lives
-// safely above start_routine_wrapper_spec's.
-__attribute__((noinline, no_stack_protector))
-static void run_with_musl_cleanup(func_start_routine routine, void* params) {
-    pthread_cleanup_push(cleanup_unregister, nullptr);
-    routine(params);
-    pthread_cleanup_pop(1);
 }
 
 // Wrapper around the real start routine.
@@ -172,9 +216,9 @@ static void* start_routine_wrapper_spec(void* args) {
     delete_routine_info(thr);
     init_tls_and_register();
     // cleanup_unregister fires on pthread_exit() or cancellation from within
-    // routine(params).  The push/pop pair lives inside run_with_musl_cleanup so
+    // routine(params).  The push/pop pair lives inside run_with_cleanup so
     // that `struct __ptcb` does not land in this frame's DEOPT-corruption zone.
-    run_with_musl_cleanup(routine, params);
+    run_with_cleanup(routine, params);
     // pthread_exit instead of 'return': the saved LR in this frame is corrupted
     // by DEOPT PACKING; returning would jump to a garbage address.
     pthread_exit(nullptr);
@@ -227,13 +271,8 @@ static void* start_routine_wrapper(void* args) {
         ProfiledThread::currentSignalSafe()->startInitWindow();
         Profiler::registerThread(ProfiledThread::currentTid());
     }
-    // RAII cleanup: reads tid from TLS in the destructor (same rationale as
-    // start_routine_wrapper_spec: avoids storing state on a potentially corruptible frame).
-    // unregister_and_release() wraps the two calls under SignalBlocker (PROF-14603).
-    struct Cleanup {
-        ~Cleanup() { unregister_and_release(ProfiledThread::currentTid()); }
-    } cleanup;
-    routine(params);
+    // Use POSIX cleanup instead of C++ RAII to handle pthread_exit(): see run_with_cleanup.
+    run_with_cleanup(routine, params);
     return nullptr;
 }
 
--- a/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
+++ b/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
@@ -129,16 +129,27 @@ void run_with_cleanup(func_start_routine routine, void* params,
     static_assert(offsetof(__pthread_unwind_buf_t, __cancel_jmp_buf) == 0 &&
                   sizeof(cancel_buf.__cancel_jmp_buf[0]) == offsetof(struct __jmp_buf_tag, __saved_mask),
                   "glibc __pthread_unwind_buf_t inner layout incompatible with struct __jmp_buf_tag");
-    // Uses __sigsetjmp/longjmp which only intercepts _Unwind_ForcedUnwind, but not
-    // regular C++ exception from routine(params), which should be handled by JVM
+    // __sigsetjmp/longjmp only intercepts _Unwind_ForcedUnwind (pthread_exit /
+    // cancellation).  routine(params) must NOT throw a regular C++ exception
+    // across this boundary: an escaping exception would skip both
+    // __pthread_unregister_cancel and cleanup_fn below, leaking the thread
+    // registration and leaving cancel_buf linked against this (unwound) frame.
+    // We cannot defend with a try/catch here — a handler frame adds an LSDA
+    // action, which is exactly what triggers the static-libgcc abort this
+    // function exists to avoid.  Production routines are JVM/native start
+    // routines that handle their own exceptions and do not throw across here.
     if (__builtin_expect(
-            // set __sigsetjmp's savemask=0 (the second parameter, noting that the signal mask is NOT 
+            // set __sigsetjmp's savemask=0 (the second parameter, noting that the signal mask is NOT
             // saved/restored, which is correct because the cancel mechanism does not depend on signal mask state.
             __sigsetjmp((struct __jmp_buf_tag*)(void*)cancel_buf.__cancel_jmp_buf, 0), 0)) {
         // Reached via longjmp from glibc's stop function when pthread_exit
         // (or cancellation) fires.  Run cleanup and continue unwinding.
         cleanup_fn(cleanup_arg);
         __pthread_unwind_next(&cancel_buf);
+        // __pthread_unwind_next is [[noreturn]]; this fails loudly rather than
+        // falling through into __pthread_register_cancel on a torn-down frame
+        // should a future/variant glibc ever return from it.
+        __builtin_unreachable();
     }
     __pthread_register_cancel(&cancel_buf);
     routine(params);
@@ -300,9 +311,14 @@ static void* start_routine_wrapper_spec(void* args) {
     run_with_cleanup(routine, params, cleanup_unregister, nullptr);
     // pthread_exit instead of 'return': the saved LR in this frame is corrupted
     // by DEOPT PACKING; returning would jump to a garbage address.
-    // cleanup_unregister has already run via run_with_cleanup’s normal return path;
-    // TLS is cleared. pthread_exit triggers a second unwind with no registered cancel
-    // handler — safe because currentSignalSafe() returns null.
+    // cleanup_unregister has already run via run_with_cleanup's normal return
+    // path, so there is no registered cancel handler left.  The forced unwind
+    // raised by pthread_exit walks this frame, but it is safe because no
+    // destructor-bearing local (and hence no LSDA cleanup/handler action) is
+    // live at this call site: __gxx_personality_v0 returns continue-unwind
+    // without ever calling _Unwind_SetGR, avoiding the static-libgcc abort.
+    // WARNING: adding any RAII local with a destructor between run_with_cleanup
+    // and pthread_exit would reintroduce that crash.
     pthread_exit(nullptr);
     __builtin_unreachable();
 }
@@ -354,8 +370,12 @@ static void* start_routine_wrapper(void* args) {
         Profiler::registerThread(ProfiledThread::currentTid());
     }
     // Use POSIX cleanup instead of C++ RAII to handle pthread_exit(): see run_with_cleanup.
+    // cleanup_unregister has already run on run_with_cleanup's normal return path.
+    // The pthread_exit forced unwind is safe here for the same reason as in
+    // start_routine_wrapper_spec: no destructor-bearing local is live at this
+    // call site, so __gxx_personality_v0 never calls _Unwind_SetGR.
     run_with_cleanup(routine, params, cleanup_unregister, nullptr);
-    pthread_exit(nullptr); 
+    pthread_exit(nullptr);
     __builtin_unreachable();
 }
 
--- a/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
+++ b/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
@@ -126,9 +126,9 @@ void run_with_cleanup(func_start_routine routine, void* params,
     // Uses __sigsetjmp/longjmp which only intercepts _Unwind_ForcedUnwind, but not
     // regular C++ exception from routine(params), which should be handled by JVM
     if (__builtin_expect(
-            // set __sigsetjmp's savemask=1 (the second parameter) to match glibc's internal
-            //usage and ensure the signal mask is saved and restored across the setjmp/longjmp pair.
-            __sigsetjmp((struct __jmp_buf_tag*)(void*)cancel_buf.__cancel_jmp_buf, 0), 1)) {
+            // set __sigsetjmp's savemask=0 (the second parameter, noting that the signal mask is NOT 
+            // saved/restored, which is correct because the cancel mechanism does not depend on signal mask state.
+            __sigsetjmp((struct __jmp_buf_tag*)(void*)cancel_buf.__cancel_jmp_buf, 0), 0)) {
         // Reached via longjmp from glibc's stop function when pthread_exit
         // (or cancellation) fires.  Run cleanup and continue unwinding.
         cleanup_fn(cleanup_arg);
@@ -223,10 +223,14 @@ static void* start_routine_wrapper_spec(void* args) {
     init_tls_and_register();
     // cleanup_unregister fires on pthread_exit() or cancellation from within
     // routine(params).  The push/pop pair lives inside run_with_cleanup so
-    // that `struct __ptcb` does not land in this frame's DEOPT-corruption zone.
+    // that __pthread_unwind_buf_t (glibc) / struct __ptcb (musl) does not land
+    // in this frame's DEOPT-corruption zone.
     run_with_cleanup(routine, params, cleanup_unregister, nullptr);
     // pthread_exit instead of 'return': the saved LR in this frame is corrupted
     // by DEOPT PACKING; returning would jump to a garbage address.
+    // cleanup_unregister has already run via run_with_cleanup’s normal return path;
+    // TLS is cleared. pthread_exit triggers a second unwind with no registered cancel
+    // handler — safe because currentSignalSafe() returns null.
     pthread_exit(nullptr);
     __builtin_unreachable();
 }
--- a/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
+++ b/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
@@ -118,24 +118,29 @@ extern "C" {
 }
 #endif
 
-__attribute__((noinline, no_stack_protector))
-static void run_with_cleanup(func_start_routine routine, void* params) {
+__attribute__((visibility("hidden"), noinline, no_stack_protector))
+void run_with_cleanup(func_start_routine routine, void* params,
+                      void (*cleanup_fn)(void*), void* cleanup_arg) {
 #ifdef __GLIBC__
-    __pthread_unwind_buf_t cancel_buf;
+    __pthread_unwind_buf_t cancel_buf = {};
+    // Uses __sigsetjmp/longjmp which only intercepts _Unwind_ForcedUnwind, but not
+    // regular C++ exception from routine(params), which should be handled by JVM
     if (__builtin_expect(
-            __sigsetjmp((struct __jmp_buf_tag*)(void*)cancel_buf.__cancel_jmp_buf, 0), 0)) {
+            // set __sigsetjmp's savemask=1 (the second parameter) to match glibc's internal
+            //usage and ensure the signal mask is saved and restored across the setjmp/longjmp pair.
+            __sigsetjmp((struct __jmp_buf_tag*)(void*)cancel_buf.__cancel_jmp_buf, 0), 1)) {
         // Reached via longjmp from glibc's stop function when pthread_exit
         // (or cancellation) fires.  Run cleanup and continue unwinding.
-        cleanup_unregister(nullptr);
+        cleanup_fn(cleanup_arg);
         __pthread_unwind_next(&cancel_buf);
     }
     __pthread_register_cancel(&cancel_buf);
     routine(params);
     __pthread_unregister_cancel(&cancel_buf);
-    cleanup_unregister(nullptr);
+    cleanup_fn(cleanup_arg);
 #else
     // musl / non-glibc: pthread_cleanup_push uses the C/setjmp form, no RAII.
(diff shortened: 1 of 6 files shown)


ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
```diff
@@ -73,6 +73,73 @@ static void unregister_and_release(int tid) {
...
+    [[noreturn]] extern void __pthread_unwind_next(__pthread_unwind_buf_t*);
+}
+#endif
+
+__attribute__((noinline, no_stack_protector))
+static void run_with_cleanup(func_start_routine routine, void* params) {
+#ifdef __GLIBC__
+    __pthread_unwind_buf_t cancel_buf;
+    if (__builtin_expect(
+            __sigsetjmp((struct __jmp_buf_tag*)(void*)cancel_buf.__cancel_jmp_buf, 0), 0)) {
```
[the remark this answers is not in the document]
zhengyu123: Fixed

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
```diff
@@ -73,6 +74,73 @@ static void unregister_and_release(int tid) {
...
+// them with extern "C" so we can call them directly without the header guard.
+extern "C" {
+    extern void __pthread_register_cancel(__pthread_unwind_buf_t*);
+    extern void __pthread_unregister_cancel(__pthread_unwind_buf_t*);
+    [[noreturn]] extern void __pthread_unwind_next(__pthread_unwind_buf_t*);
+}
+#endif
+
+__attribute__((noinline, no_stack_protector))
+static void run_with_cleanup(func_start_routine routine, void* params) {
```
jbachorik: **[Sphinx Review — MEDIUM]** run_with_cleanup (glibc path) uses __sigsetjmp/longjmp which only intercepts _Unwind_ForcedUnwind. If routine(params) throws a regular C++ exception, __pthread_unregister_cancel is never called, leaving cancel_buf registered with a longjmp target pointing at a destroyed stack frame. The old RAII Cleanup struct handled all unwind types.

**Suggestion:** Add a `static_assert` / comment documenting the architectural invariant that JVM start routines (func_start_routine = void*(*)(void*)) cannot throw C++ exceptions.

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
```diff
@@ -73,6 +74,73 @@ static void unregister_and_release(int tid) {
...
+#ifdef __GLIBC__
+    __pthread_unwind_buf_t cancel_buf;
```
jbachorik: **[Sphinx Review — MEDIUM]** `__pthread_unwind_buf_t cancel_buf` is declared without zero-initialisation in `run_with_cleanup` (and both test helpers at `forced_unwind_ut.cpp:229` and `thread_teardown_safety_ut.cpp:294`). Fields other than `__cancel_jmp_buf` (specifically `__priv.__pad`) contain stack garbage when passed to `__pthread_register_cancel`. While current glibc writes before reading, this is a latent UB source.

**Suggestion:** Zero-initialize in all three sites: `__pthread_unwind_buf_t cancel_buf = {};` — zero cost at runtime, eliminates latent UB.

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
```diff
@@ -73,6 +74,73 @@ static void unregister_and_release(int tid) {
...
+    if (__builtin_expect(
+            __sigsetjmp((struct __jmp_buf_tag*)(void*)cancel_buf.__cancel_jmp_buf, 0), 0)) {
```
jbachorik, line 144: **[Sphinx Review — MEDIUM]** `__sigsetjmp` is called with `savemask=0`. Glibc's internal pthread unwinder uses `savemask=1`. If a JVM routine modifies the signal mask before calling pthread_exit, the modified mask is not restored after the longjmp, potentially leaving profiling signals permanently blocked or unblocked for the lifetime of the thread's cleanup.

**Suggestion:** Change the second argument of `__sigsetjmp` from `0` to `1` to match glibc's internal usage and ensure the signal mask is saved and restored across the setjmp/longjmp pair.

jbachorik, line 144: **[Sphinx Review — MEDIUM]** C-style `(struct __jmp_buf_tag*)(void*)` cast to type-pun `__cancel_jmp_buf_tag[1]` (a struct with `__jmp_buf` + `int __mask_was_saved`) relies on undocumented glibc internal layout. No `static_assert` guards against ABI changes. The same pattern is repeated in three files.

**Suggestion:** Add a `static_assert` before each cast site checking `offsetof(__pthread_unwind_buf_t, __cancel_jmp_buf) == 0 && sizeof(cancel_buf.__cancel_jmp_buf[0].__cancel_jmp) == sizeof(struct __jmp_buf_tag)` to catch future glibc ABI breakage at compile time.

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
```diff
@@ -73,6 +74,73 @@ static void unregister_and_release(int tid) {
...
+    cleanup_unregister(nullptr);
+#else
+    // musl / non-glibc: pthread_cleanup_push uses the C/setjmp form, no RAII.
+    pthread_cleanup_push(cleanup_unregister, nullptr);
+    routine(params);
+    pthread_cleanup_pop(1);
+#endif
+}
+
 #ifdef __aarch64__
```
jbachorik, line 237: **[Sphinx Review — LOW]** No test in this PR exercises the production `run_with_cleanup` directly — all tests use `run_with_cancel_handler`. Removing `__pthread_register_cancel` from production `run_with_cleanup` would leave all glibc tests passing.

**Suggestion:** Add an integration-level test that exercises `start_routine_wrapper` or `start_routine_wrapper_spec` end-to-end (via the patched `pthread_create` hook) and verifies `ProfiledThread` is released when the wrapped thread calls `pthread_exit` or is canceled.

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
```diff
@@ -172,9 +217,9 @@ static void* start_routine_wrapper_spec(void* args) {
     delete_routine_info(thr);
     init_tls_and_register();
     // cleanup_unregister fires on pthread_exit() or cancellation from within
-    // routine(params).  The push/pop pair lives inside run_with_musl_cleanup so
+    // routine(params).  The push/pop pair lives inside run_with_cleanup so
     // that `struct __ptcb` does not land in this frame's DEOPT-corruption zone.
```
jbachorik: **[Sphinx Review — LOW]** Comment says 'struct __ptcb does not land in this frame's DEOPT-corruption zone' after the refactor. On glibc, `run_with_cleanup` now uses `__pthread_unwind_buf_t` (registered via `__pthread_register_cancel`), not `struct __ptcb`.

**Suggestion:** Update the comment: '__pthread_unwind_buf_t (glibc) / struct __ptcb (musl) does not land in this frame's DEOPT-corruption zone.'

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
```diff
@@ -172,9 +217,9 @@ static void* start_routine_wrapper_spec(void* args) {
...
-    run_with_musl_cleanup(routine, params);
+    run_with_cleanup(routine, params);
```
jbachorik: **[Sphinx Review — MEDIUM]** In `start_routine_wrapper_spec`, when `routine` returns normally, `run_with_cleanup` calls `cleanup_unregister` (which calls `ProfiledThread::release()` and clears TLS) before returning. Then `pthread_exit(nullptr)` triggers a second forced-unwind through this frame with no registered cancel handler. A profiling signal arriving during this second unwind finds a null `ProfiledThread` (safe by design), but the comment does not document this ordering, which could mislead maintainers.

**Suggestion:** Add a comment before `pthread_exit(nullptr)`: 'cleanup_unregister has already run via run_with_cleanup’s normal return path; TLS is cleared. pthread_exit triggers a second unwind with no registered cancel handler — safe because currentSignalSafe() returns null.'

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
```diff
@@ -73,6 +74,73 @@ static void unregister_and_release(int tid) {
...
+// directly — the same thing the C macro form of pthread_cleanup_push does.
+// This registers cleanup via a setjmp buffer in a runtime linked-list, NOT
+// via an LSDA destructor.  _Unwind_ForcedUnwind's stop function
+// (__pthread_unwind_stop) handles the cleanup without ever calling
+// __gxx_personality_v0 for this frame, so _Unwind_SetGR is never called and
+// the cross-version incompatibility is never triggered.
+//
+// On musl: pthread_cleanup_push already uses the C/setjmp form (no RAII),
+// and pthread_exit does not use _Unwind_ForcedUnwind, so there is no issue.
+// The __GLIBC__ guard keeps the musl path unchanged.
```
jbachorik, line 109: **[Sphinx Review — LOW]** No test verifies that `Profiler::unregisterThread` is called inside `cleanup_unregister` — tests only assert the callback ran via atomic bool. Replacing `unregister_and_release` with a bare `ProfiledThread::release()` (dropping the PROF-14603 invariant) would be undetected.

**Suggestion:** Add an assertion verifying the thread's TID is no longer registered in the profiler after the cleanup callback returns.

ddprof-lib/src/test/cpp/thread_teardown_safety_ut.cpp
```diff
@@ -255,27 +265,51 @@ static std::atomic<bool> g_t07_release_ran{false};
...
+//    from glibc's stop function, __pthread_unwind_next calls _Unwind_ForcedUnwind
+//    again to continue unwinding through this frame.  Any LSDA cleanup entry
+//    (e.g. SigGuard, std::unique_ptr) would make __gxx_personality_v0 call
+//    _Unwind_SetGR with a cross-version context → abort().
+//
+// 2. No try/catch: handler frames also add LSDA entries, same problem.
+//
+// Signal handler save/restore is done manually (plain struct sigaction, no
+// destructor) so there is nothing in the LSDA for this frame.
 static void *t07_body(void *) {
```
jbachorik: **[Sphinx Review — MEDIUM]** `t07_body` (glibc path) uses the `__pthread_register_cancel`/longjmp pattern but lacks `__attribute__((no_stack_protector))`. With `-fstack-protector-strong`, a canary is inserted because `cancel_buf`'s address is taken, risking SIGABRT from `__stack_chk_fail` after the longjmp. The function also re-implements the mechanism inline rather than calling `run_with_cleanup`, so a bug in `run_with_cleanup` would not be caught.

**Suggestion:** Add `__attribute__((noinline, no_stack_protector))` to the glibc variant of `t07_body`. Consider restructuring it to call `run_with_cleanup` to validate the actual production code path.

ddprof-lib/src/test/cpp/thread_teardown_safety_ut.cpp
```diff
@@ -255,27 +265,51 @@ static std::atomic<bool> g_t07_release_ran{false};
...
+    __pthread_unwind_next(&cancel_buf);
+  }
+  __pthread_register_cancel(&cancel_buf);
+  // Inject a signal before the cancellation point to exercise the combined path.
+  pthread_kill(pthread_self(), SIGVTALRM);
+  while (true) {
+    pthread_testcancel();
+    usleep(100);
   }
+  __pthread_unregister_cancel(&cancel_buf);
```
jbachorik: **[Sphinx Review — MEDIUM]** Lines 311–313 (`__pthread_unregister_cancel`, `sigaction` restore, `ProfiledThread::release()`) are unreachable dead code. The `while(true)` loop above only exits via `pthread_testcancel()` triggering cancellation, which longjmps to the setjmp handler block at line 295 — the normal-return path can never reach line 311.

**Suggestion:** Remove lines 311–313. The thread exits exclusively via the forced-unwind path through `__pthread_unwind_next`.

ddprof-lib/src/test/cpp/forced_unwind_ut.cpp
```diff
@@ -34,31 +47,64 @@
...
+extern "C" {
+    extern void __pthread_register_cancel(__pthread_unwind_buf_t*);
+    extern void __pthread_unregister_cancel(__pthread_unwind_buf_t*);
+    [[noreturn]] extern void __pthread_unwind_next(__pthread_unwind_buf_t*);
+}
+
+// Helper: run `fn(arg)` with a glibc C-form cleanup handler that calls
+// `cleanup(cleanup_arg)` if the thread is canceled or calls pthread_exit
+// before fn returns.  Mirrors run_with_cleanup in libraryPatcher_linux.cpp.
+static void run_with_cancel_handler(
```
jbachorik: **[Sphinx Review — MEDIUM]** `run_with_cancel_handler` is missing `__attribute__((noinline, no_stack_protector))` that the production `run_with_cleanup` carries. Without `noinline`, the compiler may merge the helper into the caller's frame. Without `no_stack_protector`, a stack canary is inserted when `cancel_buf`'s address is taken, risking SIGABRT via `__stack_chk_fail` on the longjmp return path. Additionally, all three glibc tests call this helper rather than the production `run_with_cleanup` — a bug specific to `run_with_cleanup` would leave all tests green.

**Suggestion:** Add `__attribute__((noinline, no_stack_protector))` to `run_with_cancel_handler`. Consider exposing `run_with_cleanup` via a test-only symbol so at least one test exercises the production function directly.

ddprof-lib/src/test/cpp/stress_callTraceStorage.cpp
```diff
@@ -2034,7 +2034,17 @@ TEST_F(StressTestSuite, HashTableSpinWaitEdgeCasesTest) {
 // correctness across expansion boundaries independently of memory ordering;
 // it catches logic regressions in all build configurations.
 TEST_F(StressTestSuite, FindCallTraceAtomicReadRaceTest) {
-#if !defined(__SANITIZE_THREAD__) && !(defined(__has_feature) && __has_feature(thread_sanitizer))
+#if defined(__SANITIZE_THREAD__)
+    #define USE_ASAN 1
+#else
```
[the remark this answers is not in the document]
zhengyu123: @datadog fix this

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: @datadog fix this

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: @datadog fix this

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: @datadog fix this

zhengyu123: @datadog fix this

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: @datadog fix this

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: @datadog fix this

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: @datadog fix this

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: @datadog fix this

ddprof-lib/src/test/cpp/thread_teardown_safety_ut.cpp
zhengyu123: @datadog fix this

ddprof-lib/src/test/cpp/stress_callTraceStorage.cpp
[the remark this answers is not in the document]
zhengyu123: Fixed

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: Fixed

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: Added comment

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: Fixed

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: Fixed

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: Fixed

ddprof-lib/src/test/cpp/thread_teardown_safety_ut.cpp
zhengyu123: Fixed

ddprof-lib/src/test/cpp/forced_unwind_ut.cpp
zhengyu123: No longer applies

ddprof-lib/src/test/cpp/stress_callTraceStorage.cpp
```diff
@@ -2034,7 +2034,19 @@ TEST_F(StressTestSuite, HashTableSpinWaitEdgeCasesTest) {
 // correctness across expansion boundaries independently of memory ordering;
 // it catches logic regressions in all build configurations.
 TEST_F(StressTestSuite, FindCallTraceAtomicReadRaceTest) {
-#if !defined(__SANITIZE_THREAD__) && !(defined(__has_feature) && __has_feature(thread_sanitizer))
+
+#define HAS_ASAN 0
```
jbachorik: **[Sphinx Review — MEDIUM]** Macro `HAS_ASAN` is named for AddressSanitizer but detects ThreadSanitizer: it guards on `__SANITIZE_THREAD__` and `__has_feature(thread_sanitizer)`, and the `GTEST_SKIP` message says "TSan-only race regression". Future developers will assume this is ASan-gated and may add conflicting guards.

**Suggestion:** Rename `HAS_ASAN` to `HAS_TSAN` everywhere in this test function to match what it actually detects.

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
```diff
@@ -73,6 +74,78 @@ static void unregister_and_release(int tid) {
...
+
+__attribute__((visibility("hidden"), noinline, no_stack_protector))
+void run_with_cleanup(func_start_routine routine, void* params,
+                      void (*cleanup_fn)(void*), void* cleanup_arg) {
+#ifdef __GLIBC__
+    __pthread_unwind_buf_t cancel_buf = {};
+    // Uses __sigsetjmp/longjmp which only intercepts _Unwind_ForcedUnwind, but not
+    // regular C++ exception from routine(params), which should be handled by JVM
+    if (__builtin_expect(
+            // set __sigsetjmp's savemask=1 (the second parameter) to match glibc's internal
```
jbachorik: **[Sphinx Review — MEDIUM]** The comment says "set `__sigsetjmp`'s savemask=1 to match glibc's internal usage and ensure the signal mask is saved and restored" but the actual call still passes `0`. This comment is wrong: `savemask=0` is correct — glibc's own C-form `pthread_cleanup_push` macro also uses `savemask=0` for `__pthread_unwind_buf_t` cleanup. The comment will mislead future maintainers into changing a working argument.

Note: this appears to be a partial application of [a previous review suggestion](https://github.com/DataDog/java-profiler/pull/591#discussion_r...) — the developer changed the `__builtin_expect` hint to `1` (see next comment) but left the `savemask` at `0`. Both the comment and the hint change should be reverted.

**Suggestion:** Remove this comment entirely, or replace with: `// savemask=0: signal mask save/restore is not needed here; __pthread_unwind_buf_t cleanup uses the longjmp path, not sigsetjmp mask semantics.`

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
```diff
@@ -73,6 +74,78 @@ static void unregister_and_release(int tid) {
...
+            //usage and ensure the signal mask is saved and restored across the setjmp/longjmp pair.
+            __sigsetjmp((struct __jmp_buf_tag*)(void*)cancel_buf.__cancel_jmp_buf, 0), 1)) {
```
jbachorik: **[Sphinx Review — LOW]** `__builtin_expect(__sigsetjmp(...), 1)` tells the compiler the cleanup if-body is the **hot** path. But `__sigsetjmp` returns `0` on the initial call (the normal path — hot) and non-zero only after `longjmp` (the rare teardown path). The hint `1` is backwards, pessimizing every thread's normal execution through `run_with_cleanup`.

This appears to be an accidental change: the previous review said to change `__sigsetjmp`'s second argument (`savemask`) from `0` to `1`, but the `__builtin_expect` hint was changed instead. The identical pattern in `t07_body` (line ~296) correctly uses `__builtin_expect(..., 0)`.

**Suggestion:** Change `__builtin_expect(..., 1)` back to `__builtin_expect(..., 0)` to match `t07_body`.

ddprof-lib/src/test/cpp/thread_teardown_safety_ut.cpp
```diff
@@ -255,27 +265,51 @@ static std::atomic<bool> g_t07_release_ran{false};
...
 
-  try {
-    // Inject a signal before the cancellation point to exercise the combined path.
-    pthread_kill(pthread_self(), SIGVTALRM);
-    while (true) {
-      pthread_testcancel();
-      usleep(100);
-    }
-  } catch (abi::__forced_unwind &) {
+  __pthread_unwind_buf_t cancel_buf;
```
jbachorik: **[Sphinx Review — MEDIUM]** `cancel_buf` in `t07_body` is declared without zero-initialization (`__pthread_unwind_buf_t cancel_buf;`), while the production `run_with_cleanup` uses `cancel_buf = {}`. The PR description explicitly calls out zero-initialization as a key change. The `__priv.__pad` fields of `__pthread_unwind_buf_t` passed to `__pthread_register_cancel` may contain stack garbage.

**Suggestion:** Change to `__pthread_unwind_buf_t cancel_buf = {};` to match the production pattern.

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
```diff
@@ -227,13 +277,8 @@ static void* start_routine_wrapper(void* args) {
...
-    // RAII cleanup: reads tid from TLS in the destructor (same rationale as
-    // start_routine_wrapper_spec: avoids storing state on a potentially corruptible frame).
-    // unregister_and_release() wraps the two calls under SignalBlocker (PROF-14603).
-    struct Cleanup {
-        ~Cleanup() { unregister_and_release(ProfiledThread::currentTid()); }
-    } cleanup;
-    routine(params);
+    // Use POSIX cleanup instead of C++ RAII to handle pthread_exit(): see run_with_cleanup.
+    run_with_cleanup(routine, params, cleanup_unregister, nullptr);
     return nullptr;
```
jbachorik: **[Sphinx Review — MEDIUM]** `start_routine_wrapper` calls `run_with_cleanup(...)` then `return nullptr`. The companion function `start_routine_wrapper_spec` explicitly documents: *"the saved LR in this frame is corrupted by DEOPT PACKING; returning would jump to a garbage address"* — and uses `pthread_exit(nullptr)`. The non-spec wrapper has the same risk on musl/aarch64 but uses `return nullptr` without explanation.

**Suggestion:** Replace `return nullptr` with `pthread_exit(nullptr); __builtin_unreachable();` (matching the spec variant), or add a comment explaining why DEOPT PACKING doesn't apply to this wrapper (e.g., it is only registered via a hook that is never used on musl/aarch64).

ddprof-lib/src/test/cpp/thread_teardown_safety_ut.cpp
jbachorik: **[Sphinx Review — MEDIUM]** `t07_body` (glibc path) uses `__pthread_unwind_buf_t cancel_buf` on the stack with its address passed to `__pthread_register_cancel`, yet lacks `__attribute__((no_stack_protector))` that `run_with_cleanup` carries for exactly this reason. On aarch64 with `-fstack-protector-strong`, the address-take triggers stack canary insertion; the canary can be corrupted by the longjmp and trigger `__stack_chk_fail`, causing the test to SIGABRT instead of validating the production fix.

**Suggestion:** Add `__attribute__((no_stack_protector))` to the glibc variant of `t07_body`, matching `run_with_cleanup`.

ddprof-lib/src/test/cpp/thread_teardown_safety_ut.cpp
jbachorik: **[Sphinx Review — MEDIUM]** Lines 311–314 are unreachable dead code. The `while(true)` loop exits only via `pthread_testcancel()` triggering cancellation, which longjmps to the branch at line 296 and then calls `__pthread_unwind_next` (`[[noreturn]]`). The dead section includes a **second `ProfiledThread::release()` call** that would be a double-release use-after-free if somehow reached.

**Suggestion:** Remove lines 311–314 entirely.

ddprof-lib/src/test/cpp/forced_unwind_ut.cpp
```diff
@@ -34,31 +47,42 @@
...
+static void* bare_cancel_thread(void*) {
+    g_bare_cleanup_called.store(false, std::memory_order_relaxed);
+    run_with_cleanup(bare_spin, nullptr, bare_cleanup, nullptr);
     return nullptr;
 }
 
-// Regression (glibc): RAII destructor fires on pthread_cancel.
-TEST(ForcedUnwindTest, RaiiDestructorRunsOnPthreadCancel) {
+// Regression (glibc): __pthread_register_cancel callback fires on pthread_cancel.
+TEST(ForcedUnwindTest, CleanupCallbackRunsOnPthreadCancel) {
```
jbachorik, line 85: **[Sphinx Review — LOW]** No test exercises the **normal-return path** of `run_with_cleanup` (where `routine` returns normally → `__pthread_unregister_cancel` + `cleanup_fn` are called at lines 79–80 of libraryPatcher_linux.cpp). All three glibc tests (`bare_spin`, `profiled_spin`, `exit_fn`) either spin until cancelled or call `pthread_exit` immediately. A regression on the post-routine path would be invisible.

**Suggestion:** Add a test with a routine that returns normally and verify `cleanup_fn` was invoked.

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
```diff
@@ -73,6 +74,78 @@ static void unregister_and_release(int tid) {
     ProfiledThread::release();
 }
 
+// pthread_cleanup_push callback for thread wrappers.
+// Fires when the wrapped routine calls pthread_exit() or the thread is
+// canceled.  Kept noinline so its stack frame (which may hold a SignalBlocker
+// via unregister_and_release) lives outside the DEOPT-corruption zone of the
```
jbachorik, line 80: **[Sphinx Review — LOW]** No test in this PR detects removal of the `cleanup_fn(cleanup_arg)` call on the **glibc normal-return path** (after `__pthread_unregister_cancel`). All existing tests use routines that cancel or call `pthread_exit` — threads that return normally from `routine` are not covered, so `unregister_and_release` being silently skipped on that path would not be detected.

**Suggestion:** Add a glibc test asserting that `cleanup_fn` fires when `routine` returns normally.

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
```diff
@@ -73,6 +74,78 @@ static void unregister_and_release(int tid) {
...
+// caller on musl/aarch64, and so that the SignalBlocker's sigset_t does not
+// appear in the caller's frame on platforms with stack-protector canaries.
+__attribute__((noinline))
+static void cleanup_unregister(void*) {
+    unregister_and_release(ProfiledThread::currentTid());
```
jbachorik, line 85: **[Sphinx Review — LOW]** The **musl path of `run_with_cleanup`** (lines 82–86) is completely untested. The musl test section in `forced_unwind_ut.cpp` uses `pthread_cleanup_push` directly, bypassing `run_with_cleanup`. The `extern "C++"` declaration for `run_with_cleanup` is inside `#ifdef __GLIBC__` only, making it impossible to call on musl without changing the guard.

**Suggestion:** Move the `extern "C++" void run_with_cleanup(...)` declaration outside `#ifdef __GLIBC__`. Add musl tests that call `run_with_cleanup` directly (mirroring the glibc tests).

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: Fixed and updated comments

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: Fixed

ddprof-lib/src/test/cpp/thread_teardown_safety_ut.cpp
zhengyu123: Fixed

ddprof-lib/src/test/cpp/forced_unwind_ut.cpp
zhengyu123: Test added

ddprof-lib/src/test/cpp/thread_teardown_safety_ut.cpp
zhengyu123: Done

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: Done

ddprof-lib/src/test/cpp/thread_teardown_safety_ut.cpp
zhengyu123: Fixed

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: Added

ddprof-lib/src/test/cpp/thread_teardown_safety_ut.cpp
zhengyu123: Removed

ddprof-lib/src/test/cpp/thread_teardown_safety_ut.cpp
```diff
@@ -255,27 +265,51 @@ static std::atomic<bool> g_t07_release_ran{false};
...
+    __pthread_unwind_next(&cancel_buf);
+  }
+  __pthread_register_cancel(&cancel_buf);
+  // Inject a signal before the cancellation point to exercise the combined path.
+  pthread_kill(pthread_self(), SIGVTALRM);
+  while (true) {
+    pthread_testcancel();
+    usleep(100);
   }
+  __pthread_unregister_cancel(&cancel_buf);
```
jbachorik: **[Sphinx Review — MEDIUM]** Lines 311–314 are still unreachable dead code. The `while(true)` loop (lines 307–310) can only exit via `pthread_testcancel()` → longjmp → `[[noreturn]] __pthread_unwind_next`. The dead section includes a second `ProfiledThread::release()` at line 313 — the cleanup path (lines 299–302) already called it — so line 313 is a latent double-release use-after-free hazard if the code structure ever changes.

**Suggestion:** Remove lines 311–314 entirely.

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
```diff
@@ -227,13 +281,8 @@ static void* start_routine_wrapper(void* args) {
...
-    // RAII cleanup: reads tid from TLS in the destructor (same rationale as
-    // start_routine_wrapper_spec: avoids storing state on a potentially corruptible frame).
-    // unregister_and_release() wraps the two calls under SignalBlocker (PROF-14603).
-    struct Cleanup {
-        ~Cleanup() { unregister_and_release(ProfiledThread::currentTid()); }
-    } cleanup;
-    routine(params);
+    // Use POSIX cleanup instead of C++ RAII to handle pthread_exit(): see run_with_cleanup.
+    run_with_cleanup(routine, params, cleanup_unregister, nullptr);
     return nullptr;
```
jbachorik: **[Sphinx Review — MEDIUM]** `start_routine_wrapper` uses `return nullptr` after `run_with_cleanup` without explaining why it is safe, while the neighboring `start_routine_wrapper_spec` explicitly requires `pthread_exit(nullptr)` because *"the saved LR in this frame is corrupted by DEOPT PACKING; returning would jump to a garbage address"*.

The `return nullptr` is actually safe here because `start_routine_wrapper` is only compiled for non-aarch64 targets (it lives in the `#else` branch of `#ifdef __aarch64__`), where DEOPT PACKING does not corrupt the LR. But this is not documented, creating a maintenance trap for anyone who ports to a new architecture where the corruption applies.

**Suggestion:** Add a comment: `// safe to return on non-aarch64: DEOPT PACKING only corrupts the LR on aarch64 (see start_routine_wrapper_spec above)`. Alternatively use `pthread_exit(nullptr); __builtin_unreachable();` for consistency.

ddprof-lib/src/test/cpp/forced_unwind_ut.cpp
```diff
@@ -110,38 +134,97 @@ TEST(ForcedUnwindTest, ProfiledThreadReleasedOnForcedUnwind) {
...
+    g_nr_past_run.store(false, std::memory_order_relaxed);
+    run_with_cleanup(nr_routine, nullptr, nr_cleanup, nullptr);
+    g_nr_past_run.store(true, std::memory_order_relaxed);
+    // Spin with a cancellation point so the main thread can probe whether
+    // __pthread_unregister_cancel took effect after run_with_cleanup returned.
+    while (true) {
+        pthread_testcancel();
+        usleep(50);
+    }
+}
```
jbachorik, line 199: **[Sphinx Review — MEDIUM]** `nr_thread` returns `void*` but has no `return` statement (or `__builtin_unreachable()`) after the `while(true)` spin loop. The loop is infinite and only exits via forced-unwind (cancellation), which never returns to the caller — but the compiler does not know that. This is undefined behaviour per the C++ standard and will trigger `-Werror=return-type` on hardened build configurations.

**Suggestion:** Add `__builtin_unreachable();` after the closing brace of the `while(true)` loop, mirroring the pattern in `start_routine_wrapper_spec` (line 235 of libraryPatcher_linux.cpp).

ddprof-lib/src/test/cpp/forced_unwind_ut.cpp
```diff
@@ -110,38 +134,97 @@ TEST(ForcedUnwindTest, ProfiledThreadReleasedOnForcedUnwind) {
...
+
+// Normal-return path (matches start_routine_wrapper on non-aarch64 glibc):
+// cleanup_fn is called once AND __pthread_unregister_cancel removes the buf.
+// Removing either call on this path is detected:
+//   - no cleanup_fn             → g_nr_cleanup_count stays 0
+//   - no __pthread_unregister_cancel → post-return cancel longjmps into the
+//     freed cancel_buf (UB: crash or g_nr_cleanup_count becomes 2)
+TEST(ForcedUnwindTest, CleanupCalledExactlyOnceOnNormalReturn) {
```
jbachorik, line 211: **[Sphinx Review — LOW]** The `CleanupCalledExactlyOnceOnNormalReturn` test detects a missing `__pthread_unregister_cancel` by relying on UB: the comment on lines 201–203 documents that the detection mechanism is "UB: crash or g_nr_cleanup_count becomes 2". Under ASAN/MSAN or with aggressive optimization, the longjmp into the freed stack frame of `run_with_cleanup` may silently succeed without incrementing the counter, producing a false pass.

**Suggestion:** The test is sound for normal builds. Add a comment acknowledging the sanitizer limitation, e.g.: `// Note: under ASAN/MSAN the UAF detection path (count==2) is not guaranteed — a crash is also acceptable evidence.`

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: Fixed

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: Fixed

ddprof-lib/src/test/cpp/thread_teardown_safety_ut.cpp
zhengyu123: Fixed

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: Fixed

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: Done

ddprof-lib/src/test/cpp/forced_unwind_ut.cpp
zhengyu123: Added comment

ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp
zhengyu123: Added test

jbachorik (approved):
The changes look feasible. All sanitizers and fuzzers are running fine. 
LGTM

PR #596, opened 2026-06-12
build(native): compile x86_64 Linux lib with -mtls-dialect=gnu2

nsavoire:
## What

Compile the profiler shared library with `-mtls-dialect=gnu2` on **x86_64 Linux**.

## Why

TLS descriptors (GNU2 dialect) is required for [thread context sharing](https://github.com/open-telemetry/opentelemetry-specification/pull/4947).

## How

- Factored the shared flags into `commonCompilerArgs()`.
- Added a new `commonLinuxCompilerArgs()` helper that appends `-mtls-dialect=gnu2` gated on the x86_64 architecture.
- The ELF-only flag is therefore excluded from macOS (Mach-O, clang would reject it) and ARM64 builds.

## Testing

- `./gradlew -p build-logic :conventions:compileKotlin` — build-logic plugin compiles.
- Full `./gradlew build` on macOS/ARM64 succeeds; verified the flag is **absent** from the local native compile command (correct, since it's x86_64-Linux-only). The x86_64 Linux path is exercised in CI.

### DIFF
--- a/build-logic/conventions/src/main/kotlin/com/datadoghq/native/config/ConfigurationPresets.kt
+++ b/build-logic/conventions/src/main/kotlin/com/datadoghq/native/config/ConfigurationPresets.kt
@@ -4,8 +4,10 @@ package com.datadoghq.native.config
 import com.datadoghq.native.model.Architecture
 import com.datadoghq.native.model.BuildConfiguration
 import com.datadoghq.native.model.Platform
+import com.datadoghq.native.util.CompilerInfo
 import com.datadoghq.native.util.PlatformUtils
 import java.io.File
+import org.gradle.api.logging.Logger
 
 /**
  * Provides factory methods for creating standard build configurations
@@ -35,23 +37,25 @@ object ConfigurationPresets {
         val compiler = PlatformUtils.findCompiler(project)
 
         project.logger.lifecycle("Setting up standard build configurations for $currentPlatform-$currentArch")
-        project.logger.lifecycle("Using compiler: $compiler")
+        project.logger.lifecycle("Using compiler: ${compiler.executable} (version ${compiler.majorVersion})")
+
+        val commonCompilerArgs = commonCompilerArgs(version, compiler, currentArch, currentPlatform, project.logger)
 
         extension.buildConfigurations.apply {
             register("release") {
-                configureRelease(this, currentPlatform, currentArch, version, compiler)
+                configureRelease(this, currentPlatform, currentArch, commonCompilerArgs)
             }
             register("debug") {
-                configureDebug(this, currentPlatform, currentArch, version, compiler)
+                configureDebug(this, currentPlatform, currentArch, commonCompilerArgs)
             }
             register("asan") {
-                configureAsan(this, currentPlatform, currentArch, version, rootDir, compiler)
+                configureAsan(this, currentPlatform, currentArch, rootDir, compiler, commonCompilerArgs)
             }
             register("tsan") {
-                configureTsan(this, currentPlatform, currentArch, version, rootDir, compiler)
+                configureTsan(this, currentPlatform, currentArch, rootDir, compiler, commonCompilerArgs)
             }
             register("fuzzer") {
-                configureFuzzer(this, currentPlatform, currentArch, version, rootDir, compiler)
+                configureFuzzer(this, currentPlatform, currentArch, rootDir, commonCompilerArgs)
             }
         }
 
@@ -59,7 +63,13 @@ object ConfigurationPresets {
         project.logger.lifecycle("Active configurations: ${activeConfigs.map { it.name }.joinToString(", ")}")
     }
 
-    private fun commonCompilerArgs(version: String): List<String> {
+    private fun commonCompilerArgs(
+        version: String,
+        compiler: CompilerInfo,
+        architecture: Architecture,
+        platform: Platform,
+        logger: Logger
+    ): List<String> {
         val args = mutableListOf(
             "-fPIC",
             "-fno-omit-frame-pointer",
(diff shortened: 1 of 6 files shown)


build-logic/conventions/src/main/kotlin/com/datadoghq/native/config/ConfigurationPresets.kt
```diff
@@ -78,6 +78,22 @@ object ConfigurationPresets {
...
+        version: String,
+        architecture: Architecture = Architecture.X64,
+        compiler: String = "gcc"
+    ): List<String> {
+        val args = commonCompilerArgs(version).toMutableList()
+        // Use TLS descriptors (GNU2 dialect) for thread-local storage on x86_64.
+        // This is required by thread context (https://github.com/open-telemetry/opentelemetry-specification/pull/4947)
+        // The dialect is only supported by GCC and clang >= 19, so probe the
+        // compiler before adding the flag rather than relying on its version.
+        if (architecture == Architecture.X64 && PlatformUtils.supportsTlsDialectGnu2(compiler)) {
```
jbachorik: **[Sphinx Review — MEDIUM]** On x86_64 Linux the gnu2 TLS dialect is documented as required for thread context, yet `supportsTlsDialectGnu2` returns `false` for any environmental probe failure (timeout, swallowed exception, process-spawn restriction). In all such cases the flag is silently dropped and the build still succeeds, producing a binary whose thread-context feature is silently broken with no warning.

**Suggestion:** Log a `lifecycle`/`warn` message when the probe returns `false` on x86_64 Linux so a silent drop of the required flag is visible, or distinguish 'compiler genuinely rejects flag' from 'probe could not run' and fail loudly on the latter.

build-logic/conventions/src/main/kotlin/com/datadoghq/native/util/PlatformUtils.kt
```diff
@@ -199,6 +199,45 @@ object PlatformUtils {
...
     }
 
+    /**
+     * Check whether the given C++ compiler accepts -mtls-dialect=gnu2 on x86_64.
+     * GCC has supported the GNU2 TLS dialect for a long time; clang only gained
+     * x86 support in clang 19. Probing the real compiler is more robust than
+     * parsing version strings (it also handles vendor-patched toolchains).
+     * Always false off x86_64 Linux — the flag is ELF/x86_64 specific.
+     */
+    fun supportsTlsDialectGnu2(compiler: String): Boolean {
```
jbachorik: **[Sphinx Review — MEDIUM]** `supportsTlsDialectGnu2` spawns a compiler subprocess each time it is called and is invoked once per build configuration (release, debug, asan, tsan, fuzzer = up to 5 times per build). The result depends only on the fixed compiler path and host platform, so it is recomputed redundantly. There is no memoization, in contrast to `currentPlatform`/`currentArchitecture` which use `by lazy`.

**Suggestion:** Memoize the result keyed by compiler path (e.g. a `private val tlsGnu2Cache = ConcurrentHashMap<String, Boolean>()` with `cache.getOrPut(compiler) { ... }`) so the probe runs at most once per distinct compiler per build.

build-logic/conventions/src/main/kotlin/com/datadoghq/native/util/PlatformUtils.kt
```diff
@@ -199,6 +199,45 @@ object PlatformUtils {
...
+        if (currentPlatform != Platform.LINUX || currentArchitecture != Architecture.X64) {
```
jbachorik: **[Sphinx Review — LOW]** The X64 architecture gate is duplicated across two functions keyed off different sources: `commonLinuxCompilerArgs` checks the target `architecture` parameter, while `supportsTlsDialectGnu2` checks the host `currentArchitecture`. The two guards would silently diverge under cross-compilation where host ≠ target.

**Suggestion:** Either remove the inner X64 guard from `supportsTlsDialectGnu2` (let callers decide when to probe) and update the KDoc, or add a comment explaining why the redundant guard is intentional.

build-logic/conventions/src/main/kotlin/com/datadoghq/native/util/PlatformUtils.kt
```diff
@@ -199,6 +199,45 @@ object PlatformUtils {
...
+            return false
+        }
+        return try {
+            val testFile = createTempFile("tls_dialect_check", ".cpp")
+            try {
+                testFile.writeText("int main() { return 0; }")
+                val process = ProcessBuilder(
+                    compiler,
+                    "-mtls-dialect=gnu2",
+                    "-fsyntax-only",
```
jbachorik: **[Sphinx Review — LOW]** The probe uses `-fsyntax-only`, which skips code generation. `-mtls-dialect` is a codegen option; a compiler (notably clang) may accept it under `-fsyntax-only` yet fail or ignore it during an actual compile, so the probe can over-report support.

**Suggestion:** Compile to an object (`-c -o <tmp.o>`) instead of `-fsyntax-only` so codegen actually exercises the TLS dialect, then check exit status.

build-logic/conventions/src/main/kotlin/com/datadoghq/native/util/PlatformUtils.kt
```diff
@@ -199,6 +199,45 @@ object PlatformUtils {
...
+                    testFile.toAbsolutePath().toString()
+                ).redirectErrorStream(true).start()
```
jbachorik: **[Sphinx Review — LOW]** The probe process uses `redirectErrorStream(true)` but never drains `process.inputStream` before `waitFor`. If the compiler produced enough output to fill the OS pipe buffer, the child would block on write and the 10s timeout would expire, silently returning `false` and disabling the TLS flag.

**Suggestion:** Drain the combined stream before `waitFor` (e.g. `process.inputStream.bufferedReader().readText()`), or redirect output to `ProcessBuilder.Redirect.DISCARD` so the child never blocks on a full pipe.

build-logic/conventions/src/main/kotlin/com/datadoghq/native/config/ConfigurationPresets.kt
nsavoire: `-mtls-dialect=gnu2` is now gated on clang version and an exception is thrown when clang version cannot be determined.

build-logic/conventions/src/main/kotlin/com/datadoghq/native/util/PlatformUtils.kt
nsavoire: I don't think that's an issue, just getting the compiler version is fast.

build-logic/conventions/src/main/kotlin/com/datadoghq/native/util/PlatformUtils.kt
nsavoire: Cross-compilation does not seem to be supported, but removed the check from `commonLinuxCompilerArgs`.

build-logic/conventions/src/main/kotlin/com/datadoghq/native/util/PlatformUtils.kt
nsavoire: Switched to checking clang version.

build-logic/conventions/src/main/kotlin/com/datadoghq/native/util/PlatformUtils.kt
nsavoire: Fixed (same issue is present in `isCompilerAvailable`)

PR #598, opened 2026-06-15
feat(PROF-15098): add batch reapply-by-id-and-bytes context API

jbachorik:
**What does this PR do?**:

Adds `setContextAttributesByIdAndBytes(int[], byte[][])` to `ThreadContext`, `JavaProfiler`, and `ContextSetter`. For each slot whose `constantId > 0`, writes both the DD sidecar encoding (read by the signal handler) and the OTEP `attrs_data` UTF-8 value (read by external eBPF/OTEP profilers) inside a single `detach`/`attach` window — no String allocation, hashing, or Dictionary lookup.

Also fixes the `snapshotTags(int[])` guard from `<=` to `>=` (the old condition silently skipped the copy when the snapshot array was exactly the right size), and links the two independent `MAX_CUSTOM_SLOTS` / `TAGS_STORAGE_LIMIT` constants with cross-referencing comments.

Fixes `ContextWallClockTest` weight attribution: `method1Impl` accumulates wall-clock samples during `monitor.wait()` while `method3` runs concurrently on the executor thread. Those samples were inflating `method1`'s share to ~55% instead of the expected ~33%. Excluding `Object.wait` frames from the `method1Impl` attribution measures self-time only. The 0.3 `allowedError` for vm/vmx/fp/dwarf modes is retained to cover DWARF/VM trace-ID fragmentation that sheds some `method2`/`method3` samples.

**Motivation**:

PROF-15098 — `dd-trace-java`'s `reapplyAppContext` hot path must restore application-managed context attributes after a `setContext` span-activation wipes all custom slots. The existing `setContextAttribute(int, String)` API re-hashes and re-encodes the value on every call; the new API accepts the precomputed constant ID (from `snapshotTags`) and UTF-8 bytes (retained from the original String) so the hot path is zero-allocation.

**Implementation notes**:

- Two-pass validation (check all slots before any write) prevents half-written records on malformed input.
- `setContextAttributesByIdAndBytes` returns `false` and skips all writes when the record is already invalid (`valid=0`), preventing accidental resurrection of a cleared span-less record.
- All N slot writes share one `detach`/`attach` window so a signal handler never observes a partially re-applied attribute set.
- A shared `writeSlot` helper is used by both `setContextAttributeDirect` and the new batch path to avoid duplication of the sidecar + attrs_data write logic.
- The `(constantId, bytes)` pairing is unverifiable by the callee; the contract is documented in the Javadoc.

**Benchmarks** (macOS/arm64, JDK 21, `avgt` ns/op):

| Benchmark | Score | Error |
|---|---|---|
| `reapplyByIdAndBytes` — bare reapply, IDs+bytes already in hand | **115 ± 1.5 ns** | |
| `reapplyCycle` — full cycle: `setContext` wipe + reapply (1t) | **45 ± 0.8 ns** | |
| `reapplyCycle_2t` | **46 ± 0.3 ns** | |
| `reapplyCycle_4t` | **60 ± 1.4 ns** | |
| `spanLifecycle` (baseline: `setContext` + one `setContextAttribute`) | 14 ns | |
| `setContextFull` | 6 ns | |

`reapplyCycle` at 45 ns restores 5 slots (compact + insert each) vs `spanLifecycle` at 14 ns for a single attribute — expected given the 5× attrs_data work. Scales cleanly under thread contention (46→60 ns at 4 threads, same pattern as `spanLifecycle`).

**How to test the change?**:

Six new tests in `TagContextTest` (deterministic `readContextAttribute` / `snapshotTags` readback — no flaky JFR-sample assertions):

- `testReapplyByIdAndBytes` — happy path: set, snapshot, wipe via `setContext`, reapply, verify both sidecar and attrs_data views
- `testReapplyByIdAndBytesRejectsBadArgs` — null arrays, length mismatch, null/oversized bytes
- `testReapplyByIdAndBytesReplacesExistingValue` — compact-then-insert path (overwrites existing value)
- `testReapplyByIdAndBytesAfterClear` — reapply after per-slot `clearContextValue` (`valid` stays 1)
- `testReapplyByIdAndBytesClearedRecord` — reapply on a `clearContext`-cleared record (`valid=0`); must return `false` and not resurrect
- `testReapplyByIdAndBytesOverflowRollback` — 10×255-byte values overflow `attrs_data`; overflowed slot's sidecar zeroed; slots written before the overflow are retained

Stress coverage: `ReapplyContextAntagonist` added to the chaos harness — drives the full reapply cycle from 8 threads while wall-clock signals fire, racing the `detach`/`attach` window. Run with `--antagonists reapply-context`.

**For Datadog employees**:

- [x] JIRA: [PROF-15098](https://datadoghq.atlassian.net/browse/PROF-15098)

[PROF-15098]: https://datadoghq.atlassian.net/browse/PROF-15098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddprof-lib/src/main/java/com/datadoghq/profiler/ThreadContext.java
```diff
@@ -313,17 +314,89 @@ private boolean setContextAttributeDirect(int keyIndex, String value) {
...
+     * @return true if every slot with {@code constantId > 0} was written; false on invalid
+     *         arguments (null arrays, length mismatch, or a missing/oversized {@code utf8} entry),
+     *         if the record was not valid before the call (nothing published), or if any slot
+     *         overflowed attrs_data (that slot's sidecar is zeroed)
+     */
+    public boolean setContextAttributesByIdAndBytes(int[] constantIds, byte[][] utf8) {
+        if (constantIds == null || utf8 == null || constantIds.length != utf8.length) {
+            return false;
+        }
+        int len = Math.min(constantIds.length, MAX_CUSTOM_SLOTS);
```
jbachorik: 🔵 **LOW** · `completeness`

Arrays longer than `MAX_CUSTOM_SLOTS` are silently truncated: slots >= 10 with `constantId` > 0 are never written, yet the method still returns true, violating the Javadoc contract that 'true' means every slot with `constantId` > 0 was written.

> 💡 **Suggestion:** Either reject inputs where `constantIds`.length > `MAX_CUSTOM_SLOTS` by returning false, or document the truncation behavior explicitly in the Javadoc so callers know slots beyond `MAX_CUSTOM_SLOTS` are ignored.

ddprof-lib/src/main/java/com/datadoghq/profiler/ContextSetter.java
```diff
@@ -29,8 +30,12 @@ public int[] snapshotTags() {
         return snapshot;
     }
 
+    /**
+     * Copies current sidecar encodings into {@code snapshot}. The copy only runs when
+     * {@code snapshot.length >= attributes.size()}; callers must size the array accordingly.
+     */
     public void snapshotTags(int[] snapshot) {
```
jbachorik, line 54: 🟡 **MEDIUM** · `correctness`

`snapshotTags(int[])` now copies into oversized arrays, filling indices attributes.`size()`..`MAX_CUSTOM_SLOTS`-1 with raw sidecar values not owned by this `ContextSetter` instance

> 💡 **Suggestion:** Restrict the copy to the managed range: only invoke `copyTags` when snapshot.length == attributes.`size()`, or zero indices beyond attributes.`size()` after copying, and document that the array must be exactly attributes.`size()`.

ddprof-lib/src/main/java/com/datadoghq/profiler/ThreadContext.java
jbachorik: Fixed in 00194f8ae.

ddprof-lib/src/main/java/com/datadoghq/profiler/ContextSetter.java
jbachorik: Fixed in 00194f8ae.

jbachorik:
@codex review

ddprof-test/src/test/java/com/datadoghq/profiler/context/TagContextTest.java
```diff
@@ -239,6 +240,345 @@ public void testCrossSlotIsolation() throws Exception {
...
+                new int[] {id, 0}, new byte[][] {null, null}));
+        assertFalse(contextSetter.setContextValuesByIdAndBytes(
+                new int[] {id, 0}, new byte[][] {new byte[256], null}));
+
+        // 255 bytes is the boundary and must be accepted.
+        byte[] ok255 = new byte[255];
+        Arrays.fill(ok255, (byte) 'x');
+        assertTrue(contextSetter.setContextValuesByIdAndBytes(
+                new int[] {id, 0}, new byte[][] {ok255, null}));
+    }
```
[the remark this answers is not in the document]
jbachorik: Fixed in 3d3676470.

ddprof-lib/src/main/java/com/datadoghq/profiler/ThreadContext.java
```diff
@@ -313,17 +314,93 @@ private boolean setContextAttributeDirect(int keyIndex, String value) {
...
+     *
+     * @param constantIds per-slot Dictionary constant IDs; entries {@code <= 0} are skipped
+     * @param utf8        per-slot UTF-8 value bytes; must be non-null and at most
+     *                    {@value #MAX_VALUE_BYTES} bytes for every slot whose constantId {@code > 0}
+     * @return true if every slot with {@code constantId > 0} was written; false on invalid
+     *         arguments (null arrays, length mismatch, a missing/oversized {@code utf8} entry,
+     *         or {@code constantIds.length > MAX_CUSTOM_SLOTS}), if the record was not valid
+     *         before the call (nothing published), or if any slot overflowed attrs_data (that
+     *         slot's sidecar is zeroed)
+     */
```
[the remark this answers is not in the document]
jbachorik: Fixed in 3d3676470.

ddprof-stresstest/src/jmh/java/com/datadoghq/profiler/stresstest/scenarios/throughput/ThreadContextBenchmark.java
```diff
@@ -81,6 +82,45 @@ public void setup(ProfilerState ps) {
...
+        @Setup(Level.Iteration)
+        public void resetToSteadyState() {
+            // Re-establish a live span (valid=1) and pre-populate attrs_data with all slots
+            // before each measurement window. Without this, reapplyByIdAndBytes sees a
+            // different attrs_data state on the first invocation of each iteration (empty
+            // after put() in the previous iteration or after the trial setup), causing a
+            // bimodal distribution across forks due to JIT profile divergence.
+            ctx.put(localRootSpanId, spanId, 0, spanId);
+            ctx.setContextAttributesByIdAndBytes(constantIds, utf8);
+        }
```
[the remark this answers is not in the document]
jbachorik: Fixed in 3d3676470.

ddprof-stresstest/src/chaos/java/com/datadoghq/profiler/chaos/ReapplyContextAntagonist.java
```diff
@@ -0,0 +1,131 @@
...
+ * Drives the reapply-by-id-and-bytes hot path continuously from multiple threads while the
+ * profiler's wall-clock signal fires, racing the {@code detach}/{@code attach} window.
+ *
+ * <p>Each worker thread loops: span-activate (wiping slots) → reapply snapshot. This mirrors
+ * dd-trace-java's {@code reapplyAppContext} pattern and exercises the single-window invariant
+ * (no partial publish visible to a signal handler) under high thread count and signal pressure.
+ *
+ * <p>The only failure signal is a JVM crash or a reapply returning {@code false} while the record
+ * is active — the latter is logged and counted but does not abort the run (false positives from
+ * genuine attrs_data overflow are possible with many threads).
```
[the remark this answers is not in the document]
jbachorik: Fixed in 3d3676470.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
```diff
@@ -221,6 +221,10 @@ public void clearContextAttribute(int offset) {
         tlsContextStorage.get().clearContextAttribute(offset);
     }
 
+    public boolean setContextAttributesByIdAndBytes(int[] constantIds, byte[][] utf8) {
```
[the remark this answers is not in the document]
jbachorik: Fixed in 3d3676470.

ddprof-lib/src/main/java/com/datadoghq/profiler/ContextSetter.java
```diff
@@ -1,12 +1,14 @@
 package com.datadoghq.profiler;
 
 import java.util.ArrayList;
+import java.util.Arrays;
```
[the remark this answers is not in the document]
jbachorik: Fixed in 3d3676470.

jbachorik:
@codex review

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
```diff
@@ -213,14 +214,53 @@ public void clearContext() {
         tlsContextStorage.get().put(0, 0, 0, 0);
     }
 
+    /**
+     * Sets a custom context attribute at the given slot offset for the current thread.
+     *
+     * @param offset slot index (0-based, must be less than {@code ThreadContext.MAX_CUSTOM_SLOTS})
+     * @param value  the string value to record; null or empty clears the slot
+     * @return true if the value was recorded; false if the slot index is out of range, the
+     *         Dictionary is full, or {@code attrs_data} overflows for this slot
```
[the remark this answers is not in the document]
jbachorik: JavaProfiler.java lines 220-226 in the PR head already state the concrete range [0, 9] and correctly describe null (returns false) and empty string (zero-length entry, not a clear). The private-constant reference and the null/empty clearing claim are gone. No further change needed.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
```diff
@@ -213,14 +214,53 @@ public void clearContext() {
...
     public boolean setContextAttribute(int offset, String value) {
         return tlsContextStorage.get().setContextAttribute(offset, value);
     }
 
+    /**
+     * Clears the custom context attribute at the given slot offset for the current thread.
+     * Zeros the sidecar encoding and removes it from OTEP {@code attrs_data}.
+     *
+     * @param offset slot index (0-based, must be less than {@code ThreadContext.MAX_CUSTOM_SLOTS})
+     */
```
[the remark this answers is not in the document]
jbachorik: JavaProfiler.java line 236 in the PR head documents the concrete range [0, 9] without any reference to the private ThreadContext.MAX_CUSTOM_SLOTS constant.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
```diff
@@ -213,14 +214,53 @@ public void clearContext() {
...
+     *   <li>Returns {@code false} without writing if the thread's record is not currently valid
+     *       (span-less), to avoid resurrecting a cleared record.</li>
+     *   <li>On {@code attrs_data} overflow, the overflowed slot's sidecar is zeroed and
+     *       {@code false} is returned; slots written before the overflow are retained.</li>
+     * </ul>
+     *
+     * @param constantIds per-slot Dictionary constant IDs; entries {@code <= 0} are skipped
+     * @param utf8        per-slot UTF-8 value bytes; must be non-null and at most
+     *                    {@code ThreadContext.MAX_VALUE_BYTES} bytes for every slot whose
+     *                    {@code constantId > 0}
```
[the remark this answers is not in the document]
jbachorik: JavaProfiler.java line 257 in the PR head says '255 bytes' explicitly, replacing the private-constant reference. ThreadContext's own Javadoc uses the valid {@value #MAX_VALUE_BYTES} form (self-reference to a constant in the same class). Both usages are correct.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
```diff
@@ -213,14 +214,53 @@ public void clearContext() {
         tlsContextStorage.get().put(0, 0, 0, 0);
     }
 
+    /**
+     * Sets a custom context attribute at the given slot offset for the current thread.
+     *
+     * @param offset slot index (0-based, must be less than {@code ThreadContext.MAX_CUSTOM_SLOTS})
+     * @param value  the string value to record; null or empty clears the slot
```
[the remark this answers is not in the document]
jbachorik: JavaProfiler.java lines 221-225 in the PR head already document the correct contract: 'null returns false without writing; an empty string is written as a zero-length entry (not a clear — use clearContextAttribute(int) to remove a value)'. This matches ThreadContext.setContextAttribute's actual behavior at lines 274-279.

ddprof-lib/src/main/java/com/datadoghq/profiler/ContextSetter.java
```diff
@@ -29,9 +46,17 @@ public int[] snapshotTags() {
...
+     * copying to prevent stale data from leaking to the caller.
+     * Use the no-arg {@link #snapshotTags()} overload to obtain a correctly sized array.
+     */
     public void snapshotTags(int[] snapshot) {
-        if (snapshot.length <= attributes.size()) {
+        if (snapshot.length >= attributes.size()) {
             profiler.copyTags(snapshot);
+            Arrays.fill(snapshot, attributes.size(), snapshot.length, 0);
         }
     }
```
ivoanjo, line 59: I was grepping dd-trace-java AND java-profiler and `void snapshotTags(int[] snapshot)` never gets used directly other than for the `TagContextTest` that validates the behavior of mis-sized arrays.

This means that:
* The `snapshot` array is always new (does not need zeroing)
* The `snapshot` array is always exactly `attributes.size`

Going deeper -- `JavaProfiler::copyTags` is also only called exactly once -- here.

So summing it all up, I think we could simplify this by going:

```java
    public int[] snapshotTags() {
        int[] snapshot = new int[attributes.size()];
        profiler.copyTags(snapshot);
        return snapshot;
    }
```

and this would mean deleting at least 3 tests for sizing of the array in the API.

ddprof-lib/src/main/java/com/datadoghq/profiler/ContextSetter.java
ivoanjo: This constant is used exactly once -- in the constructor. Rather than having two things + comments to keep them in-sync, I suggest making `MAX_CUSTOM_SLOTS` package-protected or public and using it here directly OR having a `Profiler::getMaxSlots()` since we get that instance anyway, and avoid the complexity and potential bugs.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
```diff
@@ -213,14 +214,53 @@ public void clearContext() {
...
     public void clearContextAttribute(int offset) {
         tlsContextStorage.get().clearContextAttribute(offset);
     }
 
+    /**
+     * Re-applies multiple custom attributes from precomputed constant IDs and UTF-8 bytes for
+     * the current thread in a single detach/attach window.
+     *
+     * <p>Delegates to {@link ThreadContext#setContextAttributesByIdAndBytes(int[], byte[][])}
+     * on the current thread's context. Key contract points:
```
ivoanjo: Minor: TMI? It doesn't seem particularly interesting for the caller to know that this is implemented by delegating to something else 👀

ddprof-lib/src/main/java/com/datadoghq/profiler/ThreadContext.java
```diff
@@ -313,17 +314,95 @@ private boolean setContextAttributeDirect(int keyIndex, String value) {
...
+     * same value via {@link #copyCustoms(int[])} within the current profiler session, and
+     * {@code utf8[i]} must be that value's UTF-8 bytes. The (id, bytes) pairing is not verifiable
+     * here; a mismatch silently diverges the sidecar and attrs_data views.
+     *
+     * @param constantIds per-slot Dictionary constant IDs; entries {@code <= 0} are skipped
+     * @param utf8        per-slot UTF-8 value bytes; must be non-null and at most
+     *                    {@value #MAX_VALUE_BYTES} bytes for every slot whose constantId {@code > 0}
+     * @return true if every slot with {@code constantId > 0} was written successfully; false on
+     *         invalid arguments (null arrays, length mismatch, a missing/oversized {@code utf8}
+     *         entry, or {@code constantIds.length > MAX_CUSTOM_SLOTS}), if the record was not
```
ivoanjo: Minor: To match some of my above comments -- I think a bunch of these would only happen if there were real logic bugs so I'd suggest throwing exceptions rather than silently going `return false`. E.g. `constantIds` being null? That seems pretty bad? Or both arrays having mismatched lengths?

ddprof-lib/src/main/java/com/datadoghq/profiler/ThreadContext.java
```diff
@@ -313,17 +314,95 @@ private boolean setContextAttributeDirect(int keyIndex, String value) {
...
+        int len = constantIds.length;
+        // Validate before mutating so malformed input never leaves a half-written record.
+        for (int i = 0; i < len; i++) {
+            if (constantIds[i] > 0) {
+                byte[] bytes = utf8[i];
+                if (bytes == null || bytes.length > MAX_VALUE_BYTES) {
+                    return false;
+                }
+            }
+        }
```
ivoanjo, line 393: This validation is weird to me. _Specifically_ for this whole joint to work, we already strongly rely on the caller not lying about the constant ids -- e.g. not saying that constant 5 is "potato" where it was actually "carrot".

So there's already a lot of coupling with the caller needing to behave properly.

So it seems weird that we'd go "oh yeah, but actually, we'll go to the trouble and check the length separately anyway, just in case". It seems to me way more likely that the wrong bytes will be passed in, but they will still be under max value.

TL;DR: I suggest removing this, throwing an exception _somewhere_, or at most blend this check with the loop below, since I don't think it's important enough to run on its own.

ddprof-stresstest/src/chaos/java/com/datadoghq/profiler/chaos/ReapplyContextAntagonist.java
```diff
@@ -0,0 +1,132 @@
...
+        } catch (Exception e) {
+            System.err.println("[chaos] reapply-context: failed to get profiler: " + e);
+            return;
+        }
+        ThreadContext ctx = profiler.getThreadContext();
+
+        // Prime the per-thread encoding cache and capture a stable snapshot.
+        long spanId = Thread.currentThread().getId() + 1;
+        long localRootSpanId = spanId * 31L;
+        ctx.put(localRootSpanId, spanId, 0, spanId);
```
ivoanjo: Minor: Passing in the `spanId` in place of a `traceId` is always wrong so I'd suggest passing in a different random number to make it clear they are always distinct. (Also applies to `ThreadContextBenchmark` and `TagContextTest`)

(I wonder if claude mislearned this from the deprecated `public long put(long spanId, long rootSpanId)` in `ThreadContext` -- should we maybe remove that method now?)

ddprof-stresstest/src/chaos/java/com/datadoghq/profiler/chaos/ReapplyContextAntagonist.java
```diff
@@ -0,0 +1,132 @@
...
+        for (int i = 0; i < ROUTES.length; i++) {
+            ctx.setContextAttribute(i, ROUTES[i]);
+        }
+        int[] constantIds = new int[ROUTES.length];
+        ctx.copyCustoms(constantIds);
+        byte[][] utf8 = new byte[ROUTES.length][];
+        for (int i = 0; i < ROUTES.length; i++) {
+            utf8[i] = ROUTES[i].getBytes(StandardCharsets.UTF_8);
+        }
```
ivoanjo: Minor: This suspiciously looks like something that users of this new API will need to always do (see how `ThreadContextBenchmark` already duplicates it as well) -- should we maybe provide this as a method somewhere perhaps?

ddprof-stresstest/src/chaos/java/com/datadoghq/profiler/chaos/ReapplyContextAntagonist.java
```diff
@@ -0,0 +1,132 @@
...
+        while (running) {
+            // Span activation wipes all custom slots.
+            ctx.put(localRootSpanId, spanId, 0, spanId);
+            // Reapply restores them — this is the hot path under test.
+            if (!ctx.setContextAttributesByIdAndBytes(constantIds, utf8)) {
+                reapplyFailures.incrementAndGet();
+            }
+            // Burn a little CPU so wall-clock signals have something to sample.
+            r = r * 1103515245L + 12345L;
+            sink.addAndGet(r);
```
ivoanjo: Minor: Won't the above lines in the loop already burn cpu? I'm surprised that this would still be needed.

rkennke (approved):
Good change!
Have you verified that dd-trace-java is ok with the flip <= vs >= semantics in  snapshotTags(int[])?

  - Oversized (length > size): old = silent no-op (the bug); new = copies correctly.
  - Undersized (length < size): old = partial copy up to length; new = no-op.

So any external caller that passed a deliberately-undersized buffer expecting a partial copy would now get nothing.

ddprof-stresstest/src/chaos/java/com/datadoghq/profiler/chaos/ReapplyContextAntagonist.java
```diff
@@ -0,0 +1,132 @@
...
+            utf8[i] = ROUTES[i].getBytes(StandardCharsets.UTF_8);
+        }
+
+        long r = spanId;
+        while (running) {
+            // Span activation wipes all custom slots.
+            ctx.put(localRootSpanId, spanId, 0, spanId);
+            // Reapply restores them — this is the hot path under test.
+            if (!ctx.setContextAttributesByIdAndBytes(constantIds, utf8)) {
+                reapplyFailures.incrementAndGet();
```
ivoanjo: Should this throw instead? Specifically, in this test, reapply should never fail, so it's weird to treat it as a counter vs a "something really really weird is going on, please stop and look at it"?

ddprof-test/src/test/java/com/datadoghq/profiler/context/TagContextTest.java
```diff
@@ -239,6 +255,357 @@ public void testCrossSlotIsolation() throws Exception {
         assertNull(readTag(contextSetter, "tag2"));
     }
 
+    @Test
+    public void testReapplyByIdAndBytes() throws Exception {
+        Assumptions.assumeTrue(!Platform.isJ9());
```
ivoanjo: Minor: Every single test in this class starts with `Assumptions.assumeTrue(!Platform.isJ9());` -- maybe move it to the constructor or a `@BeforeEach`? ;)

jbachorik:
> Have you verified that dd-trace-java is ok with the flip

There will be a followup PR for dd-trace-java - it will need to get more adjustments for this functionality, anyhow

ivoanjo (approved):
👍 This looks good to me. I've left a bunch of comments, but they're all in the "we can remove a bit of code here, we can throw instead of swallowing failure here, we can simplify a bit here" vibe so nothing big

ddprof-lib/src/main/java/com/datadoghq/profiler/ContextSetter.java
jbachorik: Fixed in 18faa110d.

ddprof-lib/src/main/java/com/datadoghq/profiler/ContextSetter.java
jbachorik: Fixed in 18faa110d.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
jbachorik: Fixed in 18faa110d.

ddprof-lib/src/main/java/com/datadoghq/profiler/ThreadContext.java
jbachorik: Fixed in 18faa110d.

ddprof-lib/src/main/java/com/datadoghq/profiler/ThreadContext.java
jbachorik: Fixed in 18faa110d.

ddprof-stresstest/src/chaos/java/com/datadoghq/profiler/chaos/ReapplyContextAntagonist.java
jbachorik: Fixed in 18faa110d.

ddprof-lib/src/main/java/com/datadoghq/profiler/ContextSetter.java
jbachorik: `snapshotTags(int[])` is actually used — dd-trace-java PR 11646 calls it with a caller-managed buffer. Reverted the removal in 02e1b02af: the overload is back with the fixed `>=` guard and the three contract tests restored.

ddprof-stresstest/src/chaos/java/com/datadoghq/profiler/chaos/ReapplyContextAntagonist.java
jbachorik: True, it's duplicated. The String values are always caller-specific so a helper would need to take them as a parameter anyway. If dd-trace-java ends up calling this from multiple places we can revisit, but for now it's just 3 lines.

ddprof-stresstest/src/chaos/java/com/datadoghq/profiler/chaos/ReapplyContextAntagonist.java
jbachorik: Yes, still needed. `setContext` + `setContextValuesByIdAndBytes` are each ~50-100 ns, well below the wall-clock interval (~1 ms). Without the spin the profiler would mostly sample this thread in `addAndGet` or scheduler sleep rather than in the detach/attach window we want to stress.

ivoanjo: Wait, what scheduler sleep? I only see a `while (running) { do stuff on cpu }` above? :eyes:

jbachorik: 🤦 Serves me well for not doing review in GH UI ... was looking on an old version where I was actually playing with some extra sleeps. And trusting AI :/
This can be removed without any impact on the results - we are already running CPU and wallclock profilers so there is enough opportunities for interleaving the reapply-context operations with the signals.

ddprof-lib/src/main/java/com/datadoghq/profiler/ThreadContext.java
```diff
@@ -313,17 +314,101 @@ private boolean setContextAttributeDirect(int keyIndex, String value) {
...
+        if (constantIds.length > MAX_CUSTOM_SLOTS) {
+            throw new IllegalArgumentException("constantIds.length exceeds MAX_CUSTOM_SLOTS");
+        }
+        int len = constantIds.length;
+        // Validate active slots before touching the buffer so a bad input never leaves
+        // the record detached (valid=0) after an exception unwinds past attach().
+        for (int i = 0; i < len; i++) {
+            if (constantIds[i] > 0) {
+                byte[] bytes = Objects.requireNonNull(utf8[i], "utf8[" + i + "]");
```
ivoanjo, line 388: Wait, isn't this an allocation? 👀

ivoanjo: Specifically

```
"utf8[" + i + "]"
```

...?

jbachorik: Yes 😱 

https://github.com/DataDog/java-profiler/pull/606

PR #600, opened 2026-06-17
An alternative TLS implementation to fix a crash on musl/aarch64

zhengyu123:
**What does this PR do?**:
This PR is intended to replace `thread_local` variables (except Otel ones) with an alternative implementation, which is based on `pthread_get/setspecific()` API.

**Motivation**:
We have encountered a quite few defects related to TLS - whose implementation is compiler specific. In some circumstances, the implementation is very restrictive that does not fit our usage pattern - it manifests in [SCP-1238](https://datadoghq.atlassian.net/browse/SCP-1238)

**Additional Notes**:
To limit the scope, I only replaced 3 thread locals in `livenessTracker` which led to the crash reported in SCP-1238](https://datadoghq.atlassian.net/browse/SCP-1238)

Converting `ProfiledThread` to use newly introduced `ThreadLocal` will be done in separate PR.

**How to test the change?**:
- CI tests
- New test case.

**For Datadog employees**:
- [X] JIRA: [PROF-15114](https://datadoghq.atlassian.net/browse/PROF-15114)

[SCP-1238]: https://datadoghq.atlassian.net/browse/SCP-1238?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PROF-15114]: https://datadoghq.atlassian.net/browse/PROF-15114?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,129 @@
+/*
+ * Copyright 2026, Datadog, Inc.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef __THREADLOCAL_H__
+#define __THREADLOCAL_H__
```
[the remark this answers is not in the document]
zhengyu123: Fixed

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,129 @@
...
+        memcpy(&value, &val, sizeof(val));
+        return value;
+    }
+
+    void clear() {
+        pthread_setspecific(_key, nullptr);
+    }
+};
+
+#endif // __THREADLOCAL_H__
```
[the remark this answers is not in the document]
zhengyu123: Fixed

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,129 @@
...
+class ThreadLocal {
+protected:
+    pthread_key_t _key;
+ 
+public:
+    ThreadLocal() {
+        int err = pthread_key_create(&_key, F);
+        // What to do if we can not create a key?
+        assert(err == 0);
+    }
```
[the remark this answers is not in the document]
zhengyu123: Fixed

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,142 @@
...
+class ThreadLocal {
+protected:
+    pthread_key_t _key;
+ 
+public:
+    ThreadLocal(const ThreadLocal&) = delete;
+    ThreadLocal& operator=(const ThreadLocal&) = delete;
+
+    ThreadLocal() {
+        int err = pthread_key_create(&_key, F);
```
jbachorik: **[HIGH · robustness]** `pthread_key_create` failure is guarded only by `assert()`, which is compiled out under NDEBUG. In a release build a failed key creation leaves `_key` uninitialized; all subsequent `get()`/`set()` calls then operate on an arbitrary or zero key, silently aliasing another thread's TSD slot (e.g. the `ProfiledThread` key in thread.cpp).

Suggestion: replace the assert with an unconditional abort:
```cpp
if (err != 0) { fprintf(stderr, "ThreadLocal: pthread_key_create failed: %d\n", err); abort(); }
```
This matches the existing error handling in thread.cpp and makes failures visible in production.

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,142 @@
...
+        // What to do if we can not create a key?
+        assert(err == 0);
+    }
+
+    ~ThreadLocal() {
+        pthread_key_delete(_key);
```
jbachorik: **[HIGH · robustness]** `pthread_key_delete(_key)` is called unconditionally in the destructor. If `pthread_key_create` failed in the constructor and the `assert` was compiled out (NDEBUG / release build), `_key` stays at its zero-initialized value (0). Calling `pthread_key_delete(0)` deletes a key we don't own — on musl, key 0 is legitimately held by another component (see thread.cpp:91–93) — corrupting the global pthread key table.

Suggestion: track construction success with a `bool _key_valid` member:
```cpp
_key_valid = (pthread_key_create(&_key, F) == 0);
// ...
if (_key_valid) pthread_key_delete(_key);
```
Apply the same pattern to the `double` specialization.

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,142 @@
...
+
+    void set(T value) {
+        int err = pthread_setspecific(_key, (const void*)value);
+        assert(err == 0);
+    }
+
+    T get() {
+        void* p = pthread_getspecific(_key);
+        if (p == nullptr && C != nullptr) {
+            p = C();
```
jbachorik, line 91: **[HIGH · robustness]** After `p = C()` allocates an object, `set()` asserts on `pthread_setspecific`; under NDEBUG the assert is stripped. If `pthread_setspecific` then fails silently, the TSD slot stays null. Every subsequent `get()` re-invokes `C()` and leaks the prior allocation — unbounded per-call leak for the lazy-create path.

Suggestion: check `pthread_setspecific` directly after `C()` and free on failure:
```cpp
p = C();
if (pthread_setspecific(_key, (const void*)p) != 0) {
    if (F) F(p);
    return T{};
}
```
Alternatively, make `set()` call `abort()` instead of `assert` so failure is not silently swallowed in release builds.

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,142 @@
...
+        int err = pthread_key_create(&_key, F);
+        // What to do if we can not create a key?
+        assert(err == 0);
+    }
+
+    ~ThreadLocal() {
+        pthread_key_delete(_key);
+    }
+
+    void set(T value) {
```
jbachorik, line 81: **[MEDIUM · correctness]** The primary template casts `T` directly to `const void*` with no compile-time guard that `sizeof(T) == sizeof(void*)`. Instantiation with a sub-pointer-width type (e.g. `ThreadLocal<int>`) compiles silently but produces implementation-defined behaviour on the set/get round-trip for values that set the sign bit.

Suggestion: add a `static_assert` in the primary template constructor mirroring the guards already in the `double` specialization:
```cpp
static_assert(sizeof(T) == sizeof(void*),
    "ThreadLocal<T> requires sizeof(T)==sizeof(void*); use a pointer type or add a specialization");
```

jbachorik, line 81: **[MEDIUM · correctness]** `set(nullptr)` on a lazy-create `ThreadLocal` (i.e. `C != nullptr`) is semantically identical to 'never set': the next `get()` call will invoke `C()` and return a fresh object instead of null. This makes it impossible to store a null value or 'unset' without triggering re-creation, which is a silent foot-gun for the planned future conversions mentioned in the PR description.

Suggestion: document explicitly that `set(nullptr)` is reserved for `clear()` and must not be used when `C != nullptr`, or provide a separate `reset()` method that bypasses the lazy-create path.

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,142 @@
...
+        int err = pthread_setspecific(_key, (const void*)value);
+        assert(err == 0);
+    }
+
```
jbachorik, line 86: **[MEDIUM · necessity]** `clear()` has zero callers in production code and zero test coverage. The class docstring presents it as a core usage pattern, yet no shipped path exercises it — including the subtle branch where a failed `pthread_setspecific` in `clear()` leaves a still-reachable pointer un-freed.

Suggestion: either remove `clear()` until a caller needs it, or add unit tests that cover: (1) `clear()` resets the value, (2) the cleanup function is invoked, (3) `F == nullptr` branch actually zeroes the slot. Do not ship an untested method that the public docs present as core API.

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,142 @@
...
+    T get() {
+        void* p = pthread_getspecific(_key);
+        if (p == nullptr && C != nullptr) {
```
jbachorik, line 90: **[LOW · robustness]** `get()` cannot distinguish 'never set' from 'explicitly set to nullptr'. When a create function is registered, storing null (or `C()` returning null) causes `C()` to be re-invoked on every `get()`, repeatedly allocating. This differs from real `thread_local` semantics and is a foot-gun for the planned future conversions mentioned in the PR description.

Suggestion: document that `C` must never return nullptr and that `set(nullptr)` is reserved for `clear()`; optionally `assert(p != nullptr)` after `C()` to fail fast.

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,142 @@
...
+            p = C();
+            set((T)p);
+        }
+        return (T)p;
+    }
+
+    // Clear the value
+    void clear() {
```
jbachorik, line 98: **[MEDIUM · correctness]** `clear()` silently skips `pthread_setspecific` when `F == nullptr`, leaving the TSD slot non-zero. The documented use-case — clear before re-set — does not work for pointer types without a destructor function.

Suggestion: always zero the TSD slot and only call `F()` when non-null:
```cpp
void clear() {
    void* p = pthread_getspecific(_key);
    if (p != nullptr) {
        pthread_setspecific(_key, nullptr);
        if (F != nullptr) F(p);
    }
}
```

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,142 @@
...
+        void* p = nullptr;
+        if (F != nullptr && (p = pthread_getspecific(_key)) != nullptr) {
+            int err = pthread_setspecific(_key, nullptr);
+            // Safety: if reset the value failed, get() can see staled value if
```
jbachorik: **[LOW · consistency]** The comment here says "get() can see staled value if it is freed", but in this code path the value is explicitly **not** freed (`F` is not called when `pthread_setspecific` fails). The value is stale but not freed; the comment inverts the actual safety invariant.

Suggestion:
```cpp
// Safety: only free after the slot is cleared; if the slot reset failed,
// skip the free to avoid get() returning a dangling pointer.
```

ddprof-lib/src/main/cpp/threadLocal.h
jbachorik, line 81: **[MEDIUM · completeness]** `set()` does not invoke `F` (CLEAN_FUNC) on the existing TSD value before overwriting it. If `get()` auto-created an object via `C()` and a caller subsequently calls `set()` with a different pointer, the original allocation is leaked — POSIX does not call destructors on `pthread_setspecific` overwrite, only on thread exit.

Suggestion: in `set()`, retrieve the current value first and call `F` on it if non-null:
```cpp
void set(T value) {
    void* prev = pthread_getspecific(_key);
    int err = pthread_setspecific(_key, (const void*)value);
    assert(err == 0);
    if (prev != nullptr && F != nullptr) F(prev);
}
```
Alternatively, document that `set()` must only be called when the TSD slot is empty (i.e. after `clear()`) and enforce with an assert.

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,142 @@
...
+    ~ThreadLocal() {
+        pthread_key_delete(_key);
+    }
+
+    // double <--> u64 cast, preserve bit format
+    // Can use std::bit_cast after upgrade C++ version to 20
+    void set(double value) {
+        u64 val;
+        memcpy(&val, &value, sizeof(value));
+        int err = pthread_setspecific(_key, (const void*)val);
```
jbachorik: **[LOW · consistency]** Integer-to-pointer cast at line 120 uses a C-style `(const void*)val` while the reverse pointer-to-integer cast at line 130 uses a C-style `(u64)p`. Both should use `reinterpret_cast` for clarity and to suppress potential `-Wint-to-pointer-cast` warnings:
```cpp
// line 120:
int err = pthread_setspecific(_key, reinterpret_cast<const void*>(val));
// line 130:
u64 val = reinterpret_cast<u64>(p);
```

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,142 @@
...
+            return 0.0;
+        }
+
+        u64 val = (u64)p;
+        double value;
+        memcpy(&value, &val, sizeof(val));
+        return value;
+    }
+
+    void clear() {
```
jbachorik, line 161: **[MEDIUM · robustness]** `ThreadLocal<double>::clear()` uses `assert(err == 0)` for the `pthread_setspecific` return value; under NDEBUG the assert is stripped. If `pthread_setspecific` fails silently, the TSD slot retains the stale encoded double bits; the next `get()` sees a non-null value and decodes garbage instead of returning 0.0.

Suggestion: mirror the defensive pattern in the generic `clear()` — read the existing value first, attempt the set, and only proceed if it succeeded — or use `abort()` instead of `assert` so the failure is not silently swallowed in release builds.

ddprof-lib/src/main/cpp/livenessTracker.cpp
```diff
@@ -376,7 +404,7 @@ void LivenessTracker::track(JNIEnv *env, AllocEvent &event, jint tid,
       env->DeleteWeakGlobalRef(ref);
     }
   }
-  skipped = 0; // reset the subsampling skipped bytes
+  skipped.set(0); // reset the subsampling skipped bytes
```
jbachorik: **[LOW · correctness]** (Pre-existing, not introduced by this PR) `skipped.set(0)` resets the accumulated skipped-bytes counter even when the current event was dropped (table full), losing weight that should carry forward to the next successfully recorded sample.

Suggestion: move `skipped.set(0)` inside the `idx < _table_cap` branch so it only resets when the sample was actually recorded.

ddprof-lib/src/test/cpp/threadLocal_ut.cpp
```diff
@@ -0,0 +1,237 @@
...
+TEST_F(ThreadLocalTest, Generic_UnsetIsZero) {
+    intptr_t observed = -1;
+    std::thread t([&] { observed = g_int_tl.get(); });
+    t.join();
+    EXPECT_EQ(0, observed);
+}
+
+// The create function lazily initializes storage on first get() and is invoked
+// exactly once per thread; subsequent get()s return the same pointer.
+TEST_F(ThreadLocalTest, Lazy_CreateOncePerThread) {
```
jbachorik, line 124: **[MEDIUM · completeness]** The `Lazy_*` tests share `g_tracked_tl`, `g_created`, and `g_freed` as namespace-scope statics. Counts are reset per test but the TSD key itself is shared. In parallel test runs or with a thread-pool GTest runner where worker threads are reused, freed counts can be non-deterministic.

Suggestion: use a per-test-local `ThreadLocal` instance (or scoped key creation/deletion in `SetUp`/`TearDown`) so each test owns its own TSD key and counters, eliminating cross-test contamination regardless of thread reuse.

ddprof-lib/src/test/cpp/threadLocal_ut.cpp
```diff
@@ -0,0 +1,237 @@
...
+        // Read the payload here: free_tracked() deletes it on thread exit, so
+        // dereferencing first/second after join() would be use-after-free.
+        value = *first;
+    });
+    t.join();
+
+    ASSERT_NE(nullptr, first);
+    EXPECT_EQ(first, second);          // same instance reused (pointer compare only)
+    EXPECT_EQ(1234, value);            // created via create_tracked()
+    EXPECT_EQ(1, g_created.load());    // created exactly once
```
jbachorik, line 143: **[LOW · completeness]** `Lazy_CreateOncePerThread` resets `g_freed` to 0 but never asserts its value after the thread joins. A stub implementation that omits `CLEAN_FUNC` entirely would still pass this test.

Suggestion: add `EXPECT_EQ(1, g_freed.load());` after the `g_created` assertion to verify the TSD destructor is called on thread exit.

ddprof-lib/src/test/cpp/threadLocal_ut.cpp
```diff
@@ -0,0 +1,237 @@
...
+TEST_F(ThreadLocalTest, Double_UnsetIsZero) {
+    double observed = -1.0;
+    std::thread t([&] { observed = g_double_tl.get(); });
+    t.join();
+    EXPECT_EQ(0.0, observed);
+}
+
+// Per-thread accumulation mirrors LivenessTracker's `skipped` usage: each thread
+// keeps its own running sum, isolated from the others.
+TEST_F(ThreadLocalTest, Double_PerThreadAccumulation) {
```
jbachorik, line 270: **[LOW · completeness]** The test suite has zero coverage for `clear()` on either `ThreadLocal` specialization. A stub `clear()` that did nothing would pass all 8 tests without failure.

Suggestion: add at minimum: (1) a test for generic `clear()` verifying the stored pointer is zeroed and the destructor ran; (2) a test for `ThreadLocal<double>::clear()` verifying `get()` returns `0.0` afterwards; (3) a test for `clear()` when `F == nullptr` verifying the slot is actually nulled.

ddprof-lib/src/main/cpp/threadLocal.h
zhengyu123: I don't think we should abort user application. I think we should have a mechanism to gracefully shutdown profiling in circumstances like this one.

Given hotspot JVM also only `assert` the [failure]( https://github.com/openjdk/jdk/blob/master/src/hotspot/os/posix/threadLocalStorage_posix.cpp#L48) I don't see it is high priority one.

ddprof-lib/src/main/cpp/threadLocal.h
zhengyu123: Commented above

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,142 @@
...
+    pthread_key_t _key;
+ 
+public:
+    ThreadLocal(const ThreadLocal&) = delete;
+    ThreadLocal& operator=(const ThreadLocal&) = delete;
+
+    ThreadLocal() {
+        int err = pthread_key_create(&_key, F);
+        // What to do if we can not create a key?
+        assert(err == 0);
```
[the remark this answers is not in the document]
zhengyu123: As commented below, we should shutdown profiling gracefully - defer to separate PR.

ddprof-lib/src/main/cpp/threadLocal.h
zhengyu123: See other comments

ddprof-lib/src/main/cpp/threadLocal.h
zhengyu123: Added document. We want to revisit this API when other usage pattern emerged.

ddprof-lib/src/main/cpp/livenessTracker.cpp
zhengyu123: Done

ddprof-lib/src/main/cpp/threadLocal.h
zhengyu123: Done

ddprof-lib/src/test/cpp/threadLocal_ut.cpp
zhengyu123: Done

ddprof-lib/src/test/cpp/threadLocal_ut.cpp
zhengyu123: Done

ddprof-lib/src/main/cpp/livenessTracker.cpp
```diff
@@ -322,7 +350,7 @@ void LivenessTracker::track(JNIEnv *env, AllocEvent &event, jint tid,
     _table[idx].time = TSC::ticks();
     _table[idx].ref = ref;
     _table[idx].alloc = event;
-    _table[idx].skipped = skipped;
+    _table[idx].skipped = skipped.get();
```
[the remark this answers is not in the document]
zhengyu123: Fixed

ddprof-lib/src/main/cpp/threadLocal.h
zhengyu123: Explained

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,150 @@
...
+        int err = pthread_key_create(&_key, F);
+        // What to do if we can not create a key?
+        // We probably want to shutdown profiler gracefully, instead of
+        // aborting user application - We will need this mechanism globally,
+        // defer to a separate task.
+        assert(err == 0);
+    }
+
+    ~ThreadLocal() {
+        pthread_key_delete(_key);
```
rkennke: **[HIGH · robustness]** Re-raising this independently of the `pthread_key_create` assert discussion, because the destructor fix stands on its own regardless of how we decide to handle key-creation failure.

The destructor calls `pthread_key_delete(_key)` unconditionally. `_key` is only valid if `pthread_key_create` succeeded — and under `NDEBUG` the constructor's `assert(err == 0)` is compiled out, so a failed creation leaves `_key` at its zero-initialized value. `pthread_key_delete(0)` then deletes a key this object never owned. On musl, key 0 is legitimately held by another component (`thread.cpp:91-93`) — i.e. exactly the TLS-aliasing failure mode this PR exists to fix.

This is cheap and self-contained: track creation success and guard both the destructor and the accessors, without requiring the graceful-shutdown mechanism we're deferring.

```cpp
    bool _key_valid;
...
    ThreadLocal() {
        static_assert(sizeof(T) == sizeof(void*), ...);
        _key_valid = (pthread_key_create(&_key, F) == 0);
        assert(_key_valid);
    }

    ~ThreadLocal() {
        if (_key_valid) pthread_key_delete(_key);
    }
```

Please apply the same guard to the `ThreadLocal<double>` specialization, whose destructor has the identical issue.

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    /**
+     * set(nullptr) will result in the value being recreated when get() is called
+     * when CREATE_FUNC is not nullptr
+     */
+    void set(T value) {
+        int err = pthread_setspecific(_key, (const void*)value);
```
rkennke: **[MEDIUM · completeness]** `set()` overwrites the TSD slot without freeing the value already stored there. POSIX does not run the key's destructor on a `pthread_setspecific` overwrite — only on thread exit — so if `get()` lazily created an object via `C()` and a caller later `set()`s a different pointer, the original allocation leaks until the thread dies.

No current caller hits this (the lazy-create instances `gen`/`dis` are only ever read via `get()`), so this isn't gating today. But the PR description states `ProfiledThread` and others will be converted to `ThreadLocal` in follow-ups, and this is a silent leak waiting for the first `get()`-then-`set()` user.

Either free the previous value on overwrite:

```cpp
    void set(T value) {
        void* prev = pthread_getspecific(_key);
        int err = pthread_setspecific(_key, (const void*)value);
        assert(err == 0);
        if (prev != nullptr && F != nullptr) F(prev);
    }
```

…or, if you'd rather keep `set()` minimal, document that `set()` must only be called on an empty slot (i.e. after `clear()`) so the contract is explicit before the future conversions land.

ddprof-lib/src/main/cpp/threadLocal.h
zhengyu123: Well, if the key is invalid, it blows up before destructor. Besides, `assert(_key_valid)` and `if (_key_valid)` checks are just not consistent.  ... I don't mind adding check to make AI happy.

ddprof-lib/src/main/cpp/threadLocal.h
zhengyu123: It is not obvious who should be the owner of the previous value, so add a comment - caller is the owner of previous value.

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,162 @@
...
+        _key_valid = pthread_key_create(&_key, F) == 0;
+        // What to do if we can not create a key?
+        // We probably want to shutdown profiler gracefully, instead of
+        // aborting user application - We will need this mechanism globally,
+        // defer to a separate task.
+        assert(_key_valid);
+    }
+
+    ~ThreadLocal() {
+        assert(_key_valid);
```
rkennke: I would prefer to do a if (_key_valid) check here, which does the right thing in production code, and add an assert in the else branch.

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,162 @@
...
+        // Only support 64-bit platforms, double and void* are the same size
+        static_assert(sizeof(void*) == 8);
+        static_assert(sizeof(double) == 8);
+        _key_valid = pthread_key_create(&_key, nullptr) == 0;
+        // What to do if we can not create a key?
+        assert(_key_valid);
+    }
+
+    ~ThreadLocal() {
+        assert(_key_valid);
```
rkennke: Same here.

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * set(nullptr) will result in the value being recreated when get() is called
+     * when CREATE_FUNC is not nullptr.
+     * Note: caller is responsible to free old value.
+     */
+    void set(T value) {
+        assert(_key_valid);
+        int err = pthread_setspecific(_key, (const void*)value);
```
rkennke: Minor: use C++ style cast here.

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,162 @@
...
+        pthread_key_delete(_key);
+    }
+
+    // double <--> u64 cast, preserve bit format
+    // Can use std::bit_cast after upgrade C++ version to 20
+    void set(double value) {
+        assert(_key_valid);
+        u64 val;
+        memcpy(&val, &value, sizeof(value));
+        int err = pthread_setspecific(_key, (const void*)val);
```
rkennke: Minor: Use C++ style cast here.

rkennke (changes requested):
There's still the destructor checks that I would like to have added in production code path, plus two (minor) C style casts that should be C++ casts.

ddprof-lib/src/main/cpp/threadLocal.h
zhengyu123: okay

ddprof-lib/src/main/cpp/threadLocal.h
zhengyu123: Done

ddprof-lib/src/main/cpp/threadLocal.h
zhengyu123: oaky

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,164 @@
...
+        // We probably want to shutdown profiler gracefully, instead of
+        // aborting user application - We will need this mechanism globally,
+        // defer to a separate task.
+        assert(_key_valid);
+    }
+
+    ~ThreadLocal() {
+        if(_key_valid) {
+            pthread_key_delete(_key);
+        }
```
rkennke, line 73: Ok good. are you also adding the else { assert ... } part?

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,164 @@
...
+        static_assert(sizeof(double) == 8);
+        _key_valid = pthread_key_create(&_key, nullptr) == 0;
+        // What to do if we can not create a key?
+        assert(_key_valid);
+    }
+
+    ~ThreadLocal() {
+        if(_key_valid) {
+            pthread_key_delete(_key);
+        }
```
rkennke, line 135: same here: are you also adding the else { assert ... } part?

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    double get() {
+        assert(_key_valid);
+        void* p = pthread_getspecific(_key);
+        if (p == nullptr) {
+            return 0.0;
+        }
+
+        u64 val = (u64)p;
```
rkennke: here is another C-style cast that should be `reinterpret_cast<u64>(p)`.

ddprof-lib/src/main/cpp/livenessTracker.cpp
```diff
@@ -276,6 +277,29 @@ Error LivenessTracker::initialize(Arguments &args) {
   return _stored_error = Error::OK;
 }
 
+static void* create_mt19937() {
+  // std::mt19937 itself is noexcept, but std::random_device and `new` may throw.
+  // If that happens we let the failure terminate the process (same outcome as
+  // failing thread_local initialization previously).
+  return (void*)(new std::mt19937(std::random_device{}()));
```
rkennke: Could use static_cast<void*>( .. ) here.

ddprof-lib/src/main/cpp/livenessTracker.cpp
```diff
@@ -276,6 +277,29 @@ Error LivenessTracker::initialize(Arguments &args) {
...
+}
+
+static void* create_uniform_real_distribution() {
+  // std::uniform_real_distribution<> construction is noexcept, but `new` may throw.
+  // If allocation fails the process is likely to abort anyway.
+  return (void*)(new std::uniform_real_distribution<>(0, 1.0));
```
rkennke: And here

ddprof-lib/src/main/cpp/livenessTracker.cpp
```diff
@@ -276,6 +277,29 @@ Error LivenessTracker::initialize(Arguments &args) {
...
+}
+
+static void free_mt19937(void* p) {
+  std::mt19937* mt = (std::mt19937*)p;
```
rkennke: and here

ddprof-lib/src/main/cpp/livenessTracker.cpp
```diff
@@ -276,6 +277,29 @@ Error LivenessTracker::initialize(Arguments &args) {
...
+  delete mt;
+}
+
+static void free_uniform_real_distribution(void* p) {
+  std::uniform_real_distribution<>* urd = (std::uniform_real_distribution<>*)p;
```
rkennke: and here

rkennke (approved):
Thanks for making the changes. I'm approving this now.
You might want to add the else { assert .. } parts in the destructors, and I also did a quick pass and found more places where C-style casts could be done with C++ casts. I leave that up to you.

ddprof-lib/src/main/cpp/threadLocal.h
zhengyu123: okay

ddprof-lib/src/main/cpp/livenessTracker.cpp
zhengyu123: oaky

PR #603, opened 2026-06-18
fix(jfr): skip emitting empty constant pools

jbachorik:
**What does this PR do?**:

Stops the JFR checkpoint from emitting empty (0-element) constant-pool sections. Each variable pool writer (`writeThreads`, `writeStackTraces`, `writeMethods`, `writeClasses`, `writePackages`, and the three `writeConstantPoolSection` overloads) now skips emission and returns 0 when its pool is empty. `writeCpool` writes a 1-byte placeholder pool-count and sums the actually-emitted pools; `finishChunk` back-patches the real count via `pwrite` after flush, mirroring the existing cpool size patch. `writeLogLevels` is moved up alongside the other always-present constant pools.

**Motivation**:

OpenJDK `jfr` rejects a checkpoint that declares a constant pool with zero elements:

```
jfr print: unexpected internal error, Pool java.lang.String must contain at least one element
java.lang.InternalError: Pool java.lang.String must contain at least one element
```

The previous code hard-coded the pool count to `12` and always emitted every pool, so a chunk produced before any string label / context attribute / package existed contained an empty pool and was unreadable by these tools.

**Additional Notes**:

Stacked on #602; rebase to `main` after it merges. The diff here is only the empty-pool skipping + flush-safe count back-patch (Set B); the method-pool id allocator lives in #602.

The count cannot be precomputed: the Method/Class/Package/Symbol pools are only fully populated as a side effect of `writeStackTraces`/`writeMethods`, and empty variable pools are skipped. Hence the placeholder + back-patch.

**How to test the change?**:

Validated with OpenJDK `jfr print` on recordings produced by the `MethodIdReuseTest` dump driver.

Before (build without this change):

```
$ jfr print --events 'datadog.*' prof15130-reuse-p2-6-….jfr
jfr print: unexpected internal error, Pool java.lang.String must contain at least one element
java.lang.InternalError: Pool java.lang.String must contain at least one element
(exit 1)
```

After (this change):

```
$ jfr print --events 'datadog.*' prof15130-reuse-….jfr
(exit 0, no pool error)
```

Local results (debug build, macOS):

- `:ddprof-lib:buildDebug -Pskip-gtest` → BUILD SUCCESSFUL
- `:ddprof-lib:gtestDebug_methodMapId_ut` → `[ PASSED ] 5 tests` (no regression)
- `jfr print --events 'datadog.*'` on a produced recording → exit 0 (was exit 1 / `must contain at least one element` before)

**For Datadog employees**:

- [x] JIRA: [PROF-15130](https://datadoghq.atlassian.net/browse/PROF-15130)

[PROF-15130]: https://datadoghq.atlassian.net/browse/PROF-15130?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -761,7 +761,8 @@ off_t Recording::finishChunk(bool end_recording, bool do_cleanup) {
   }
 
   off_t cpool_offset = lseek(_fd, 0, SEEK_CUR);
-  writeCpool(_buf);
+  int count_offset_in_cpool = 0;
+  int pool_count = writeCpool(_buf, &count_offset_in_cpool);
   flush(_buf);
 
   off_t cpool_end = lseek(_fd, 0, SEEK_CUR);
@@ -769,6 +770,13 @@ off_t Recording::finishChunk(bool end_recording, bool do_cleanup) {
   // Patch cpool size field
   _buf->putVar32(0, cpool_end - cpool_offset);
   ssize_t result = pwrite(_fd, _buf->data(), 5, cpool_offset);
+  (void)result;
+
+  // Patch the constant pool count placeholder (written as a 1-byte put8 in
+  // writeCpool). Done flush-safe via pwrite to the FILE offset, mirroring the
+  // size patch above: _buf has been flushed/reset, so _buf->data() is scratch.
+  _buf->put8(0, (char)pool_count);
+  result = pwrite(_fd, _buf->data(), 1, cpool_offset + count_offset_in_cpool);
   (void)result;
 
   off_t chunk_end = lseek(_fd, 0, SEEK_CUR);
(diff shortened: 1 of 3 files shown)


jbachorik:
@codex review

jbachorik:

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -1474,6 +1495,10 @@ void Recording::writeThreads(Buffer *buf) {
     _thread_ids[i][old_index].clear();
   }
 
+  if (threads.empty()) {
```
rkennke: This is effectively dead code, right? Because we do `threads.insert(_tid) ` just a few lines up. This also means we have 5 non-empty cpools, not 4.

rkennke (approved):
One minor thing, up to you to fix or not.
What might be useful (maybe as a follow-up?) is a regression test. A gtest that produces a chunk with empty variable pools and asserts (a) the back-patched count equals the number of emitted sections and (b) no empty section is present would be sufficient.

PR #605, opened 2026-06-18
fix(ci): lower aarch64 benchmark memory request to 128Gi cluster limit

jbachorik:
**What does this PR do?**:
Lowers `KUBERNETES_MEMORY_REQUEST` and `KUBERNETES_MEMORY_LIMIT` for `benchmarks-candidate-aarch64` from 200Gi to 128Gi.

**Motivation**:
All benchmark jobs are failing with:
> Preparation failed: MemoryRequest of 200Gi exceeds the allowed limit of 128Gi.

The 200Gi value was carried over from the original CI migration and presumably reflected the old cluster's limits. The arm64 runner pool now enforces a 128Gi cap.

**Additional Notes**:
The amd64 job has no explicit memory override so it is unaffected.

**How to test the change?**:
Benchmark jobs should no longer fail at preparation. No functional change to benchmark behaviour.

**For Datadog employees**:
- [ ] JIRA: N/A

### DIFF
--- a/.gitlab/benchmarks/.gitlab-ci.yml
+++ b/.gitlab/benchmarks/.gitlab-ci.yml
@@ -76,8 +76,8 @@ benchmarks-candidate-aarch64:
   tags: ["arch:arm64"]
   image: $BENCHMARK_IMAGE_ARM64
   variables:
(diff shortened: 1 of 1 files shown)


jbachorik:
This needs to be fixed differently.

PR #607, opened 2026-06-18
fix(ci): migrate benchmarks to benchmarking-platform trigger

jbachorik:
**What does this PR do?**:
Replaces the broken inline benchmark jobs with a benchmarking-platform bridge trigger. The old jobs ran the profiler inline via shell scripts and uploaded results to S3 themselves; the new approach delegates entirely to the BP pipeline at `DataDog/apm-reliability/benchmarking-platform@java-profiler`.

**Motivation**:
The existing benchmark CI was broken and unmaintained. The benchmarking-platform provides a reliable, standardised infra for running and tracking reliability benchmarks.

**Additional Notes**:
- `benchmarks-trigger` is a GitLab bridge job — it cannot appear in `needs:` of downstream jobs, so `post-benchmarks-pr-comment` and `publish-benchmark-gh-pages` now run in a `post-benchmarks` stage ordered after `benchmarks`.
- `download-bp-reports.sh` (renamed from `download-s3-reports.sh`) fetches BP job artifacts via the GitLab CI API — no AWS CLI or S3 access required.
- The `images.yml` include is removed — the BP side manages its own Docker image build.
- `jb/bench-memory-limit-fix` is superseded by this PR (the memory-heavy aarch64 jobs are gone).
- The BP side implementation is in `DataDog/benchmarking-platform` PR #190.

**How to test the change?**:
Push to a branch and verify the GitLab pipeline shows a `benchmarks-trigger` bridge job that fires the BP downstream pipeline. The `post-benchmarks-pr-comment` and `publish-benchmark-gh-pages` jobs should run in the `post-benchmarks` stage after the trigger completes.

**For Datadog employees**:
- [ ] JIRA: [JIRA-XXXX]


.gitlab/benchmarks/download-bp-reports.sh
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env bash
+# Downloads result JSONs from the BP downstream pipeline via the GitLab CI API.
```
rkennke, line 2: Naming nit: the file is called `download-s3-reports.sh` but it fetches via the **GitLab CI API**, not S3 (the final commits dropped `aws s3 sync`). The header comment here is accurate; consider renaming the file (and the PR description, which still says "S3") to avoid confusion.

.gitlab/benchmarks/download-s3-reports.sh
```diff
@@ -0,0 +1,97 @@
...
+# ── helper: curl with explicit HTTP status checking ──────────────────────────
+# Usage: api_get <url> <output_file>
+# Returns 0 on 2xx, prints diagnostics and returns 1 otherwise.
+api_get() {
+  local url="$1" out="$2"
+  local http_code
+  http_code=$(curl -s -o "${out}" -w "%{http_code}" \
+    --header "JOB-TOKEN: ${CI_JOB_TOKEN}" "${url}")
+  if [[ "${http_code}" != 2* ]]; then
+    echo "  API ${url##*/}: HTTP ${http_code} — $(cat "${out}" | python3 -c "import sys,json; d=json.load(sys.stdin); print(d.get('message','?'))" 2>/dev/null || echo 'see above')"
```
rkennke: Minor: useless use of `cat` — `python3 ... < "${out}"` (or pass the filename) avoids the extra process.

.gitlab/benchmarks/download-s3-reports.sh
```diff
@@ -0,0 +1,97 @@
...
+  echo "benchmarks-trigger bridge not found or did not run — skipping download"
+  exit 0
+fi
+echo "BP downstream pipeline: project=${BP_PROJECT_ID}  pipeline=${DOWNSTREAM_PIPELINE_ID}"
+
+# ── 2. list jobs in the downstream pipeline ──────────────────────────────────
+JOBS_FILE="${TMPDIR_LOCAL}/jobs.json"
+echo "Listing BP pipeline jobs…"
+if ! api_get \
+  "${CI_API_V4_URL}/projects/${BP_PROJECT_ID}/pipelines/${DOWNSTREAM_PIPELINE_ID}/jobs?per_page=100" \
```
rkennke: No pagination: `per_page=100` with no follow-up loop. If a BP pipeline ever exceeds 100 jobs, artifacts from later jobs are silently dropped, yielding fewer/zero reports. With `allow_failure: true` this won't fail the pipeline, just confusingly under-publish. Consider a pagination loop, or at minimum log the job count so a truncation is visible.

.gitlab/benchmarks/download-bp-reports.sh
```diff
@@ -0,0 +1,97 @@
...
+# ── 3. download result_*.json from each job's artifact zip ───────────────────
+DOWNLOADED=0
+for JOB_ID in ${JOB_IDS}; do
+  ART_ZIP="${TMPDIR_LOCAL}/art_${JOB_ID}.zip"
+  ART_STATUS=$(curl -s -o "${ART_ZIP}" -w "%{http_code}" \
+    --header "JOB-TOKEN: ${CI_JOB_TOKEN}" \
+    "${CI_API_V4_URL}/projects/${BP_PROJECT_ID}/jobs/${JOB_ID}/artifacts" 2>/dev/null)
+  if [[ "${ART_STATUS}" == 2* ]]; then
+    # -j: junk paths (strip artifacts/ prefix), -q: quiet, -o: overwrite
+    if unzip -q -j "${ART_ZIP}" "artifacts/result_*.json" -d "${DEST}/" 2>/dev/null; then
```
rkennke, line 98: Tight coupling to BP's artifact layout: hard-codes the `artifacts/` prefix and `result_*.json` naming. If BP changes either, this silently extracts nothing and the script exits 1 with only a WARNING. Acceptable given `allow_failure`, but worth documenting this as a contract on the BP side (#190) so a rename there doesn't quietly break publishing here.

rkennke (commented):
Reviewed the benchmark migration — direction is clearly right and the GitLab bridge mechanics (`strategy: depend`, `interruptible: false`, moving the `CANCELLED` gate into `rules:` since bridges can't have `before_script`, and the project-wide `resource_group: gh-pages-publish` to serialize force-pushes) are all handled correctly and well-commented. Inline comments below are minor/cleanup. Two findings can't be anchored inline because they touch files outside this diff:

**1. (verify before merge) `mem_watch.sh` deletion vs. reliability check.** This PR deletes `.gitlab/benchmarks/steps/mem_watch.sh`, but `.gitlab/reliability/memory_trend_check.sh:58` still calls `sh ./benchmarks/steps/mem_watch.sh ...`. The path differs by a `.gitlab/` prefix and is relative to the job CWD, so it may resolve to a checked-out benchmarking-platform copy rather than the deleted file — but please confirm. If it resolves to the deleted file, `set +e` at the top of that script means it fails **silently** and you lose memory-trend logging with no error.

**2. (follow-up) Orphaned image config.** `images.yml` is correctly left in place (still included by the reliability pipelines), but `BENCHMARK_IMAGE_AMD64/ARM64` and the `*-benchmarks` entries in `.gitlab/scripts/check-image-updates.sh` and `rebuild-images.sh` now build/track images no benchmark job consumes. Not breaking; worth a cleanup pass.

Non-blocking overall — I'd approve after confirming #1.

jbachorik:
@rkennke Regarding mem_watch.sh deletion - it was indeed overzealous, the liveness detection failed. The script needs to still be there. Restored from backup.

rkennke (approved):
Looks good to me, thank you!

jbachorik:
@rkennke thanks for the patient review!

PR #614, opened 2026-06-23
[PROF-15201] fix(profiler): close two TOCTOU races between SIGPROF handler and JFR lifecycle

r1viollet:
**What does this PR do?**

Fixes the profiler's start/stop sequences against SIGPROF/SIGVTALRM signal handlers writing JFR. Two TOCTOU races between engine `_enabled` state and JFR lifecycle transitions could cause use-after-free or hangs during the 60-second recording cycle rotation. The fix reworks the shutdown around three primitives.

**1. `_enabled` — unified RELEASE/ACQUIRE across every engine**

Every JFR-writing signal handler now reads `_enabled` with `__ATOMIC_ACQUIRE`, and every `enableEvents()` writes it with `__ATOMIC_RELEASE`. Previously `ITimer` and `PerfEvents` used plain `volatile bool` with no ordering guarantees — on weakly ordered arm64, a newly delivered signal could observe stale `_enabled=true` after `disableEngines()` had already returned, letting a handler enter `recordSample()` while stop was tearing JFR down.

Handlers covered: `CTimer`, `CTimerJvmti`, `ITimer`, `ITimerJvmti`, `PerfEvents`, `WallClockASGCT`, `WallClockJvmti`.

**2. Ordering — drain happens right after `disableEngines()`, before every other engine stop**

`Profiler::stop()` now runs:

1. `disableEngines()` — RELEASE store of `_enabled=false` on every engine
2. `SignalInflight::drain()` — wait for in-flight handlers to leave their JFR-write paths
3. Individual engine `stop()` calls (wall pthread join, CTimer per-thread timer teardown, etc.)
4. `_jfr.stop()`

`Profiler::start()` is reordered symmetrically. `_wall_engine->enableEvents(true)` must happen **before** `_wall_engine->start()` because the wall pthread checks `_enabled` once at startup and exits if false. `_cpu_engine->enableEvents(true)` is deferred until **after** `_jfr.start()` returns so no SIGPROF handler ever runs against partially-initialised JFR structures.

**3. Drain — `SignalInflight`, a new module tracking JFR-writing handlers**

Every JFR-writing signal handler constructs an `InflightGuard` (RAII) at entry. The guard increments a cache-line-aligned counter on construction and decrements it on destruction, so it participates in the drain regardless of which exit path the handler takes. `SignalInflight::drain()` spins until the counter reaches zero or a 200 ms deadline elapses.

- Drain succeeds → `Profiler::stop()` proceeds through the rest of teardown.
- Drain times out (pathological, e.g. a stuck handler) → `Profiler::stop()` returns non-OK `Error` and leaves `_state == RUNNING` so the caller cannot `start()` against half-torn-down JFR. `disableEngines()` is idempotent and no engine `stop()` has run yet, so a retry-`stop()` is safe.

**Motivation**

Intermittent SIGSEGV (exit 139) and hang failures in DataDog/profiling-backend CI. Bisected to a dd-trace-java commit that changed instrumentation initialization timing, shifting *when* the 60-second recording cycle boundary fell relative to test thread activity — exposing the races reliably.

**How to test the change?**

Controlled reproducer in DataDog/profiling-backend using `AnalysisEndpointTest.testResourceExhausted` against a patched `libjavaProfiler.so`:

- Without fix: ~60% failure rate per iteration (SIGSEGV / hang)
- With fix: 12/12 iterations clean against `v_1.44.0` baseline

Local coverage of the shutdown paths: `ShutdownTest` (start/stop across CPU + wall combinations), `SmokeWallTest`, `WallclockDumpSmokeTest`, `SleepTest`, `CollapsingSleepTest`, `JavaProfilerTest.vmStackwalkerCrashRecoveryTest` (SIGSEGV-during-stackwalk longjmp path).

**Additional Notes**

- **Timeout path**: on drain timeout, `_jfr.stop()` is skipped (a bounded ~few-MB leak) but no use-after-free is possible. State stays `RUNNING`; caller retries. The idempotent-retry property is why drain sits between `disableEngines()` and any other engine `stop()` — nothing non-idempotent (`pthread_join`, `timer_delete`) has run yet.
- **Documented longjmp limitation** (`signalInflight.h`): J9's SIGSEGV null-pointer-check handler can `siglongjmp` past an `InflightGuard` frame, leaking the counter by +1. Doesn't fire from profiler handlers in practice — J9 only claims faults at annotated null-check sites, and our stack-walker faults land in our own C++ code. If it did, subsequent `stop()`s would hit the timeout — graceful degradation, no crash. Same limitation, same precedent as `SignalHandlerScope` (`guards.h`).
- **Follow-ups tracked in [PROF-15268](https://datadoghq.atlassian.net/browse/PROF-15268)**:
  - Per-thread `_inflight` counter (removes cross-core cache-line traffic on the counter, closes the J9 longjmp leak by mirroring `SIGNAL_HANDLER_GUARD_RELEASE`).
  - Stop the wall pthread before draining (removes tgkill churn during the drain spin; requires either idempotent engine `stop()` or a partial-teardown flag in `Profiler::stop()`).
  - Performance looks untouched, following the fixes on the enabled flag for wall events. However **if we see a 200ms flake** , we can assume that this PR is responsible for it.

**For Datadog employees**
- [x] JIRA: [PROF-15201](https://datadoghq.atlassian.net/browse/PROF-15201)

[PROF-15268]: https://datadoghq.atlassian.net/browse/PROF-15268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


r1viollet:
@codex review

r1viollet:
I'm worried about perf of such an approach.

r1viollet:
@codex review

ddprof-lib/src/main/cpp/ctimer_linux.cpp
```diff
@@ -176,10 +178,48 @@ Error CTimer::start(Arguments &args) {
...
+    struct timespec now;
+    clock_gettime(CLOCK_MONOTONIC, &now);
+    if (now.tv_sec > deadline.tv_sec ||
+        (now.tv_sec == deadline.tv_sec && now.tv_nsec >= deadline.tv_nsec)) {
+      Log::warn("CTimer::drainInflight: timed out after %ldms waiting for "
+                "%d in-flight SIGPROF handler(s); proceeding with JFR teardown. "
+                "This may indicate a stuck signal handler.",
+                DRAIN_TIMEOUT_NS / 1000000L,
+                __atomic_load_n(&_inflight, __ATOMIC_ACQUIRE));
+      break;
```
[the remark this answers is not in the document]
r1viollet: makes sense

r1viollet:
@codex review

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -1558,11 +1571,22 @@ Error Profiler::stop() {
     }
   }
 
+  // Wait for all SIGPROF handlers to drain before tearing down JFR.
+  // If draining times out (handlers are stuck), skip JFR teardown to prevent
+  // use-after-free. This leaks JFR resources (~few MB) but is far safer than
+  // freeing structures that stuck handlers are still accessing.
+  bool drained = CTimer::drainInflight();
```
jbachorik: `drainInflight()` only tracks `CTimer::_inflight`. When `selectCpuEngine()` returns `&perf_events` or `&itimer` (the standard fallback when `ctimer.check()` fails — e.g. kernels without `timer_create`, or `-Devent=itimer`), those handlers call `recordSample()` and write JFR buffers but never construct an `InflightGuard`. `_inflight` stays 0, so `drainInflight()` returns `true` immediately — `_jfr.stop()` can free JFR buffers while a live handler is still inside `recordSample()`. The TOCTOU/UAF this PR closes for CTimer remains open for the common fallback engines.

Fix options: (a) add `InflightGuard` to `PerfEvents::signalHandler`, `ITimer::signalHandler`, and `ITimerJvmti::signalHandler`; or (b) implement `virtual Engine::drainInflight()` and call `_cpu_engine->drainInflight()` so the drain dispatches to whichever engine is active.

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -1413,9 +1410,15 @@ Error Profiler::start(Arguments &args, bool reset) {
     }
   }
   if ((_event_mask & EM_WALL) && _wall_engine != &noop_engine) {
+    // Enable wall clock BEFORE starting its pthread to close a TOCTOU race:
+    // the pthread enters timerLoopCommon(), checks _enabled, and exits if false.
+    // Wall clock is safe to enable early because it doesn't access JFR in signal
```
jbachorik: The comment is a bit confusing - neither CPU nor Wallclock are using the built-in JFR. They both are writing JFR recording data from a signal handler.

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -1558,11 +1571,22 @@ Error Profiler::stop() {
...
+  // use-after-free. This leaks JFR resources (~few MB) but is far safer than
+  // freeing structures that stuck handlers are still accessing.
+  bool drained = CTimer::drainInflight();
+
   // writing these out before stopping the JFR recording allows to report the
   // correct counts in the recording
   _thread_info.reportCounters();
 
-  rotateDictsAndRun([&]{ _jfr.stop(); });
+  if (drained) {
```
jbachorik: When `drainInflight()` times out, `_jfr.stop()` is skipped but the profiler unconditionally sets `_state = IDLE` and returns `Error::OK`. The caller has no indication teardown was partial. A subsequent `start()` will call `_jfr.start()` against un-torn-down JFR structures while stuck handlers may still be live.

Consider returning a non-OK `Error` on the timeout path, or avoiding the `IDLE` transition, so callers know teardown was skipped and cannot blindly restart.

ddprof-lib/src/main/cpp/ctimer_linux.cpp
```diff
@@ -34,6 +34,7 @@
 #include <string.h>
 #include <sys/syscall.h>
 #include <time.h>
+#include <time.h>
```
jbachorik: Duplicate `#include <time.h>` — one can be removed.

ddprof-lib/src/main/cpp/ctimer_linux.cpp
```diff
@@ -176,10 +178,47 @@ Error CTimer::start(Arguments &args) {
...
+// short enough to avoid a perceptible hang if a handler is somehow stuck.
+static const long DRAIN_TIMEOUT_NS = 200000000L;
+
+bool CTimer::drainInflight() {
+  if (__atomic_load_n(&_inflight, __ATOMIC_ACQUIRE) == 0) {
+    return true; // fast path: nothing in flight
+  }
+
+  struct timespec deadline;
+  clock_gettime(CLOCK_MONOTONIC, &deadline);
```
jbachorik: `clock_gettime()` return value not checked; on failure `deadline` is uninitialized and the timeout loop behaviour is incorrect. Defensive fix: `if (clock_gettime(CLOCK_MONOTONIC, &deadline) != 0) { return false; }` (and similarly for the in-loop call).

ddprof-lib/src/main/cpp/ctimer.h
```diff
@@ -28,6 +28,20 @@ class CTimer : public Engine {
 protected:
   // This is accessed from signal handlers, so must be async-signal-safe
   static bool _enabled;
+
+public:
```
rkennke: Instead of making `_inflight` public, I would suggest to make it private, and provide accessors to it, e.g.:
```
  private:                                                                                                    
    alignas(64) static int _inflight;

  public:                                                                                                     
     static void enterHandler() { __atomic_fetch_add(&_inflight, 1, __ATOMIC_ACQUIRE); }
    static void exitHandler()  { __atomic_fetch_sub(&_inflight, 1, __ATOMIC_RELEASE); }
    static bool hasInflightHandlers() {
      return __atomic_load_n(&_inflight, __ATOMIC_ACQUIRE) > 0;
    }
 ```

And you also need to restore the `protected` access of the subsequent members, right now you're making a bunch of stuff `public` that used to be `protected`.

File: ddprof-lib/src/main/cpp/profiler.cpp
Comment (rkennke) [inline ddprof-lib/src/main/cpp/profiler.cpp]: This becomes dead code and can be removed (including its declaration).

File: ddprof-lib/src/main/cpp/guards.h [hunk 8]
```diff
@@ -102,6 +102,39 @@ class SignalHandlerScope {
 void signalHandlerUnwindAfterLongjmp();
 #define SIGNAL_HANDLER_UNWIND_AFTER_LONGJMP() signalHandlerUnwindAfterLongjmp()
 
+/**
```
rkennke: I would move that into ctimer.h. Combined with accessors, the whole thing collapses to header-only and several warts disappear:
  - guards.cpp loses the out-of-line ctor/dtor and the `#ifdef` include of ctimer.h — no more engine coupling in the generic guards module.                                                                                  
  - The `_active` bool goes away. It only existed to make the destructor a no-op on non-Linux; with separate Linux/non-Linux class definitions there's nothing to guard, so the member and its branch vanish.
  - `_inflight` stays private, accessed only through enterHandler/exitHandler/hasInflightHandlers/drainInflight, all co-located.

rkennke (changes requested):
Nice one. I have a few suggestions.

r1viollet:
Todo: check code paths for crashes and signal handlers. Could we hit a point where we forget to decrement a counter ?

ddprof-lib/src/main/cpp/profiler.cpp
r1viollet: This makes the PR more complex. I'll give it a shot to cover all variants.

r1viollet: Addressed in 1e751027 (rename/move) + 97e9f995 (extend coverage).

Two commits:

1. **1e751027** — moves the inflight counter and drain logic out of `CTimer` into a new `SignalInflight` module (header + impl). No behavior change; pure rename/move. The `InflightGuard` RAII type lives there too. `Profiler::stop()` now calls `SignalInflight::drain()`.

2. **97e9f995** — adds `InflightGuard` at handler entry (post foreign-signal filter, pre JFR-touching code) for the previously-uncovered handlers: `PerfEvents`, `ITimer`, `ITimerJvmti`, `WallClockASGCT::sharedSignalHandler`, `WallClockJvmti::sharedSignalHandler`. For wall, guarding at the dispatcher level transitively covers the inner per-engine `signalHandler` methods.

Went with the option-(a) extension — `InflightGuard` everywhere — rather than a virtual `Engine::drainInflight()`. The question 'can JFR be torn down right now?' is a single global question, not a per-engine question, so one shared counter + one drain site keeps it simple.

Risk-profile note: CTimer / PerfEvents are the high-probability paths (per-thread streams, many concurrent in-flight handlers). Wall is medium (up to ~reservoir_size handlers in flight from the last `tgkill` batch). ITimer is mostly theoretical (single in-flight max). All five plug into the same counter regardless, since the cost is one extra atomic per signal.

This is a larger surface than the original PR set out to fix — happy to split B2 (97e9f995) into a follow-up PR if you'd rather keep this one focused on the CTimer-specific fix. WDYT?

r1viollet:
**Reviewer call-out — longjmp leak limitation** (cc @jbachorik @rkennke)

While reviewing the new `SignalInflight` counter for ways it could go wrong, I noticed `InflightGuard`'s destructor can be bypassed by a longjmp that unwinds a signal handler frame. In this codebase the only realistic trigger is J9's SIGSEGV null-pointer-check handler: our `segvHandler` chains to J9 for unclaimed faults, and J9 may siglongjmp to a setjmp in normal Java code, unwinding past every active `InflightGuard`.

Documented in 8b069cbd (signalInflight.h header comment). Summary:

- **Probability**: very low. J9 only siglongjmps when the fault matches an annotated null-check site in Java bytecode. Our stack-walker faults are at addresses inside our own C++ code that J9 has no reason to claim — its handler should reject and let the next handler in the chain process the fault.
- **Failure mode if it ever fires**: counter pinned > 0 permanently. Every subsequent `Profiler::stop()` hits `drain()`'s 200 ms timeout, returns Error, and refuses to tear down JFR. The JVM keeps running; the profiler is wedged until process exit. **No crash, no use-after-free** — the timeout path that the counter exists to gate is exactly the safety net for this case.
- **Precedent**: `SignalHandlerScope` (guards.h) has the same limitation on the same J9 chain path for its `_signal_depth` counter, and the codebase accepts it. I followed that precedent rather than introducing a new mechanism.

Options if either of you considers this insufficient:

- **Watchdog**: track whether `_counter` moved during the drain spin; log a more pointed warning ("counter pinned at N — possible longjmp leak") when it didn't. ~10 lines, diagnostic only.
- **Per-thread counter**: move the counter into `ProfiledThread::_inflight`, zero it in `segvHandler` / `busHandler` before chaining (mirroring `SIGNAL_HANDLER_GUARD_RELEASE`). Actually fixes the leak, but `drain()` would iterate threads. ~30-50 lines plus a `ProfiledThread` field. Would also incidentally remove the global cache-line contention on the counter.

My lean is to ship with the documented limitation and revisit only if it actually fires in production. Happy to take either of the alternatives if you disagree.

r1viollet:
@codex review

ddprof-lib/src/main/cpp/itimer.cpp
```diff
@@ -38,6 +39,7 @@ void ITimer::signalHandler(int signo, siginfo_t *siginfo, void *ucontext) {
   // is therefore vulnerable to the foreign-SIGPROF deadlock scenario this
   // feature addresses. Use CTimer (the default) when signal-origin
   // validation is required.
+  InflightGuard inflight;
   if (!_enabled)
```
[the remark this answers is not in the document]
r1viollet: I agree.

rkennke (approved):
Looks good now. Maybe worth to re-run the benchmarks to check if the renaissance regression is just benchmark noise (likely the case, renaissance is not exactly very stable).

ddprof-lib/src/main/cpp/profiler.cpp
r1viollet: We now have an error path. Please consider whether this is what we want.

r1viollet:
I'm trying to decide if we are likely to hit the 200ms lag at shutdown. This would explain bench results.

r1viollet:
Re-running benchmarks again [here](https://gitlab.ddbuild.io/DataDog/java-profiler/-/pipelines/122041327)
The nice thing would be to have a history of all runs, that would help analyze at what point we no longer get counter performance.

PR #615, opened 2026-06-24
Use ThreadLocal to setup protection longjmp buffer and fix jmp_buf chaining

zhengyu123:
**What does this PR do?**:
This PR intents to resolve some of mysterious crashes seen in later releases, e.g.  [PROF-15206](https://datadoghq.atlassian.net/browse/PROF-15206), although, it might not actually solve this particular problem, but I believe it should have positive impact - Cloude and I believe that overwriting JavaThread's exception field is dangerous, in fact, overwriting *any* JVM data structures are dangerous, because the writes can result in races and readers may see wrong values, that can lead to crashes.

Before introducing Thread Local Handshake in JVM, thread object is usually only read by 'current` thread, so it is relative safe to alter, then restore values inside signal handler,  but not any more - a thread may be able to read other thread's states before it decides how to proceed the handshake, so it may read wrong values.

This PR:
***Avoid altering `JavaThread` for jmp_buf***
Uses a thread local for storing longjmp buffer and eliminates the dependency on overwriting `JavaThread`'s `exception` field, so
- Avoiding other thread reads wrong value of the exception field.
- `longjmp` now protects all threads, not just `JavaThread` - should improve stability. 

***Chaining jmp_buf***
A none signal based sampler can be interrupted by signal based sampler. Therefore, there may be multiple `walkVM()` calls on stack, each of them sets up a  protection `jmp_buf` for stack walk, the `jmp_buf`s need to be chained to ensure to jump back to expected location. 

**Motivation**:
Improve profiler's stability.

**Additional Notes**:
In additional, `exception` field only exists in `JavaThread`, but detection code in Profiler is somewhat questionable, especially, for the subclasses of `JavaThread`

**How to test the change?**:
- CI tests
- New unit tests

**For Datadog employees**:
- [x] JIRA: [PROF-15206](https://datadoghq.atlassian.net/browse/PROF-15206)

[PROF-15206]: https://datadoghq.atlassian.net/browse/PROF-15206?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -145,9 +146,12 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
     }
 }
 
+static ThreadLocal<jmp_buf*> jmp_ctx;
```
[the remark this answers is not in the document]
zhengyu123: Although `pthread_getspecific()` is marked as not async-signal-safe, but we worked around it by avoiding allocation.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -167,7 +171,14 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
...
     }
-    void* saved_exception = vm_thread != NULL ? vm_thread->exception() : NULL;
+    // Only write _exception_file for real Java application threads.
+    // JVM-internal threads like MonitorDeflationThread are JavaThread subclasses
+    // in JDK 25+ (different vtable → isJavaThread() returns false) but they are
+    // not regular application threads.  Writing _exception_file for them disturbs
+    // JDK 25's ObjectMonitorDeflationSafepointer, which reads thread state at
+    // safepoint boundaries and can crash in deflate_monitor_list when the field
+    // holds a stale jmp_buf address instead of a C-string or NULL.
+    bool setup_crash_protection = vm_thread != NULL && VMThread::isJavaThread(vm_thread);
```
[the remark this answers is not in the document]
zhengyu123: no longer testing for Java thread, but longjmp protecting all threads.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -183,18 +199,20 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
...
-        vm_thread->exception() = &crash_protection_ctx;
-        if (profiled_thread != nullptr) {
+        if (setup_crash_protection) {
+            jmp_ctx.set(&crash_protection_ctx);
+        }
+        if (profiled_thread != nullptr && setup_crash_protection) {
             profiled_thread->setCrashProtectionActive(true);
         }
-        if (setjmp(crash_protection_ctx) != 0) {
+        if (setup_crash_protection && setjmp(crash_protection_ctx) != 0) {
```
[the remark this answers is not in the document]
zhengyu123: We don't want to use `vm_thread->exception` to store longjmp buffer - it can be a problem for pre-existing threads.

ddprof-lib/src/main/cpp/jvmThread.h
```diff
@@ -6,48 +6,48 @@
...
+    }
+
+    // Per-thread initialization
+    static void initThread();
+
     static inline void* current() {
-        assert(isInitialized());
-        return pthread_getspecific(_thread_key);
+        return _jvm_thread.get();
     }
```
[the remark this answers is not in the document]
zhengyu123: Failed initialization is checked via `JVMThread::isInitialized()`

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -167,7 +167,15 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
     } else {
         Counters::increment(WALKVM_VMTHREAD_OK);
     }
-    void* saved_exception = vm_thread != NULL ? vm_thread->exception() : NULL;
+    // Only write _exception_file for real Java application threads.
+    // JVM-internal threads like MonitorDeflationThread are JavaThread subclasses
+    // in JDK 25+ (different vtable → isJavaThread() returns false) but they are
+    // not regular application threads.  Writing _exception_file for them disturbs
+    // JDK 25's ObjectMonitorDeflationSafepointer, which reads thread state at
```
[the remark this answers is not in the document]
zhengyu123: Fixed

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -183,18 +194,20 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
         if (anchor == NULL) {
             Counters::increment(WALKVM_ANCHOR_NULL);
         }
```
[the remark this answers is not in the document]
zhengyu123: Fixed

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -40,52 +49,60 @@
...
     ThreadLocal(const ThreadLocal&) = delete;
     ThreadLocal& operator=(const ThreadLocal&) = delete;
 
-    ThreadLocal() {
+    ThreadLocal() : _key(INVLID_KEY) {
         static_assert(sizeof(T) == sizeof(void*),
                       "ThreadLocal<T> requires sizeof(T)==sizeof(void*); use a pointer type or add a specialization");
-        _key_valid = pthread_key_create(&_key, F) == 0;
+        pthread_key_create(&_key, F);
         // What to do if we can not create a key?
```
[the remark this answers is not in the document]
zhengyu123: Fixed

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -179,22 +186,23 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
 
     VMJavaFrameAnchor* anchor = NULL;
     if (vm_thread != NULL) {
+        _jmp_ctx.set(&crash_protection_ctx);
+        if (profiled_thread != nullptr) {
+            profiled_thread->setCrashProtectionActive(true);
+        }
```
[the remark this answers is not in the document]
zhengyu123: Refactored

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -179,22 +186,23 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
 
     VMJavaFrameAnchor* anchor = NULL;
     if (vm_thread != NULL) {
+        _jmp_ctx.set(&crash_protection_ctx);
+        if (profiled_thread != nullptr) {
```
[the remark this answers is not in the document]
zhengyu123: Refactored

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -885,18 +892,17 @@ void HotspotSupport::checkFault(ProfiledThread* thrd) {
     // early init or in crash recovery tests). sameStack uses a fixed 8KB threshold which
     // can fail with ASAN-inflated frames, but the crashProtectionActive path handles that.
     bool protected_walk = (thrd != nullptr && thrd->isCrashProtectionActive())
-                       || sameStack(vm_thread->exception(), &vm_thread);
+                       || sameStack(_jmp_ctx.get(), &vm_thread);
     if (!protected_walk) {
         return;
     }
```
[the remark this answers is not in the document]
zhengyu123: Refactored

ddprof-lib/src/test/cpp/hotspot_crash_protection_ut.cpp
```diff
@@ -0,0 +1,255 @@
...
+ * Unit tests for the MonitorDeflationThread crash-protection gate.
+ *
+ * Root cause (JDK 25.0.2): walkVM() unconditionally wrote a jmp_buf address
+ * into ThreadShadow::_exception_file for every non-null VMThread, including
+ * JVM-internal threads like MonitorDeflationThread.  In JDK 25,
+ * ObjectMonitorDeflationSafepointer reads _exception_file at safepoint
+ * boundaries; a stale jmp_buf address caused a crash in deflate_monitor_list.
+ *
+ * Fix: gate the write on `VMThread::isJavaThread(vm_thread)`.
+ *
```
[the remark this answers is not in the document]
zhengyu123: Refactored test

ddprof-lib/src/test/cpp/hotspot_crash_protection_ut.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+// Gate is off for a JVM-internal thread (TYPE_NOT_JAVA_THREAD):
+// _exception_file must NOT be written.
+TEST_F(CrashProtectionGateTest, GateOffForNonJavaThread) {
+    _pt->setJavaThread(false);
+    // Simulate: vm_thread != NULL (non-null pointer means JVM thread struct exists)
+    bool vm_thread_non_null = true;
+    bool setup_crash_protection = vm_thread_non_null && fastPathIsJavaThread(_pt);
+    EXPECT_FALSE(setup_crash_protection);
+}
```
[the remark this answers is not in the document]
zhengyu123: Refactored

ddprof-test/src/test/java/com/datadoghq/profiler/cpu/MonitorDeflationThreadSafetyTest.java
```diff
@@ -0,0 +1,87 @@
...
+import com.datadoghq.profiler.AbstractProfilerTest;
+import org.junitpioneer.jupiter.RetryingTest;
+
+/**
+ * Regression test for the MonitorDeflationThread crash in JDK 25.0.2+.
+ *
+ * Root cause: walkVM() unconditionally wrote a jmp_buf address into
+ * ThreadShadow::_exception_file for every non-null VMThread, including
+ * JVM-internal threads such as MonitorDeflationThread.  In JDK 25,
+ * ObjectMonitorDeflationSafepointer reads thread state (including
```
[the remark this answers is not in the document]
zhengyu123: Fixed

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -167,39 +175,31 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
...
     int fp_chain_depth = 0;
 
-    ProfiledThread* profiled_thread = ProfiledThread::currentSignalSafe();
-
+    _jmp_ctx.set(&crash_protection_ctx);
+    if (setjmp(crash_protection_ctx) != 0) {
+        // checkFault() does a longjmp from inside segvHandler, bypassing
+        // segvHandler's SignalHandlerScope destructor.  Compensate.
+        SIGNAL_HANDLER_UNWIND_AFTER_LONGJMP();
+        _jmp_ctx.clear();
```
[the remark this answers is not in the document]
zhengyu123: Had a workaround for it.

ddprof-lib/src/main/cpp/jvmThread.cpp
```diff
@@ -4,31 +4,46 @@
...
+ */
+void JVMThread::initThread() {
+    if (VM::isHotspot()) {
+      HotspotSupport::initThread();
     }
-  }
-  // _tid is initialized in currentThreadSlow()
-  assert(_tid != nullptr);
-  return _thread_key != pthread_key_t(-1);
 }
```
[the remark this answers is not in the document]
zhengyu123: Yes, we have problem with pre-existing threads, will deal with the problem in separate PR.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -875,28 +871,21 @@ void HotspotSupport::checkFault(ProfiledThread* thrd) {
...
-        return;
-    }
-
     // Prefer the semantic crash protection flag (reliable regardless of stack frame sizes).
     // Fall back to sameStack heuristic when ProfiledThread TLS is unavailable (e.g. during
     // early init or in crash recovery tests). sameStack uses a fixed 8KB threshold which
     // can fail with ASAN-inflated frames, but the crashProtectionActive path handles that.
-    bool protected_walk = (thrd != nullptr && thrd->isCrashProtectionActive())
-                       || sameStack(vm_thread->exception(), &vm_thread);
+    bool protected_walk = isThreadProtectedByLongjmp();
```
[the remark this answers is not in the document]
zhengyu123: updated.

ddprof-lib/src/main/cpp/jvmThread.cpp
```diff
@@ -4,31 +4,46 @@
...
 bool JVMThread::initialize() {
   void* current_thread = currentThreadSlow();
-  if (current_thread == nullptr) {
-    return false;
-  }
+  _jvm_thread.initialize(current_thread);
+  // _tid is side-effect of currentThreadSlow()
+  assert(_tid != nullptr);
+  return _jvm_thread.isKeyValid();
+}
```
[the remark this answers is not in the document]
zhengyu123: Add assert and comment

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -167,39 +175,31 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
...
+        // checkFault() does a longjmp from inside segvHandler, bypassing
+        // segvHandler's SignalHandlerScope destructor.  Compensate.
+        SIGNAL_HANDLER_UNWIND_AFTER_LONGJMP();
+        _jmp_ctx.clear();
+        if (depth < max_depth) {
+            fillFrame(frames[depth++], BCI_ERROR, "break_not_walkable");
+        }
+        return depth;
+    }
+
```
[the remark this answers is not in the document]
zhengyu123: Fixed

ddprof-lib/src/main/cpp/jvmThread.cpp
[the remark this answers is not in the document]
zhengyu123: Fixed for attached thread

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -167,39 +175,31 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
...
     int actual_max_depth = truncated ? max_depth + 1 : max_depth;
     bool fp_chain_fallback = false;
     int fp_chain_depth = 0;
 
-    ProfiledThread* profiled_thread = ProfiledThread::currentSignalSafe();
+    if (setjmp(crash_protection_ctx) != 0) {
+        // checkFault() does a longjmp from inside segvHandler, bypassing
+        // segvHandler's SignalHandlerScope destructor.  Compensate.
+        SIGNAL_HANDLER_UNWIND_AFTER_LONGJMP();
+        _jmp_ctx.clear();
```
jbachorik: 🗿 🤖 🔴

**[Sphinx Review — LOW]** On the longjmp-landing path _jmp_ctx.clear() calls pthread_setspecific (not async-signal-safe) while still effectively in signal-handler context; benign only because set() primed the slot in the same walkVM, an undocumented invariant.

**Suggestion:** Document that clear() here only ever runs after set() primed the slot in the same walkVM invocation, or reset the slot via a path that does not call pthread_setspecific.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -875,28 +871,17 @@ void HotspotSupport::checkFault(ProfiledThread* thrd) {
...
-
-    // Prefer the semantic crash protection flag (reliable regardless of stack frame sizes).
-    // Fall back to sameStack heuristic when ProfiledThread TLS is unavailable (e.g. during
-    // early init or in crash recovery tests). sameStack uses a fixed 8KB threshold which
-    // can fail with ASAN-inflated frames, but the crashProtectionActive path handles that.
-    bool protected_walk = (thrd != nullptr && thrd->isCrashProtectionActive())
-                       || sameStack(vm_thread->exception(), &vm_thread);
-    if (!protected_walk) {
+    // Check if longjmp is setup for this thread
+   if (!isThreadProtectedByLongjmp()) {
```
jbachorik: 🗿 🤖 🔴

**[Sphinx Review — LOW]** Indentation defect: `if (!isThreadProtectedByLongjmp()) {` is indented with 3 spaces, misaligned with the surrounding 4-space block; likely to trip spotlessCheck/clang-format.

**Suggestion:** Reindent to 4 spaces.

ddprof-test/src/test/java/com/datadoghq/profiler/cpu/MonitorDeflationThreadSafetyTest.java
```diff
@@ -0,0 +1,85 @@
...
+     * deflation on the next MonitorDeflationThread pass.
+     */
+    private static void inflateAndDeflateMonitors(long durationMs, long waveSleepMs)
+            throws InterruptedException {
+        Object[] monitors = new Object[MONITOR_COUNT];
+        for (int i = 0; i < monitors.length; i++) {
+            monitors[i] = new Object();
+        }
+
+        long deadline = System.currentTimeMillis() + durationMs;
```
jbachorik: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** The regression test never exercises the primary vmStructs.h TOCTOU change: the new `safeFetch64` branch in `VMKlass::fromOop()` is gated on `_compact_object_headers` (only set under `-XX:+UseCompactObjectHeaders`, non-default), and `getProfilerCommand()` returns only `cpu=1ms` with no such flag. The churn loop also spends nearly all its time in `mon.wait(1)`/`Thread.sleep`, so `cpu=1ms` makes the `verifyEvents` assertion flaky.

**Suggestion:** Add a JVM-args variant enabling `-XX:+UseCompactObjectHeaders` (assumeTrue-gated for older JDKs) so the deflation TOCTOU branch is covered, and either drive real CPU work during churn or relax the `verifyEvents` assertion so the test fails only on the targeted crash.

jbachorik:
🗿 🤖 🔴

**[Sphinx Review — HIGH]** `walkVM()` calls `_jmp_ctx.set(&crash_protection_ctx)` (i.e. `pthread_setspecific`) from signal-handler context for any thread whose slot has not been primed. `HotspotSupport::initThread()` is only called in:
- `Profiler::onThreadStart` (JVMTI callback — fires on the starting thread ✓)
- `libraryPatcher_linux.cpp` pthread-create wrappers (fires on the new thread ✓)
- `VM::initProfilerBridge(attach=true)` (fires on the attaching thread ✓)

None of these cover threads that exist **before** profiling starts: GC threads, compiler threads, `MonitorDeflationThread`, VM threads, and any app threads enrolled by `CTimer::start()` via `listThreads()`. Those threads enter `walkVM()` without a primed slot, so `_jmp_ctx.set()` inside the signal handler is their first `pthread_setspecific` call — not async-signal-safe (may malloc on glibc for key indices ≥ `PTHREAD_KEY_2NDLEVEL_SIZE`).

Note: calling `JVMThread::initThread()` inside `CTimer::registerThread(tid)` does **not** fix this — `pthread_setspecific` only sets the calling thread's slot, and `registerThread` runs on the profiler/registrar thread, not on `tid`.

**Valid fix options:**
1. Guarantee `_jmp_ctx`'s pthread key index stays < `PTHREAD_KEY_2NDLEVEL_SIZE` (32 on glibc) so no lazy allocation ever occurs in `set()`.
2. Use a JVMTI `ThreadStart` callback (which fires on the starting thread) already wired via `onThreadStart` — but this only covers threads started *after* the profiler. For pre-existing threads, a supplemental mechanism is needed (e.g. a one-time `SIGPROF`-safe null-set from a thread-attach shim, or a `pthread_key_create` destructor-only design where the first `set` is idempotent and allocation-free).

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
zhengyu123: Piggyback `jmp_buf` to per-thread `ProfiledThread`

ddprof-test/src/test/java/com/datadoghq/profiler/cpu/MonitorDeflationThreadSafetyTest.java
zhengyu123: Defer for later - we will need to setup tests for various JVM configurations.

zhengyu123:
> 🗿 🤖 🔴
> 
> **[Sphinx Review — HIGH]** `walkVM()` calls `_jmp_ctx.set(&crash_protection_ctx)` (i.e. `pthread_setspecific`) from signal-handler context for any thread whose slot has not been primed. `HotspotSupport::initThread()` is only called in:
> 
> * `Profiler::onThreadStart` (JVMTI callback — fires on the starting thread ✓)
> * `libraryPatcher_linux.cpp` pthread-create wrappers (fires on the new thread ✓)
> * `VM::initProfilerBridge(attach=true)` (fires on the attaching thread ✓)
> 
> None of these cover threads that exist **before** profiling starts: GC threads, compiler threads, `MonitorDeflationThread`, VM threads, and any app threads enrolled by `CTimer::start()` via `listThreads()`. Those threads enter `walkVM()` without a primed slot, so `_jmp_ctx.set()` inside the signal handler is their first `pthread_setspecific` call — not async-signal-safe (may malloc on glibc for key indices ≥ `PTHREAD_KEY_2NDLEVEL_SIZE`).
> 
> Note: calling `JVMThread::initThread()` inside `CTimer::registerThread(tid)` does **not** fix this — `pthread_setspecific` only sets the calling thread's slot, and `registerThread` runs on the profiler/registrar thread, not on `tid`.
> 
> **Valid fix options:**
> 
> 1. Guarantee `_jmp_ctx`'s pthread key index stays < `PTHREAD_KEY_2NDLEVEL_SIZE` (32 on glibc) so no lazy allocation ever occurs in `set()`.
> 2. Use a JVMTI `ThreadStart` callback (which fires on the starting thread) already wired via `onThreadStart` — but this only covers threads started _after_ the profiler. For pre-existing threads, a supplemental mechanism is needed (e.g. a one-time `SIGPROF`-safe null-set from a thread-attach shim, or a `pthread_key_create` destructor-only design where the first `set` is idempotent and allocation-free).

@jbachorik Piggyback `jmp_buf` to per-thread's `ProfiledThread`, skip the sample if `ProfiledThread` is not available, for now.

Will file separate PR to address `ProfiledThread`

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
zhengyu123: Refactored

rkennke (commented):
Reviewed the crash-protection rework. The core mechanism is correct and a real improvement — moving the `jmp_buf` out of `JavaThread::exception` into `ProfiledThread` TLS removes a genuine race, and the nested-`walkVM` chaining discipline is sound (verified no exit path bypasses the `done:`/recovery restore; `depth` is `volatile` and `prev_jmp_buf`/`prof_thread` are fixed before `setjmp`). No blocking bugs. Left inline notes: one behavior change to confirm, one latent ODR fragility, plus a few stale comments / dead code / style nits.

One finding that isn't on a changed line: in `profiler.cpp`, the comments at line ~936 ("checked via sameStack heuristic in HotspotSupport::checkFault") and line ~974 ("it uses sameStack() heuristic to check if we're in a protected stack walk") are now stale — `checkFault` no longer has a `sameStack` fallback and returns early when `thrd == nullptr`. `stackWalker.h::sameStack()` also appears to be dead now. See the inline note near line 947.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -227,15 +227,31 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
 __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontext, ASGCT_CallFrame* frames, int max_depth,
                         StackWalkFeatures features, EventType event_type,
                         const void* pc, uintptr_t sp, uintptr_t fp, int lock_index, bool* truncated) {
+
     // VMStructs is only available for hotspot JVM 
     assert(VM::isHotspot());
+
+    ProfiledThread* prof_thread = ProfiledThread::currentSignalSafe();
+    if (prof_thread == nullptr) {
+        Counters::increment(SAMPLES_DROPPED_THREAD_LOCAL);
```
rkennke, line 237: Behavior change worth a conscious sign-off: `walkVM` now drops the sample entirely (returns 0) when there's no `ProfiledThread` TLS, whereas before it proceeded using `vm_thread`. This is self-consistent with the new model (protection can't be installed without TLS), but it's a small coverage loss for transient/not-yet-registered threads. Counted via `SAMPLES_DROPPED_THREAD_LOCAL`, so it's observable — just flagging it's intentional.

ddprof-lib/src/main/cpp/hotspot/vmStructs.inline.h
```diff
@@ -74,6 +76,14 @@ VMMethod* VMThread::compiledMethod() {
     return NULL;
 }
 
+inline bool crashProtectionActive() {
```
rkennke: Latent ODR fragility: `crashProtectionActive()` is declared `inline` in `vmStructs.h` (line 44) and odr-used by the asserts inside `cast_to`/`at`/`load_then_cast` there, but its definition now lives only here. Several TUs include `vmStructs.h` without this inline header (`flightRecorder.cpp`, `vmEntry.cpp`, `libraries.cpp`, `jitCodeCache.cpp`, `livenessTracker.cpp`). It links today (CI green), but a future debug-build TU that instantiates one of those asserted methods without also including `vmStructs.inline.h` would fail with an undefined reference. Previously the definition was reachable from `vmStructs.h` directly. Consider keeping a definition reachable from the header the asserts live in.

ddprof-lib/src/main/cpp/threadLocalData.h
```diff
@@ -1,5 +1,25 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#ifndef _THREADLOCALDATA_H
+#define _THREADLOCALDATA_H
+
+#include "context.h"
+#include "otel_context.h"
+#include "os.h"
+#include "threadLocalData.h"
```
rkennke: Self-include — `threadLocalData.h` includes itself. Harmless (include-guarded) but a leftover from merging the old small `ThreadLocalData` header; should be removed.

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -946,8 +945,8 @@ int Profiler::crashHandlerInternal(int signo, siginfo_t *siginfo, void *ucontext
     have_tls_protection = true;
   }
   // If thrd == nullptr, we proceed but with limited handling capability.
```
rkennke: Good that this comment was updated. Two sibling comments in this same function still reference the removed fallback: line ~936 ("checked via sameStack heuristic in HotspotSupport::checkFault") and line ~974 ("it uses sameStack() heuristic ..."). `checkFault` now just returns early when `thrd == nullptr`, so those are misleading. `stackWalker.h::sameStack()` looks dead now too.

ddprof-lib/src/main/cpp/jvmThread.h
```diff
@@ -6,48 +6,46 @@
...
-
     /*
      * The initialization happens in early startup, in single-threaded mode,
      * no synchronization is needed
      */
     static bool initialize();
+    static bool isInitialized();
+
+    // Per-thread initialization
+    static void initThread();
```
rkennke: `initThread()` is declared but never called (the empty definition was dropped in the current head). Dead declaration — either wire it in or remove it.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -227,15 +227,31 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
...
     HotspotStackFrame frame(ucontext);
     uintptr_t bottom = (uintptr_t)&frame + MAX_WALK_SIZE;
 
     Profiler* profiler = Profiler::instance();
     int bcp_offset = InterpreterFrame::bcp_offset();
 
+
     jmp_buf crash_protection_ctx;
+    // Chaining jmp_buf
+    // None signal based sampler can be interrupted by signal based sampler,
```
rkennke: Nit: "None signal based sampler" reads oddly — should be "A non-signal-based sampler".

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -953,33 +957,25 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
...
 
-    // Prefer the semantic crash protection flag (reliable regardless of stack frame sizes).
-    // Fall back to sameStack heuristic when ProfiledThread TLS is unavailable (e.g. during
-    // early init or in crash recovery tests). sameStack uses a fixed 8KB threshold which
-    // can fail with ASAN-inflated frames, but the crashProtectionActive path handles that.
-    bool protected_walk = (thrd != nullptr && thrd->isCrashProtectionActive())
-                       || sameStack(vm_thread->exception(), &vm_thread);
-    if (!protected_walk) {
+    // Check if longjmp is setup for this thread
+   if (!thrd->isProtected()) {
```
rkennke: Nit: this line is indented 3 spaces, misaligned with the 4-space function body.

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -112,41 +131,46 @@ template <>
...
     ThreadLocal& operator=(const ThreadLocal&) = delete;
 
-    ThreadLocal() {
+    ThreadLocal() : _key(INVALID_KEY) {
         // Only support 64-bit platforms, double and void* are the same size
         static_assert(sizeof(void*) == 8);
         static_assert(sizeof(double) == 8);
-        _key_valid = pthread_key_create(&_key, nullptr) == 0;
+        if (pthread_key_create(&_key, nullptr) != 0) {
+            _key =INVALID_KEY;
```
rkennke: Nit: missing space — `_key =INVALID_KEY;`.

ddprof-lib/src/main/cpp/jvmThread.cpp
```diff
@@ -4,29 +4,27 @@
...
-  }
-
-  for (int i = 0; i < 1024; i++) {
-    if (pthread_getspecific((pthread_key_t)i) == current_thread) {
-        _thread_key = pthread_key_t(i);
-        break;
-    }
-  }
-  // _tid is initialized in currentThreadSlow()
+  // Called by known JavaThread, cannot be nullpltr
```
rkennke: Typo: "nullpltr" → "nullptr".

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
zhengyu123: Yes. We will bring back priming/or converting to native agent in a future PR.

ddprof-lib/src/main/cpp/hotspot/vmStructs.h
```diff
@@ -1221,18 +1215,4 @@ class InterpreterFrame : VMStructs {
...
-    if (pt != nullptr && pt->isCrashProtectionActive()) return true;
-    // Fallback for threads without ProfiledThread TLS (e.g. JVM internal threads):
-    // if walkVM has set up setjmp protection via vm_thread->exception(), the assert
-    // is equally redundant — any bad read will be caught by the SIGSEGV handler.
-    // Uses VMThread::isExceptionActive() which reads the field directly without
-    // going through at() to avoid recursive assertion.
-    return JVMThread::key() != pthread_key_t(-1) && VMThread::isExceptionActive();
-}
-
 #endif // _HOTSPOT_VMSTRUCTS_H
```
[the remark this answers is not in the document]
zhengyu123: Fixed

ddprof-lib/src/main/cpp/profiler.cpp
zhengyu123: Fixed comments

ddprof-lib/src/main/cpp/jvmThread.h
zhengyu123: Removed

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
zhengyu123: Fixed

ddprof-lib/src/main/cpp/threadLocal.h
zhengyu123: Fixed

ddprof-lib/src/main/cpp/jvmThread.cpp
zhengyu123: Fixed

ddprof-lib/src/main/cpp/jvmThread.h
```diff
@@ -6,48 +6,43 @@
...
      * no synchronization is needed
      */
     static bool initialize();
+    static bool isInitialized();
+
     static inline void* current() {
-        assert(isInitialized());
-        return pthread_getspecific(_thread_key);
+        return _jvm_thread.get();
     }
```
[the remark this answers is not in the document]
zhengyu123: Added assert to capture the scenario that `JVMThread::initialize()` has not yet called - debug build should be able to capture the mistake. 

Other than that, the key *should* not be invalid, JVM depends on it.

ddprof-lib/src/test/cpp/hotspot_crash_protection_ut.cpp
```diff
@@ -0,0 +1,395 @@
...
+// At least 2 of 3 must match — tolerating product-vs-debug vtable variation.
+// ---------------------------------------------------------------------------
+
+namespace {
+
+// Replicates the 2-of-3 vote in hasJavaThreadVtable().
+static bool vtableVote(void** thread_vtbl, void** known_vtbl) {
+    int matches = (thread_vtbl[1] == known_vtbl[1])
+                + (thread_vtbl[3] == known_vtbl[3])
+                + (thread_vtbl[5] == known_vtbl[5]);
```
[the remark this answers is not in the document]
zhengyu123: This test is meaningless - removed.

ddprof-lib/src/main/cpp/jvmThread.cpp
```diff
@@ -4,29 +4,27 @@
...
-  }
-  // _tid is initialized in currentThreadSlow()
+  // Called by known JavaThread, cannot be nullptr
+  assert(current_thread != nullptr && "Must not be nullptr");
+  _jvm_thread.initialize(current_thread);
+  // _tid is side-effect of currentThreadSlow()
   assert(_tid != nullptr);
-  return _thread_key != pthread_key_t(-1);
+  return _jvm_thread.isKeyValid();
+}
```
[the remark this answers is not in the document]
zhengyu123: `currentThreadSlow()` should never return `nullptr`, it is called from known Java thread.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -227,15 +227,31 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
...
+    // Chaining jmp_buf
+    // A non-signal-based-sampler can be interrupted by signal based sampler,
+    // then we end up multiple HotspotSupport::walkVM() calls on stack,
+    // each one sets up jmp_buf, they need to be chained to jump back to
+    // correct location.
+    jmp_buf* prev_jmp_buf = prof_thread->getJmpCtx();
+
     VMThread* vm_thread = VMThread::current();
     if (vm_thread != NULL && !vm_thread->isThreadAccessible()) {
         Counters::increment(WALKVM_THREAD_INACCESSIBLE);
```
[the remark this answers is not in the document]
zhengyu123: Pre-existing - move up the protection

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -946,8 +945,8 @@ int Profiler::crashHandlerInternal(int signo, siginfo_t *siginfo, void *ucontext
...
-  // Only HotspotSupport::checkFault (which has its own sameStack fallback)
-  // and the JDK-8313796 workaround can safely handle faults without TLS.
+  // Only HotspotSupport::checkFault and the JDK-8313796 workaround can safely 
+  // handle faults without TLS.
```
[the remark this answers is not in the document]
zhengyu123: Fixed

ddprof-lib/src/main/cpp/threadLocalData.h
```diff
@@ -14,4 +34,305 @@ class ThreadLocalData {
...
+    _jmp_buf = buf;  
+  }
+
+  inline jmp_buf* getJmpCtx() const {
+    return _jmp_buf;
+  }
+
+  inline bool isProtected() const {
+    return _jmp_buf != nullptr;
+  }
```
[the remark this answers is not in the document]
zhengyu123: Why, the reads/writes are on the same thread - inside signal handler recursion and reads/writes are not overlapped.

ddprof-lib/src/main/cpp/threadLocalData.h
```diff
@@ -14,4 +34,295 @@ class ThreadLocalData {
...
```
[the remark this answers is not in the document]
zhengyu123: No, I don't concurrent access to `jmp_buf`

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -934,9 +933,8 @@ int Profiler::crashHandlerInternal(int signo, siginfo_t *siginfo, void *ucontext
 
   // Reentrancy protection: use TLS-based tracking if available.
   // If TLS is not available, we can only safely handle faults that we can
-  // prove are from our protected code paths (checked via sameStack heuristic
-  // in HotspotSupport::checkFault). For anything else, we must chain immediately
-  // to avoid claiming faults that aren't ours.
+  // prove are from our protected code paths (checked via jmp_buf is set for the thread).
+  // For anything else, we must chain immediately to avoid claiming faults that aren't ours.
   bool have_tls_protection = false;
```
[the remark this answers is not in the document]
zhengyu123: Fixed

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -227,15 +227,46 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
...
                         StackWalkFeatures features, EventType event_type,
                         const void* pc, uintptr_t sp, uintptr_t fp, int lock_index, bool* truncated) {
+
     // VMStructs is only available for hotspot JVM 
     assert(VM::isHotspot());
+
+    ProfiledThread* prof_thread = ProfiledThread::currentSignalSafe();
+    if (prof_thread == nullptr) {
+        Counters::increment(SAMPLES_DROPPED_THREAD_LOCAL);
+        return 0;
```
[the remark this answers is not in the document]
zhengyu123: This is known, to be resolved in future PR.

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -159,10 +183,71 @@ class ThreadLocal<double> {
...
+        return _key != INVALID_KEY;
+    }
+
+    pthread_key_t key() const {
+        return _key;
+    }
+
+    void set(JVMThread* value) {
+        assert(false && "Should not reach here, value is set by JVM");
+    }
```
kaahos: could it be replaced by a deleted member function? it would replace the check by a compiled-time check, make the cpp more idiomatic and would be present in every build (not just debug).

```suggestion
void set(JVMThread* value) = delete;
```

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -159,10 +183,71 @@ class ThreadLocal<double> {
...
+
+    void* get() const {
+        assert(isKeyValid() && "Invalid pthread key");
+        return pthread_getspecific(_key);
+    }
+
+    void clear() {
+        assert(false && "Should not reach here");
+    }
+};
```
kaahos: same here.

```suggestion
void clear() = delete;
```

ddprof-lib/src/main/cpp/jvmThread.cpp
```diff
@@ -4,29 +4,27 @@
...
 
-pthread_key_t JVMThread::_thread_key = pthread_key_t(-1);
 jfieldID JVMThread::_tid = nullptr;
+ThreadLocal<JVMThread*> JVMThread::_jvm_thread;
 
 bool JVMThread::initialize() {
   void* current_thread = currentThreadSlow();
-  if (current_thread == nullptr) {
-    return false;
-  }
```
kaahos, line 19: shouldn't this guard be restored just in case? i know `currentThreadSlow()` shouldn't return null but the guard already pre-existed this PR so why not keeping it and being extra-careful? No strong opinion on that though.

ddprof-lib/src/main/cpp/threadLocalData.h
```diff
@@ -14,4 +34,295 @@ class ThreadLocalData {
...
+  void resetCrashHandler() {
+    __atomic_store_n(&_crash_depth, 0u, __ATOMIC_RELAXED);
+  }
+
+  bool isDeepCrashHandler() {
+    return __atomic_load_n(&_crash_depth, __ATOMIC_RELAXED) > CRASH_HANDLER_NESTING_LIMIT;
+  }
+
+  inline void setJmpCtx(jmp_buf* buf) {
+    _jmp_buf = buf;
```
kaahos, line 208: `_jmp_buf` is not atomic and is modified from a signal handler, I think this is a C++ UB situation: https://en.cppreference.com/w/cpp/utility/program/signal. An atomic could be added, as for `_crash_depth`?

kaahos (commented):
I left some comments, feel free to push back if you disagree with some!

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -227,15 +227,46 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
...
     uintptr_t bottom = (uintptr_t)&frame + MAX_WALK_SIZE;
 
     Profiler* profiler = Profiler::instance();
     int bcp_offset = InterpreterFrame::bcp_offset();
 
+
     jmp_buf crash_protection_ctx;
+    // Chaining jmp_buf
+    // A non-signal-based-sampler can be interrupted by signal based sampler,
+    // then we end up multiple HotspotSupport::walkVM() calls on stack,
```
jbachorik: Typo: 'end up with multiple ...'

ddprof-lib/src/main/cpp/jvmThread.h
```diff
@@ -6,48 +6,46 @@
...
      */
     static bool initialize();
+    static bool isInitialized();
+
     static inline void* current() {
-        assert(isInitialized());
-        return pthread_getspecific(_thread_key);
+        // Assertion to ensure initialize() is called. Otherwise,
+        // the key should be valid, JVM depends on it
+        assert(_jvm_thread.isKeyValid() && "Must be");
```
jbachorik: The assert message should be more descriptive

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -946,8 +944,8 @@ int Profiler::crashHandlerInternal(int signo, siginfo_t *siginfo, void *ucontext
     have_tls_protection = true;
   }
   // If thrd == nullptr, we proceed but with limited handling capability.
-  // Only HotspotSupport::checkFault (which has its own sameStack fallback)
-  // and the JDK-8313796 workaround can safely handle faults without TLS.
+  // Only HotspotSupport::checkFault and the JDK-8313796 workaround can safely 
+  // handle faults without TLS.
```
jbachorik: This does not seem to be true. If 'thrd == nullptr' we would definitely crash at line 957, so we can not really proceed wih `thrd` being `nullptr`

ddprof-lib/src/main/cpp/threadLocal.h
```diff
@@ -159,10 +183,71 @@ class ThreadLocal<double> {
...
+    }
+
+    // The key is created by JVM, find out the key.
+    // This method should be called very early at Profiler startup
+    // time
+    void initialize(void* current_thread) {
+        // Called from known JavaThread, it must not be nullptr.
+        assert(current_thread != nullptr && "Should not reach here");
+
+        long max_keys = sysconf(_SC_THREAD_KEYS_MAX);
```
jbachorik, line 218: Will this be read for each ThreadLocal instance? What is the anticipated number of such instances?
In case we are expecting this happening ~frequently, would it make sense to cache the `max_keys` across instances?

ddprof-test/src/test/java/com/datadoghq/profiler/cpu/MonitorDeflationThreadSafetyTest.java
```diff
@@ -0,0 +1,90 @@
...
+ * dereferenced those fields unconditionally for every non-null VMThread,
+ * which caused crashes inside MonitorDeflationThread.
+ *
+ * Fix: VMThread::isJavaThread() gates all JavaThread-only field accesses
+ * (anchor(), inDeopt(), compiledMethod(), …).  It uses a two-step check:
+ *   1. Fast path — a JVMTI ThreadStart-cached flag on ProfiledThread.
+ *   2. Slow path — vtable majority vote (2-of-3 entries match a known
+ *      JavaThread) for threads that bypass the JVMTI callback (compiler
+ *      threads, MonitorDeflationThread, etc.).
+ *
```
jbachorik: I don't see this in the changeset. Is this still valid? Where can I see the code?

ddprof-lib/src/main/cpp/threadLocal.h
zhengyu123: No, we should have only one instance and it is a static instance.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
zhengyu123: Fixed

ddprof-lib/src/main/cpp/jvmThread.h
zhengyu123: Okay, added disabling profiler logic (which I intended to do in next PR) here, to silent AI.

Modified comment accordingly.

ddprof-test/src/test/java/com/datadoghq/profiler/cpu/MonitorDeflationThreadSafetyTest.java
zhengyu123: My bad, I removed the tests, they seem meaningless. Removed comment.

ddprof-lib/src/main/cpp/profiler.cpp
zhengyu123: Fixed the comment. Line #957 is guarded by `have_tls_protection` which is `false` when 'thrd == nullptr'

ddprof-lib/src/main/cpp/jvmThread.cpp
zhengyu123: Yea, added

ddprof-lib/src/main/cpp/vmEntry.cpp
```diff
@@ -324,8 +324,17 @@ bool VM::initShared(JavaVM* vm) {
...
   }
 
+  // Initialize VMStructs
   VMStructs::init(lib);
 
+  // Check JVMSupport initialization state.
+  // Must call after JVM library is loaded and JVM type, version number
+  // and etc. are resolved
+  if (!JVMSupport::initialize()) {
+    return false;
```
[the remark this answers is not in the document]
zhengyu123: In theory, `JVMSupport::initialize()` should be called inside `JVMTI::VMInit()` callback, but the timing is not always right, `Porfiler::start()` can arrive earlier - so cakk `JVMSupport::initialize()` inside `Profiler::start()` - not very pretty :-(

ddprof-lib/src/main/cpp/jvmThread.h
```diff
@@ -6,48 +6,49 @@
...
+    }
+
     static inline void* current() {
-        assert(isInitialized());
-        return pthread_getspecific(_thread_key);
+        // If the key is invalid, we should *never* reach here, because
+        // profiling should have been disabled. See VM::initShared()d
+        assert(_jvm_thread.isKeyValid() && "Should not reach here");
+        return _jvm_thread.get();
     }
```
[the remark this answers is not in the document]
zhengyu123: If the key is invalid, profiler should *never* start, therefore, should never reach here.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -227,15 +228,46 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
...
                         const void* pc, uintptr_t sp, uintptr_t fp, int lock_index, bool* truncated) {
+
     // VMStructs is only available for hotspot JVM 
     assert(VM::isHotspot());
+
+    ProfiledThread* prof_thread = ProfiledThread::currentSignalSafe();
+    if (prof_thread == nullptr) {
+        Counters::increment(SAMPLES_DROPPED_THREAD_LOCAL);
+        return 0;
+    }
```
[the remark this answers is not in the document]
zhengyu123: Ignore this, will fix in future PR.

rkennke (approved):
Looks good to me. Thank you!

zhengyu123:
> Looks good to me. Thank you!

@rkennke Thank you for reviewing.

ddprof-lib/src/main/cpp/jvmThread.cpp
```diff
@@ -4,29 +4,25 @@
...
   if (current_thread == nullptr) {
     return false;
   }
-
-  for (int i = 0; i < 1024; i++) {
-    if (pthread_getspecific((pthread_key_t)i) == current_thread) {
-        _thread_key = pthread_key_t(i);
-        break;
-    }
+  // _tid is side-effect of currentThreadSlow()
```
jbachorik: Could you please add more info how it is initialized? It apparently happens several calls below, so it is not really clear by just looking at `currentThreadSlow`

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -942,8 +940,8 @@ int Profiler::crashHandlerInternal(int signo, siginfo_t *siginfo, void *ucontext
     have_tls_protection = true;
   }
   // If thrd == nullptr, we proceed but with limited handling capability.
-  // Only HotspotSupport::checkFault (which has its own sameStack fallback)
-  // and the JDK-8313796 workaround can safely handle faults without TLS.
+  // Only HotspotSupport::checkFault and the JDK-8313796 workaround can safely
```
jbachorik: This is the same comment as in #626 - `HotspotSupport::checkFault` will just bail out, there is no special 'safe' handling with limited capability.

ddprof-lib/src/main/cpp/vmEntry.cpp
```diff
@@ -556,7 +557,7 @@ bool VM::initProfilerBridge(JavaVM *vm, bool attach) {
 void VM::ready(jvmtiEnv *jvmti, JNIEnv *jni) {
   Profiler::check_JDK_8313796_workaround();
   Profiler::setupSignalHandlers();
-  JVMThread::initialize();
+
```
jbachorik: Not sure if this is just a rendering artifact but it seems like this line has extra spaces (also, not sure why spotless hasn't formatted it out ...)

ddprof-test/src/test/java/com/datadoghq/profiler/cpu/MonitorDeflationThreadSafetyTest.java
```diff
@@ -0,0 +1,88 @@
...
+ * application threads: fields such as the frame anchor, vframe_top, and
+ * continuation entry are not valid on them.  Before the fix, walkVM
+ * dereferenced those fields unconditionally for every non-null VMThread,
+ * which caused crashes inside MonitorDeflationThread.
+ *
+ * This test forces ObjectMonitor inflation and deflation to race with CPU
+ * profiler signal delivery.  If the fix regresses, the JVM will crash with
+ * a SIGSEGV or SIGBUS inside MonitorDeflationThread before the test ends.
+ */
+public class MonitorDeflationThreadSafetyTest extends AbstractProfilerTest {
```
jbachorik, line 26: Same comment as in #626 - it might be beneficial to move this under the chaos tests so it will be run nightly over longer time period and at high intensity.

jbachorik (approved):
A few minor/clarifying comments. Otherwise good to go.

ddprof-lib/src/main/cpp/jvmThread.cpp
zhengyu123: Fixed

ddprof-lib/src/main/cpp/vmEntry.cpp
zhengyu123: Fixed

ddprof-test/src/test/java/com/datadoghq/profiler/cpu/MonitorDeflationThreadSafetyTest.java
zhengyu123: No fuzzing framework/convention exists for Java tests, defer to a separate PR.

jbachorik: `ddprof-stresstest/src/chaos` ?

jbachorik: But it's ok to defer.

ddprof-lib/src/main/cpp/profiler.cpp
zhengyu123: Fixed

PR #616, opened 2026-06-26
Fix artifact upload permissions in musl aarch64 CI job

r1viollet:
**What does this PR do?**:
Fixes permission denied errors when uploading build artifacts in the test-linux-musl-aarch64 CI job.

**Motivation**:
The Alpine Docker container creates files as root, causing GitHub Actions to fail when trying to read them during artifact zip creation:
```
Error: EACCES: permission denied, open '/home/runner/work/java-profiler/java-profiler/build/reports/problems/problems-report.html'
```

**Additional Notes**:
- Uses `sudo chmod -R a+rX` after Docker execution to ensure files are readable (umask-independent)
- Capital `X` adds execute only to directories, consolidating the previous two-command approach
- Covers all build directories: `build/`, `ddprof-test/build/`, `ddprof-lib/build/`
- Applied after both Test and Generate Unwinding Report steps
- Errors are visible in logs (`|| true` only handles missing paths gracefully)

**How to test the change?**:
Run the test-linux-musl-aarch64 job and verify artifacts upload successfully without permission errors.

**For Datadog employees**:

### DIFF
--- a/.github/workflows/test_workflow.yml
+++ b/.github/workflows/test_workflow.yml
@@ -577,8 +577,8 @@ jobs:
          run: |
            # Files created by Docker container may have restrictive permissions
            # that prevent GitHub Actions from reading them during artifact creation
-           sudo chmod -R +r build/ ddprof-test/build/ ddprof-lib/build/ 2>/dev/null || true
-           sudo find build/ ddprof-test/build/ ddprof-lib/build/ -type d -exec chmod +x {} \; 2>/dev/null || true
+           sudo chmod -R +r build/ ddprof-test/build/ ddprof-lib/build/ || true
+           sudo find build/ ddprof-test/build/ ddprof-lib/build/ -type d -exec chmod +x {} \; || true
        - name: Generate Unwinding Report
          if: success() && matrix.config == 'debug'
          run: |
@@ -588,8 +588,8 @@ jobs:
              \"${{ matrix.config }}\" \"${{ env.JAVA_HOME }}\" \"${{ env.JAVA_TEST_HOME }}\"
            "
            # Fix permissions after Docker container execution (both top-level and subproject builds)
-           sudo chmod -R +r build/ ddprof-test/build/ 2>/dev/null || true
-           sudo find build/ ddprof-test/build/ -type d -exec chmod +x {} \; 2>/dev/null || true
+           sudo chmod -R +r build/ ddprof-test/build/ ddprof-lib/build/ || true
+           sudo find build/ ddprof-test/build/ ddprof-lib/build/ -type d -exec chmod +x {} \; || true
        - name: Add Unwinding Report to Job Summary
          if: success() && matrix.config == 'debug' && hashFiles('ddprof-test/build/reports/unwinding-summary.md') != ''
          run: |
--- a/.github/workflows/test_workflow.yml
+++ b/.github/workflows/test_workflow.yml
@@ -577,8 +577,8 @@ jobs:
          run: |
            # Files created by Docker container may have restrictive permissions
            # that prevent GitHub Actions from reading them during artifact creation
-           sudo chmod -R +r build/ 2>/dev/null || true
-           sudo find build/ -type d -exec chmod +x {} \; 2>/dev/null || true
+           sudo chmod -R +r build/ ddprof-test/build/ ddprof-lib/build/ 2>/dev/null || true
+           sudo find build/ ddprof-test/build/ ddprof-lib/build/ -type d -exec chmod +x {} \; 2>/dev/null || true
        - name: Generate Unwinding Report
          if: success() && matrix.config == 'debug'
          run: |
@@ -587,9 +587,9 @@ jobs:
              \"${{ github.sha }}\" \"musl/${{ matrix.java_version }}-${{ matrix.config }}-aarch64\" \
              \"${{ matrix.config }}\" \"${{ env.JAVA_HOME }}\" \"${{ env.JAVA_TEST_HOME }}\"
            "
-           # Fix permissions after Docker container execution
-           sudo chmod -R +r ddprof-test/build/reports/ 2>/dev/null || true
-           sudo find ddprof-test/build/reports/ -type d -exec chmod +x {} \; 2>/dev/null || true
+           # Fix permissions after Docker container execution (both top-level and subproject builds)
+           sudo chmod -R +r build/ ddprof-test/build/ 2>/dev/null || true
+           sudo find build/ ddprof-test/build/ -type d -exec chmod +x {} \; 2>/dev/null || true
        - name: Add Unwinding Report to Job Summary
          if: success() && matrix.config == 'debug' && hashFiles('ddprof-test/build/reports/unwinding-summary.md') != ''
          run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/test_workflow.yml
```diff
@@ -572,6 +572,13 @@ jobs:
              echo "musl-${{ matrix.java_version }}-${{ matrix.config }}-aarch64" >> failures_musl-${{ matrix.java_version }}-${{ matrix.config }}-aarch64.txt
              exit 1
            fi
+       - name: Fix permissions for artifact upload
```
jbachorik, line 575: **[Sphinx Review — LOW]** This fixes the symptom (root-owned files left by the container) rather than the cause. The Docker invocations run as root by default (no `--user` flag) and bind-mount the workspace, so every file the container writes is owned by root and unreadable/undeletable by the runner user. The chmod-after-the-fact approach must be replicated at every container step (already two copies, with diverging scope) and re-applied for any future container step.

**Suggestion:** Consider running the containers with `--user $(id -u):$(id -g)` so emitted files are owned by the runner user, removing the need for repeated post-hoc sudo chmod. If the in-container toolchain requires root, keep the chmod but extract it to a single reused step.

.github/workflows/test_workflow.yml
```diff
@@ -572,6 +572,13 @@ jobs:
...
+         if: always()
+         run: |
+           # Files created by Docker container may have restrictive permissions
+           # that prevent GitHub Actions from reading them during artifact creation
+           sudo chmod -R +r build/ ddprof-test/build/ ddprof-lib/build/ 2>/dev/null || true
```
jbachorik: **[Sphinx Review — MEDIUM]** The permission-fix commands swallow every failure via `2>/dev/null || true`, and the step runs `if: always()`. If `sudo` is unavailable, the path is missing, or the chmod is denied, the failure is silently discarded and the step reports success. The downstream consumer `prepare_reports.sh` also copies from the same root-owned trees with `|| true` on every `cp`/`find`. If the chmod silently fails, `prepare_reports.sh` silently skips the unreadable sources and the resulting `test-reports` artifact is quietly incomplete — and when the subsequent `Upload build artifacts` step then fails for the exact permission reason this code was meant to fix, the root cause is invisible.

**Suggestion:** Drop `2>/dev/null` so chmod failures appear in the step log, enabling diagnosis when artifacts are still incomplete or the upload still fails. Keep `|| true` only if a clean exit is genuinely desired, or echo a warning when chmod returns non-zero.

jbachorik: **[Sphinx Review — LOW]** `sudo chmod -R +r build/ ddprof-test/build/ ddprof-lib/build/` makes every file under the build trees world-readable. This job runs on a self-hosted shared runner group (`group: ARM LINUX SHARED`), not an ephemeral single-tenant GitHub-hosted runner. On a shared runner the filesystem can be visible to other jobs/tenants, so blanket world-readable build output (which may contain build-injected tokens or signing material) is a non-trivial exposure surface.

**Suggestion:** Restrict the recursive chmod to the exact report/artifact subtrees that are uploaded (e.g. `build/reports`, `ddprof-test/build/reports`) rather than the entire `build/` trees, or narrow to `o+r` only where required.

.github/workflows/test_workflow.yml
```diff
@@ -580,6 +587,9 @@ jobs:
              \"${{ github.sha }}\" \"musl/${{ matrix.java_version }}-${{ matrix.config }}-aarch64\" \
              \"${{ matrix.config }}\" \"${{ env.JAVA_HOME }}\" \"${{ env.JAVA_TEST_HOME }}\"
            "
+           # Fix permissions after Docker container execution (both top-level and subproject builds)
+           sudo chmod -R +r build/ ddprof-test/build/ 2>/dev/null || true
```
jbachorik: **[Sphinx Review — LOW]** The second permission-fix block (appended to `Generate Unwinding Report`) covers only `build/` and `ddprof-test/build/`, dropping `ddprof-lib/build/` which the first block covers. The unwinding-report container runs the build as root in the same way as the test container and can leave root-owned files under `ddprof-lib/build/`. On the success path `prepare_reports.sh` reads `ddprof-lib/build` (`find ddprof-lib/build -name 'libjavaProfiler.*'` and `cp -r ddprof-lib/build/tmp`), so files newly created as root there by the unwinding container are not re-chmod'd and may be silently skipped.

**Suggestion:** Align the second chmod block to cover `ddprof-lib/build/` as well, matching the first block, or consolidate both into a single reused step to keep the scope in sync.

r1viollet:
Thanks for the detailed review! Fixed the actionable issues:

**Silent failures (MEDIUM):** Removed `2>/dev/null` so chmod errors now appear in step logs.

**Inconsistent coverage (LOW):** Added `ddprof-lib/build/` to the second chmod block to match the first.

Re: `--user` flag - Can't use it here because the Alpine containers run `apk add` which requires root. The chmod-after approach is the cleanest solution given the constraint.

Re: world-readable on shared runners - These are test builds without secrets, so the exposure is minimal. If this becomes a concern we can narrow to specific report directories.

.github/workflows/test_workflow.yml
```diff
@@ -572,6 +572,13 @@ jobs:
              echo "musl-${{ matrix.java_version }}-${{ matrix.config }}-aarch64" >> failures_musl-${{ matrix.java_version }}-${{ matrix.config }}-aarch64.txt
              exit 1
            fi
+       - name: Fix permissions for artifact upload
+         if: always()
+         run: |
+           # Files created by Docker container may have restrictive permissions
+           # that prevent GitHub Actions from reading them during artifact creation
+           sudo chmod -R +r build/ ddprof-test/build/ ddprof-lib/build/ || true
```
rkennke: Since `--user` is off the table and the chmod is staying, one robustness nit on this line: symbolic `chmod +r` with no `u`/`g`/`o` is treated as `a+r` *minus the umask bits*. On a standard runner (umask `022`) this sets `o+r` and works, but on these self-hosted `ARM LINUX SHARED` runners root's umask is not guaranteed — if it masks other-read, `sudo chmod +r` won't set the `o+r` bit the (non-root) runner user actually needs, and `|| true` would hide it, reproducing the exact EACCES this step is meant to fix.

Suggestion — make it explicit and fold in the directory-execute pass:

```sh
sudo chmod -R a+rX build/ ddprof-test/build/ ddprof-lib/build/ || true
```

`a+rX` is umask-independent, and the capital `X` adds execute only to directories (and already-executable files), so it also replaces the separate `find -type d -exec chmod +x` line. This also dovetails with the world-readable concern raised above: if you'd prefer not to go world-readable, `o+rX` is the minimal grant that still lets the runner read the tree.

r1viollet:
Great catch on the umask issue! Switched to `a+rX` which is umask-independent and consolidates both chmod commands (capital X adds execute only to directories). Comments updated to explain the approach.

.github/workflows/test_workflow.yml
r1viollet: I think we could have a --user if we refactored some of this. Though I did want to get CI to a green state before we consider this.

rkennke (approved):
Thank you, looks good now!

PR #624, opened 2026-06-30
test: thread lifecycle × profiler-state coverage

jbachorik:
**What does this PR do?**:
Adds fuzzing tests for three previously-uncovered cells in the thread-lifecycle × profiler-state matrix:

| State \ Event | Startup | Running | Teardown |
|---|---|---|---|
| **Enabled** | ✅ existing | ✅ existing | ✅ existing |
| **Disabling** | — | — | ✅ **new** (`StopWithChurningThreadsTest`) |
| **Enabling** | ✅ **new** (`StartWithChurningThreadsTest`) | — | — |
| **Disabled** | — | — | ✅ **new** (`IdleThreadLifecycleTest`) |

Also adds a C++ unit test (`ThreadFilterLifecycle`) for three `ThreadFilter` races: `unregisterThread` vs `clearActive` drain, `registerThread` vs concurrent `init()` toggle, and `unregisterThread` on a never-initialised filter.

**Motivation**:
Signal-handler teardown races during `stop()` and half-initialised engine state during `execute()` concurrent with JVMTI events are known hazard windows. The existing tests cover profiler-enabled churn but not the transition moments.

**Additional Notes**:
- Java tests use `CountDownLatch` to ensure thread churn is active before the profiler state transition, tightening the hazard window.
- Tests do not extend `AbstractProfilerTest` — they drive `execute()`/`stop()` timing directly (same pattern as `ShutdownTest`).
- C++ test is `#ifdef __linux__` only and runs clean under TSan/ASan.
- `@RetryTest(2)` + `@Timeout` guard against flakiness on slow CI.

**How to test the change?**:
- Java: `./gradlew :ddprof-test:testRelease -Ptests=StopWithChurningThreadsTest` (and the other two classes)
- C++: `./gradlew :ddprof-lib:gtestDebug_test_threadFilter_lifecycle`
- For race detection: `./utils/run-docker-tests.sh --config=tsan`

**For Datadog employees**:
- [ ] JIRA: [JIRA-XXXX]


ddprof-test/src/test/java/com/datadoghq/profiler/lifecycle/StartWithChurningThreadsTest.java
```diff
@@ -0,0 +1,120 @@
...
+                        t.join(5);
+                    } catch (InterruptedException e) {
+                        Thread.currentThread().interrupt();
+                        return;
+                    }
+                }
+            });
+        }
+        // Wait for churn to be active BEFORE calling execute()
+        assertTrue(churnRunning.await(10, TimeUnit.SECONDS), "Churn threads did not start");
```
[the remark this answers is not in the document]
jbachorik: Fixed in 4c6a5e10c — copyright header added, latch-timeout cleanup added to the finally block, and no-op `@RetryTest` removed.

ddprof-test/src/test/java/com/datadoghq/profiler/lifecycle/StopWithChurningThreadsTest.java
```diff
@@ -0,0 +1,115 @@
...
+                    try {
+                        t.join(5);
+                    } catch (InterruptedException e) {
+                        Thread.currentThread().interrupt();
+                        return;
+                    }
+                }
+            });
+        }
+        assertTrue(churnRunning.await(10, TimeUnit.SECONDS), "Churn threads did not start");
```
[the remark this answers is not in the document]
jbachorik: Fixed in 4c6a5e10c — copyright header added, latch-timeout cleanup added to the finally block, and no-op `@RetryTest` removed.

ddprof-test/src/test/java/com/datadoghq/profiler/lifecycle/StartWithChurningThreadsTest.java
```diff
@@ -0,0 +1,120 @@
...
+ * <p>Does NOT extend AbstractProfilerTest — needs direct control over
+ * execute()/stop() timing; uses the same pattern as ShutdownTest.
+ */
+public class StartWithChurningThreadsTest {
+
+    private static final int CHURN_THREADS = 64;
+    private static final int CYCLES = 50;
+    private static final String PROFILER_CMD = "start,wall=5ms,filter=*,jfr,file=";
+
+    @RetryTest(2)
```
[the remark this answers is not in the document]
jbachorik: Fixed in 4c6a5e10c — `@RetryTest` annotation removed from all new lifecycle tests.

ddprof-test/src/test/java/com/datadoghq/profiler/lifecycle/IdleThreadLifecycleTest.java
```diff
@@ -0,0 +1,61 @@
+package com.datadoghq.profiler.lifecycle;
```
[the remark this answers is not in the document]
jbachorik: Fixed in 4c6a5e10c — Datadog copyright header added to all new lifecycle test files.

jbachorik:
Both findings from the Codex review (#pullrequestreview-4608262693) are addressed in `d4debf917`:
- `filter=` added to `PROFILER_CMD` in both `StopWithChurningThreadsTest` and `StartWithChurningThreadsTest` — wall timer now delivers SIGVTALRM to all threads including churn workers.
- Churn worker futures collected and drained in the `finally` block — `ExecutionException` causes are propagated to the `errors` queue so worker failures are not silently swallowed.

ddprof-test/src/test/java/com/datadoghq/profiler/lifecycle/StartWithChurningThreadsTest.java
```diff
@@ -0,0 +1,131 @@
...
+                try {
+                    f.get();
+                } catch (ExecutionException e) {
+                    errors.add(e.getCause());
+                } catch (CancellationException ignored) {}
+            }
+        }
+
+        if (!errors.isEmpty()) {
+            fail(errors.poll());
```
kaahos: nit: `.poll()` returns only the head of the queue and not all the errors, I don't really think this is a big deal but it could be misleading for debug. if more than one thing throws (e.g. several churn workers), only the first gets reported and the rest are silently dropped.

ddprof-test/src/test/java/com/datadoghq/profiler/lifecycle/StopWithChurningThreadsTest.java
```diff
@@ -0,0 +1,136 @@
...
+                } catch (ExecutionException e) {
+                    if (!errors.offer(e.getCause())) {
+                        fail("Failed to record churn worker error", e.getCause());
+                    }
+                } catch (CancellationException ignored) {}
+            }
+        }
+
+        if (!errors.isEmpty()) {
+            fail(errors.poll());
```
kaahos: nit: same behaviour here.

kaahos (approved):
Looks good to me! I've just added a comment about a potential misleading behaviour but nothing big.

ddprof-test/src/test/java/com/datadoghq/profiler/lifecycle/StartWithChurningThreadsTest.java
jbachorik: Agreed. This would require re-working the error reporting, though. Calling `fail` will fail the test, so even if we drain the queue only the first error would be reported. 
Let me take a look how to do this properly - perhaps a synthetic exception subclass with composite reasons ...

jbachorik: Fixed in 5216effa6 — replaced `fail(errors.poll())` with a composite `AssertionError` that adds every collected `Throwable` as a suppressed exception, so all errors are reported in one `fail()` instead of only the first.

ddprof-test/src/test/java/com/datadoghq/profiler/lifecycle/StopWithChurningThreadsTest.java
jbachorik: Fixed in 52864e427 — same composite `AssertionError` approach applied here.

PR #625, opened 2026-07-01
fix(context): scope OTEL ThreadContext storage to the carrier thread (PROF-15271)

rkennke:
**What does this PR do?**:

Scopes OTEL `ThreadContext` storage to the **carrier** thread when running under virtual threads, replacing the plain `ThreadLocal` that keyed it by the virtual thread. Introduced as a selectable *mode* (`OtelContextStorage.Mode.CARRIER`) backed by `jdk.internal.misc.CarrierThreadLocal`, with a kill-switch and a safe fallback to today's behavior.

**Motivation**:

The `ThreadContext`'s `DirectByteBuffer` is a window into the OTEP record embedded in the **carrier's** native `ProfiledThread` — the record the (carrier-bound) sampler reads. Caching it in a plain `ThreadLocal` keys it by the *virtual* thread and pins it to whichever carrier was mounted at first use. This is:

- **Wrong** once the vthread migrates — writes land on the old carrier, so a sampler interrupting the *new* carrier sees stale/empty context (same family as the attribute-loss issue tracked in dd-trace-java #11646); and
- **Unsafe** once the old carrier's OS thread exits — `freeKey`/`release()` frees the `ProfiledThread` while the cached buffer keeps being written. That's a use-after-free into memory glibc may have handed back to the JVM, observed in the field as a SIGSEGV in `ThreadsSMRSupport::free_list` under a Loom workload on Corretto 26.

`CarrierThreadLocal.get()/set()/remove()` operate on the current carrier's map even from a mounted virtual thread, so a vthread always resolves to its current carrier's **live** record. Storage lifetime then matches the native record's lifetime, structurally eliminating the dangling-buffer window, and writes land where the sampler reads.

**Additional Notes**:

- **Java 8 baseline / unnamed module.** The lib compiles at `--release 8`, so the internal type can't be named at compile time. It's built reflectively once and held as its `ThreadLocal` supertype, so calls dispatch virtually with **no per-call reflection** and no compile-time dependency.
- **Mode selection** via `-Dddprof.context.storage.mode=auto|carrier|thread` (default `auto`):
  - `auto` — carrier scoping when available, else fall back to plain `ThreadLocal`. Fallback is silent on JDK < 21 (expected) and a **loud WARN** on JDK 21+ (where it's the unsafe pre-fix behavior).
  - `carrier` — carrier scoping is **required**: init **fails hard** (throws) if `CarrierThreadLocal` is inaccessible, rather than silently re-exposing the UAF. This is the fail-fast callers opt into once the runtime export is guaranteed.
  - `thread` — force legacy thread-scoped storage (kill-switch).
- **Why `auto` degrades instead of failing hard by default:** at init we only know the JVM is Loom-capable, not whether virtual threads actually route context here, and the runtime export carrier scoping needs is granted by the agent — which may load after this library. Failing hard by default would break profiler startup for **every** JDK 21+ deployment (including non-Loom apps) until that grant lands. So `auto` stays safe-and-loud; the tracer opts into `carrier` once it guarantees the export.
- **Not in this PR (follow-ups):**
  1. dd-trace-java grants `--add-exports java.base/jdk.internal.misc` to the profiler loader (via `Instrumentation.redefineModule`) and sets `mode=carrier` so `CARRIER` engages and fails fast in production.
  2. Reapply-on-mount / clear-on-unmount coherence (#11646) — this change makes that reapply *safe* (it writes to the live current carrier) but doesn't implement it.
- Sharing one `ThreadContext` per carrier is safe: its only non-buffer mutable state is the attribute cache, which is value-validated (`value.equals(attrCacheKeys[slot])`) with process-global encodings, so cross-vthread sharing costs at most a cache miss, never wrong data.

**How to test the change?**:

Automated (green on JDK 21 with carrier scoping active):
- `CarrierContextStorageTest` — 2000 virtual threads resolve to **~one `ThreadContext` per carrier** (distinct-context count ≪ vthread count), proving storage is not keyed by the virtual thread. Skips when carrier mode isn't active.
- `OtelContextStorageTest` — kill-switch forces plain `ThreadLocal`; `auto`/`carrier` select carrier when available; `carrier` **throws** when unavailable (fail-fast, runs on JDK < 21 CI).
- Existing `OtelContextStorageModeTest` passes unchanged (platform-thread isolation preserved).

The build adds `--add-exports java.base/jdk.internal.misc=ALL-UNNAMED` only on JDK ≥ 21 (the flag aborts a Java 8 JVM). The version is checked against the actual test JVM at task-execution time, so the split-JDK musl matrix (build JDK 21, test JDK 8) resolves correctly.

**For Datadog employees**:
- [ ] JIRA: [PROF-15271]

[PROF-15271]: https://datadoghq.atlassian.net/browse/PROF-15271?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
```diff
@@ -444,8 +467,29 @@ private static ThreadContext initializeThreadContext() {
...
+    }
+
+    /**
+     * Diagnostics/tests: the resolved OTEL context storage mode ({@code "CARRIER"} or
+     * {@code "THREAD"}), as selected by {@code -D}{@value OtelContextStorage#MODE_PROPERTY}
+     * and the availability of {@code jdk.internal.misc.CarrierThreadLocal}.
+     */
+    public String contextStorageMode() {
+        return OtelContextStorage.modeOf(tlsContextStorage).name();
     }
```
jbachorik, line 494: I wonder, can we just return `Mode` ?

ddprof-lib/src/main/java/com/datadoghq/profiler/OtelContextStorage.java
```diff
@@ -0,0 +1,165 @@
...
+    /**
+     * Kill-switch / selector system property: {@code auto} (default) | {@code carrier} | {@code thread}.
+     * <ul>
+     *   <li>{@code auto} — use {@link Mode#CARRIER} when available, else {@link Mode#THREAD}
+     *       (logged loudly on JDK 21+, where the fallback is unsafe under virtual threads).</li>
+     *   <li>{@code carrier} — require carrier scoping; {@link #create()} throws if unavailable.</li>
+     *   <li>{@code thread} — force legacy behavior (disables carrier scoping entirely).</li>
+     * </ul>
+     */
+    static final String MODE_PROPERTY = "ddprof.context.storage.mode";
```
jbachorik: This is probably to be used only for testing or debugging - having something like `ddprof.debug.*` or `ddprof.internal.*` would (hopefully) discourage users to try this haphazardly.

ddprof-lib/src/main/java/com/datadoghq/profiler/OtelContextStorage.java
```diff
@@ -0,0 +1,165 @@
...
+     * fail-fast behavior.
+     */
+    static ThreadLocal<ThreadContext> create() {
+        // Locale.ROOT: the values are ASCII keywords, so lower-casing must be locale-independent
+        // (a default-locale toLowerCase() maps "CARRIER" to "carrıer" under tr_TR, breaking the match).
+        String requested = System.getProperty(MODE_PROPERTY, "auto").trim().toLowerCase(Locale.ROOT);
+
+        if ("thread".equals(requested)) {
+            return new ThreadLocal<>();
+        }
```
jbachorik, line 116: Perhaps, this could be hoisted to class init? We are not really anticipating this value to change during the runtime, right? And not reparsing the system property for each new thread sounds nice :)

Also, the result could be saved in `Mode` variable so we don't have to do the string equals a few lines below.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
rkennke: Done in a42f6404 — `contextStorageMode()` now returns a typed `ContextStorageMode` (a new public top-level enum) instead of a `String`. The `OtelContextStorage` factory stays package-private, so only the value type is exposed as public API.

ddprof-lib/src/main/java/com/datadoghq/profiler/OtelContextStorage.java
rkennke: Agreed — renamed to `ddprof.debug.context.storage.mode` in a42f6404 (matching the existing `ddprof.debug.malloc_arena_max`), and the javadoc now notes it is an internal knob, not part of the supported config surface. Also updated the dd-trace-java side that sets/exports it to match.

ddprof-lib/src/main/java/com/datadoghq/profiler/OtelContextStorage.java
rkennke: Good to clarify: `create()` runs **once per `JavaProfiler` instance** — it is the `tlsContextStorage` field initializer, not a per-thread path. The per-thread get-or-init (`currentContext()`) never reads this property, so there is no per-thread reparse to eliminate; I added a comment saying so. I did tidy it (a42f6404): the selector is parsed once into locals so we do not repeat the string compares. I deliberately did **not** hoist it to a `static final`: the unit tests set the property and call `create()` repeatedly to exercise each mode, so a class-init cache would defeat them (and the agent sets the property before class-init anyway, so caching buys nothing).

jbachorik (approved):
Thanks! This was a sneaky one but it is nicely fixed.
LGTM!

PR #631, opened 2026-07-03
feat(context): all-native OTEL context write API — Phase 1 (PROF-15271)

rkennke:
## Summary

Phase 1 of eliminating the virtual-thread use-after-free in OTEL trace-context storage
(PROF-15271), by moving context writes to an **all-native** path that resolves the current
carrier's OTEP record inside the JNI call — where the mounted virtual thread is pinned to its
carrier by the native frame, so the write is **race-free by construction** with no cached
`DirectByteBuffer` to dangle.

This PR is **purely additive**: it adds the new API alongside the existing `DirectByteBuffer`
(DBB) path (now `@Deprecated`) and changes no behavior. The crash fix itself lands in **Phase 2**,
when dd-trace-java switches to the new API; until then carrier-scoping (#625 / `94686da`) remains
the mitigation.

Full rationale, options considered (A–E), the JMH campaign, the consumer audit, and the
expand → migrate → contract plan are in the design note:
**`doc/plans/2026-07-02-all-native-context-storage-design.md`**.

## What's in this PR

- **Native write primitives** (`javaApi.cpp`): `setTraceContext0`/`clearTraceContext0` (combined
  per-scope activate/deactivate) and `setContextValue0`/`clearContextValue0` (single attribute),
  using the same `detach → mutate → attach` `valid`-flag protocol as `ThreadContext` so the
  async-signal-handler sampler (`ContextApi::get`) stays coherent. The native path self-initializes
  the LRS `attrs_data` entry (no dependency on the `ThreadContext` ctor).
- **Public API + global value cache** (`JavaProfiler`, `ContextValueCache`): provisional names
  `setTraceContext`/`clearTraceContext`/`setContextValue`/`clearContextValue`. The
  `value → (encoding, utf8)` memoization moves out of the per-thread `ThreadContext` into a
  process-wide lock-free cache (safe because the encoding is a process-global immutable ID).
- **Deprecations**: the DBB write/read surface and the `ThreadContext`/`ScopeStack`/`ContextSetter`
  helper classes are marked `@Deprecated` (non-breaking) pointing at the new API.
- **Tests** (`AllNativeContextTest`, 9): round-trip, DBB-equivalence, clear semantics,
  span-transition attribute reset, single-attr set/clear, overflow, rejected values,
  DBB↔native coexistence, and coherence of native writes from 512 mounted virtual threads.
- **Perf-guard benchmark** (`ContextCombinedBenchmark`).

## Results

- **Tests**: all 9 new tests pass on JDK 21 (debug); the existing context suite (Otel / Carrier /
  Tag / ProcessContext) is unregressed.
- **Perf** (JDK 21, production per-scope activate+deactivate cycle, carrier mode): all-native
  **~136 ns vs DBB ~189 ns — ~28% faster** (~24% thread mode), a win on virtual threads too, and
  mode-independent. The DBB baseline is the faithful production cost (each of its 6 calls pays a
  `currentContext()`/`CarrierThreadLocal` lookup the 2-call native cycle avoids).
- **Perf on old runtimes (Java 8 & 11)**: same `ContextCombinedBenchmark` and JMH config as the
  JDK 21 campaign (3 forks × 3×1s warmup × 3×3s measurement, `AverageTime`, one op = one full
  activate+deactivate cycle). Only `platform_* / thread` applies on 8/11 — `carrier` mode and the
  `vthread_*` benchmarks are JDK 21+ only (they need `jdk.internal.misc.CarrierThreadLocal` /
  `Thread.startVirtualThread`).

  | JVM | native (`setTraceContext`+`clearTraceContext`) | dbb (`setContext`+2×attr, ×2) | delta |
  |-----|--------:|--------:|-------|
  | Java 8 (`1.8.0_492`)  | 110.31 ± 1.89 ns/op | 252.53 ± 1.39 ns/op  | native **−56.3%** (dbb ~2.29×) |
  | Java 11 (`11.0.31`)   | 168.85 ± 1.68 ns/op | 417.90 ± 11.45 ns/op | native **−59.6%** (dbb ~2.47×) |

  No degradation on old runtimes — the all-native cycle wins by ~2.3–2.5×, consistent with JDK 21.
  (Reliable signal is the native-vs-dbb ratio *within* each JVM; the cross-JVM absolute gap is
  noisier, as JMH itself warns.)

## Migration (expand → migrate → contract)

1. **Expand — this PR** (java-profiler): add the new API alongside the DBB path. Additive.
2. **Migrate** (dd-trace-java): switch `DatadogProfilingIntegration` to the combined API. **Crash
   fix + perf win land here.**
3. **Contract** (java-profiler): delete the DBB / `CarrierThreadLocal` / `OtelContextStorage`
   subsystem. Breaking, gated on all-consumers-migrated.

## Reviewer notes

- **Public API names are provisional** — the one Phase-1 item to coordinate with the dd-trace-java
  team so Phase 2 is a clean swap.
- **Signal-safety** is the key review area: the native `detach/attach` + release-fence protocol vs
  the sampler's acquire load of `valid` (`ContextApi::get`). Verify on aarch64 (trivial on x86).
- The full JMH campaign + native prototypes are preserved for reproducibility on branch
  **`context-storage-benchmark-experiment`**.
- Consumer audit: the DBB API's only production consumer is the dd-trace-java bridge; everything
  else is test/benchmark. Open item: confirm `ddprof-lib` standalone Maven publication (external
  consumers) before the Phase-3 removal.


ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -727,6 +727,244 @@ Java_com_datadoghq_profiler_JavaProfiler_initializeContextTLS0(JNIEnv* env, jcla
...
+// (non-zero); clearing is clearTraceContext0. Encodings are resolved by the Java value cache.
+extern "C" DLLEXPORT void JNICALL
+Java_com_datadoghq_profiler_JavaProfiler_setTraceContext0(JNIEnv* env, jclass unused,
+    jlong localRootSpanId, jlong spanId, jlong traceIdHigh, jlong traceIdLow,
+    jint slot0, jint enc0, jbyteArray utf0, jint slot1, jint enc1, jbyteArray utf1) {
+  ProfiledThread* thrd = ProfiledThread::current();
+  if (thrd == nullptr) {
+    return;
+  }
+  OtelThreadContextRecord* record = thrd->getOtelContextRecord();
```
jbachorik, line 868: The native write entry points (`setTraceContext0`/`clearTraceContext0`/`setContextValue0`/`clearContextValue0`) resolve the record via `getOtelContextRecord()` and set `valid=1`, but never call `ContextApi::initializeContextTLS()` (or otherwise mark `isContextInitialized()` / set the `otel_thread_ctx_v1` discovery pointer). `ContextApi::get()` (`context_api.cpp:45`) and the wallclock sampler (`wallClock.cpp:49`) both gate on `isContextInitialized()`, which today is only ever set via the DBB path (`initializeContextTLS0`, reached through `currentContext()`). So a thread that only calls the new all-native API — never touching `getThreadContext()` — writes a context the sampler silently ignores, and one eBPF can't discover either. The new tests don't catch this because each of them calls `getThreadContext()` first, which sets the flag as a side effect of the old path.

Fix options: (a) have the native entry points call `ContextApi::initializeContextTLS(thrd)` when not already initialized, before writing; or (b) add a test that exercises the native API on a fresh thread with no prior `getThreadContext()` call.

ddprof-lib/src/main/java/com/datadoghq/profiler/ContextValueCache.java
```diff
@@ -0,0 +1,94 @@
...
+
+import java.nio.charset.StandardCharsets;
+import java.util.concurrent.atomic.AtomicReferenceArray;
+
+/**
+ * Process-wide cache resolving an attribute value to its {@code (encoding, utf8)} pair for the
+ * all-native context write path, avoiding a JNI {@code registerConstant0} call on every write.
+ *
+ * <p>The encoding is a <em>process-global, immutable</em> ID: {@code registerConstant0} interns the
+ * value in the native Dictionary and returns the same ID for the JVM's lifetime. That immutability
```
jbachorik: `ContextValueCache` assumes native Dictionary encodings are immutable for the JVM's lifetime, but `Profiler::start(reset=true)` calls `StringDictionary::clearAll()`, which resets the ID counter to 1 (`stringDictionary.h:599`). `resetThreadContext()` only clears the old per-thread DBB caches — this new global cache on the `JavaProfiler` singleton is never touched. So after a stop/start cycle, a cache hit can hand back an encoding from the previous session that the new Dictionary has since reassigned to a different value, and `setContextValue0`/`setTraceContext0` write that stale encoding straight into the sampler-visible sidecar.

Can we invalidate/clear `ContextValueCache` from whichever restart path calls `clearAll()`, or tag entries with a generation counter so a stale generation is treated as a miss? The class javadoc should also be fixed — it currently states an invariant that doesn't hold across restart.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
```diff
@@ -45,6 +45,15 @@ static final class TSCFrequencyHolder {
...
     private final ThreadLocal<ThreadContext> tlsContextStorage = OtelContextStorage.create();
 
+    // Process-wide value->(encoding, utf8) cache for the all-native context write path
+    // (setTraceContext / setContextValue). See ContextValueCache. One instance on the singleton.
+    private final ContextValueCache contextValueCache = new ContextValueCache();
+
+    // Number of custom attribute slots on the all-native path. Mirrors the native
+    // DD_TAGS_CAPACITY (context.h) and ThreadContext.MAX_CUSTOM_SLOTS; kept local so the
+    // all-native API does not depend on the phase-3-removed ThreadContext for the bound.
+    private static final int MAX_CONTEXT_SLOTS = 10;
```
jbachorik: `MAX_CONTEXT_SLOTS` duplicates native `DD_TAGS_CAPACITY` and `ThreadContext.MAX_CUSTOM_SLOTS`, bound together only by a comment — nothing catches drift at compile time. Worth deriving the Java bound from one source (native accessor, or reuse `ThreadContext.MAX_CUSTOM_SLOTS`) before the phase-3 removal of `ThreadContext` leaves a stale constant behind.

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -727,6 +727,244 @@ Java_com_datadoghq_profiler_JavaProfiler_initializeContextTLS0(JNIEnv* env, jcla
...
+}
+
+// Copies at most 255 bytes from a Java byte[] into buf; returns the length copied (0 if arr null).
+// Callers (ContextValueCache.resolve) reject values whose UTF-8 exceeds 255 bytes before they
+// reach here, so an oversized array is a caller-contract violation, not an expected input. The
+// assert documents/enforces that in debug builds; the clamp is release-build defense that also
+// bounds the copy to the 256-byte caller stack buffer. Clamping mid-sequence could split a
+// multi-byte UTF-8 char, but is unreachable given the caller guard.
+static inline int otelReadUtf8(JNIEnv* env, jbyteArray arr, uint8_t* buf) {
+  if (arr == nullptr) {
```
jbachorik, line 824: Nit: no test exercises the `arr == nullptr` guard in `otelReadUtf8` — a mutant removing it wouldn't be caught. Worth a case that calls `setTraceContext0`/`setContextValue0` with a null `byte[]` value.

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -727,6 +727,244 @@ Java_com_datadoghq_profiler_JavaProfiler_initializeContextTLS0(JNIEnv* env, jcla
...
+}
+
+// Compacts out the attrs_data entry with the given OTEP key index; returns the new size.
+// Mirrors ThreadContext.compactOtepAttribute. Record must be detached.
+static int otelCompactAttr(OtelThreadContextRecord* record, int otepKeyIndex) {
+  int currentSize = record->attrs_data_size;
+  uint8_t* d = record->attrs_data;
+  int readPos = 0, writePos = 0;
+  bool found = false;
+  while (readPos + 2 <= currentSize) {
```
jbachorik, line 781: Nit: `otelCompactAttr`'s loop guard `readPos + 2 <= currentSize` isn't exercised at the exact 2-byte boundary (key + zero-length value) — existing tests only write substantial payloads. Worth adding that case.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
```diff
@@ -224,19 +233,104 @@ public void setContext(long spanId, long rootSpanId) {
...
+     * {@code http.route}). Returns false if the value is null, its UTF-8 exceeds 255 bytes, the
+     * native Dictionary is full, or {@code slot} is out of range; on failure the slot is cleared.
+     *
+     * @param slot  custom attribute slot index
+     * @param value the attribute value; {@code null} clears the slot
+     * @return true if the value was written; false if it was null, oversized, the Dictionary is
+     *         full, or {@code slot} is out of range
+     */
+    public boolean setContextValue(int slot, CharSequence value) {
+        if (slot < 0 || slot >= MAX_CONTEXT_SLOTS) {
```
jbachorik: Nit: the `slot >= MAX_CONTEXT_SLOTS` check in `setContextValue` is only tested with slots 0-3; nothing pins the boundary at `slot == MAX_CONTEXT_SLOTS`. Worth adding that case.

ddprof-lib/src/main/cpp/javaApi.cpp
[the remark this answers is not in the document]
rkennke: Fixed in 95351e6fa. `setTraceContext0` and `setContextValue0` now call `ContextApi::initializeContextTLS(thrd)` on first use (guarded by `isContextInitialized()`), exactly as the DirectByteBuffer path does in `initializeContextTLS0`, so a thread that only ever uses the all-native API publishes `otel_thread_ctx_v1` and is visible to `ContextApi::get` / the wall-clock sampler.

Also added `AllNativeContextSamplingTest`, a wall-clock JFR test on a thread that deliberately never calls `getThreadContext()`; it asserts a sample carries the natively-written context. I verified it is a real regression guard: with the init calls backed out, ~1600 samples are still produced but none carry the context and the test fails.

rkennke: Fixed in 95351e6fa via option (a): both native write entry points now call `ContextApi::initializeContextTLS(thrd)` when `!isContextInitialized()`, before writing. Also did (b) — `AllNativeContextSamplingTest` exercises the all-native API on a thread that never calls `getThreadContext()` and checks the wall-clock sampler picks up the context. Confirmed the test fails when the fix is reverted (samples present, but spanId 0).

I left the two clear paths (`clearTraceContext0`/`clearContextValue0`) without the init call: a clear is only meaningful after an activation, which goes through a setter and initializes there; clearing a never-initialized record is a no-op.

ddprof-lib/src/main/java/com/datadoghq/profiler/ContextValueCache.java
rkennke: Fixed in 95351e6fa. Added `ContextValueCache.clear()` and invoke it from `JavaProfiler.execute()` whenever the command's action token is `start` — matched to the native reset trigger (`Profiler::start` with `reset`, i.e. `ACTION_START` only, not `resume`). A subsequent `resolve` re-registers against the new Dictionary. Also rewrote the class javadoc: the encoding is now documented as stable only *within* a recording session, with the session-boundary behaviour spelled out.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
rkennke: Addressed in 95351e6fa. I kept `MAX_CONTEXT_SLOTS` as a literal (it bounds slot checks that can run before the native library is loaded, so it can't be derived via JNI at class-init, and I wanted it independent of the phase-3-removed `ThreadContext`). To catch drift I added native accessor `maxContextSlots0()` returning `DD_TAGS_CAPACITY`, plus `MaxContextSlotsTest` which asserts `MAX_CONTEXT_SLOTS` equals both the native capacity and `ThreadContext.MAX_CUSTOM_SLOTS`. Drift now fails a test rather than only violating a comment.

ddprof-lib/src/main/cpp/javaApi.cpp
rkennke: Added in 95351e6fa (`AllNativeContextTest.nativeNullValueByteArrayIsTreatedAsEmpty`). The public API can't reach this branch (a null value maps to a slot clear), so the test calls the `setContextValue0` JNI primitive directly with a live slot and a null `byte[]`, asserting it is treated as an empty value rather than dereferenced.

ddprof-lib/src/main/cpp/javaApi.cpp
rkennke: Added in 95351e6fa (`AllNativeContextTest.zeroLengthAttributeCompactsCleanly`). It writes an empty-string value (a bare 2-byte key+len entry), then overwrites and clears it, so `otelCompactAttr` walks the exact 2-byte boundary via the `readPos + 2 <= currentSize` guard.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
rkennke: Added in 95351e6fa (`AllNativeContextTest.slotBoundaryIsRejected`): highest valid slot (capacity-1) is accepted, `slot == capacity` and `slot > capacity` are rejected. The capacity literal (10) is itself pinned to native `DD_TAGS_CAPACITY` by `MaxContextSlotsTest`.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
```diff
@@ -166,7 +177,22 @@ public String execute(String command) throws IllegalArgumentException, IllegalSt
...
             throw new NullPointerException();
         }
-        return execute0(command);
+        String result = execute0(command);
+        // A fresh 'start' (ACTION_START) resets the native Dictionary (StringDictionary::clearAll),
+        // reassigning encodings. Drop the value cache so no stale encoding from the prior session
+        // is reused. Matched to the native reset trigger: the action is the first token, and only
+        // 'start' (not 'resume') resets. See ContextValueCache.
+        if (isStartAction(command)) {
+            contextValueCache.clear();
```
kaahos, line 187: couldn't a race condition happen here? `contextValueCache.clear()` only runs after `execute0()` returns, but the dictionary reset already happened inside that call -- so there's a window where another thread's `resolve()` can still cache-hit a stale encoding before it is cleared. maybe worth a lock around both?

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
```diff
@@ -166,7 +177,22 @@ public String execute(String command) throws IllegalArgumentException, IllegalSt
...
+        }
+        return result;
+    }
+
+    /** True if the command's action (its first comma-separated token) is {@code start}. */
+    private static boolean isStartAction(String command) {
+        int comma = command.indexOf(',');
+        String action = (comma < 0 ? command : command.substring(0, comma)).trim();
+        return action.equals("start");
```
kaahos: I feel unsure about that. this doesn't match the usual way of parsing arguments as done in `ddprof-lib/src/main/cpp/arguments.cpp`. couldn't/shouldn't we only rely on `ddprof-lib/src/main/cpp/arguments.cpp` to parse user input, WDYT? otherwise, I'd say it should at least match the parsing logic present in `ddprof-lib/src/main/cpp/arguments.cpp`.

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -727,6 +727,256 @@ Java_com_datadoghq_profiler_JavaProfiler_initializeContextTLS0(JNIEnv* env, jcla
...
+  u32* enc = thrd->getOtelTagEncodingsPtr();
+
+  __atomic_store_n(&record->valid, (uint8_t)0, __ATOMIC_RELAXED);
+  __atomic_thread_fence(__ATOMIC_RELEASE);
+
+  enc[slot] = 0;
+  record->attrs_data_size = (uint16_t)otelCompactAttr(record, slot + 1);
+
+  __atomic_thread_fence(__ATOMIC_RELEASE);
+  __atomic_store_n(&record->valid, (uint8_t)1, __ATOMIC_RELAXED);
```
kaahos: I wonder if, from an API perspective, `clearContextValue0()` shouldn't preserve the previous valid state? since `clearTraceContext0()` intentionally leaves the record invalid, a later `clearContextValue0()` would currently publish it again with `valid=1`. maybe clearing one context value should not change whether the whole context record is valid, what do you think?

kaahos (commented):
really nice changes, thanks! I only have few minor questions you can find below.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
rkennke: Good point — reworked in a71cac9ca to not parse the command in Java at all. The native side is authoritative: `Profiler::start` sets a `_context_value_dict_reset` flag in the same reset block that runs `_context_value_map.clearAll()` (reached only when `args._action == ACTION_START`, per `arguments.cpp`). `execute()` then consumes it via `consumeContextDictionaryReset0()` and drops the cache. No Java-side tokenization, so it can't drift from `arguments.cpp`.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
rkennke: You're right that a window remains. After the rework (a71cac9ca) the clear is driven by the native reset flag and happens right after `execute0()` returns, but `clearAll()` fires inside that call, so a concurrent `resolve()` can still cache-hit a stale encoding in between.

My bar here is "no worse than the existing DirectByteBuffer API." That path invalidates its per-thread caches by recreating the `ThreadContext` on restart via `resetThreadContext()`, which is also not atomic with `clearAll()` (its cache comment even assumes the dictionary is never cleared) — so the same race class exists there, and is arguably wider since it depends on the tracer calling `resetThreadContext()`. So I think this meets the bar without adding a lock on the hot path.

I want to fully close the window (e.g. a native generation/epoch counter validated lock-free in `resolve()`) as a separate follow-up so it gets proper design attention rather than being rushed in here. I'll file a Jira to track it and link it here.

ddprof-lib/src/main/cpp/javaApi.cpp
rkennke: Agreed — fixed in a71cac9ca. `clearContextValue0` now captures the record's `valid` byte up front and restores it at the end instead of hardcoding `valid=1`, so clearing a single attribute preserves whatever state a preceding `clearTraceContext0()` left (it will not resurrect a deactivated, span-less record). This mirrors the guard the DBB path already has in `ThreadContext.setContextAttribute` ("never resurrect a cleared record"). Added `AllNativeContextTest.clearContextValuePreservesInvalidState`, which reads the `valid` byte back through the record buffer and asserts it stays 0 after a clearTraceContext → clearContextValue sequence.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
rkennke: Filed as PROF-15361 (relates to PROF-15271) to track fully closing the window: https://datadoghq.atlassian.net/browse/PROF-15361

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -727,6 +727,261 @@ Java_com_datadoghq_profiler_JavaProfiler_initializeContextTLS0(JNIEnv* env, jcla
...
+
+  enc[slot] = (u32)encoding;
+  jboolean ok = JNI_TRUE;
+  if (!otelReplaceAttr(record, slot + 1, buf, len)) {
+    enc[slot] = 0;
+    ok = JNI_FALSE;
+  }
+
+  __atomic_thread_fence(__ATOMIC_RELEASE);
+  __atomic_store_n(&record->valid, (uint8_t)1, __ATOMIC_RELAXED);
```
kaahos, line 967: here I’d apply similar changes as for `clearContextValue0()`. I’d recommend not setting `valid=1` unconditionally, but first checking whether the record was valid before the call, what do you think? otherwise `setContextValue0()` can republish a record cleared by `clearTraceContext0()` with zero trace/span IDs.

rkennke: Good catch — done in 6dd6efd9a, same idiom as `clearContextValue0`: `setContextValue0` now captures the record's `valid` byte up front and restores it at the end instead of hardcoding `valid=1`, so writing an attribute won't republish a record that `clearTraceContext0` deactivated (which would expose zero trace/span as valid). This matches the DBB path's guard in `ThreadContext.setContextAttributesByIdAndBytes` ("never resurrect a cleared, span-less record").

Added `AllNativeContextTest.setContextValuePreservesInvalidState`, which asserts `valid` stays 0 after a `clearTraceContext` → `setContextValue` sequence (and that the attribute is not observable).

kaahos (approved):
Thanks for the modifications! it looks good to me.

doc/plans/2026-07-02-all-native-context-storage-design.md
```diff
@@ -0,0 +1,316 @@
+# Design note: all-native OTEL context storage (eliminating the carrier-owned DirectByteBuffer)
```
jbachorik, line 1: I have no strong opinion on committing the plan docs - being guilty of the same ...
This is more of a foot-note - we should think about how avoid those plans becoming something that we never come back to and they would just rot away. 
Anyway, nothing for this PR to solve.

rkennke:
@kaahos thanks for your review. I needed to partly revert the last changes to preserve the (correct) behaviour of the DBB implementation (see https://github.com/DataDog/java-profiler/pull/631/commits/5beb6987c03cd433d4995c50f9664c1e9d462e73), and I also tightened the argument validation (see https://github.com/DataDog/java-profiler/pull/631/commits/9908c585aff1f3e85ff2645f633be197c1b16348). Can you re-check those?

jbachorik (approved):
Looks really good.
Could you, for peace of my mind, run the benchmarks also on Java 8 and 11 and add the numbers to the PR?
I expect maybe slight degradation, but we won't try and optimize for old runtimes anyway ...

kaahos (approved):

doc/plans/2026-07-02-all-native-context-storage-design.md
rkennke: I'll remove that plan and any references to WIP (e.g. this will be done in phase X) when the DBB removal lands.

rkennke:
@jbachorik ran the `ContextCombinedBenchmark` perf-guard on Java 8 and 11 as requested — same JMH config as the JDK 21 campaign (3 forks × 3×1s warmup × 3×3s measurement, `AverageTime` ns/op, one op = one full activate+deactivate cycle). Good news: no degradation on the old runtimes — the all-native path is a decisive win over the DBB path.

Only the `platform_*` methods and `thread` storage mode are applicable on 8/11:
- `vthread_*` benchmarks are JDK 21+ only (`Thread.startVirtualThread`).
- `carrier` storage mode is JDK 21+ only — it requires `jdk.internal.misc.CarrierThreadLocal` (Loom), which doesn't exist on 8/11, so it correctly errors out there (N/A, not a regression).

**thread mode, ns/op (± 99.9% CI), 9 measurements each:**

| JVM | native (`setTraceContext`+`clearTraceContext`) | dbb (`setContext`+2×attr, ×2) | delta |
|-----|--------:|--------:|-------|
| Java 8  (`1.8.0_492`)  | 110.31 ± 1.89 | 252.53 ± 1.39  | native −56.3% (dbb ~2.29×) |
| Java 11 (`11.0.31`)    | 168.85 ± 1.68 | 417.90 ± 11.45 | native −59.6% (dbb ~2.47×) |

So the combined 2-JNI-call cycle beats the 6-call DBB sequence by ~2.3–2.5× on both — consistent with the JDK 21 result, no slight degradation observed. (Caveat: the reliable signal is the native-vs-dbb ratio *within* each JVM; the absolute 8-vs-11 gap is cross-JVM and noisier, as JMH itself warns.)

jbachorik:
@rkennke Cool, thanks!

PR #639, opened 2026-07-07
Adopt ThreadLocal to store ProfiledThread

zhengyu123:
**What does this PR do?**:
Replaces ProfiledThread's hand-rolled pthread_key_t + pthread_once/atomic-flag TLS management with the existing `ThreadLocal<T, CREATE_FUNC, CLEAN_FUNC>` template, and closes two related gaps that fell out of that consolidation:

`ProfiledThread::current()` is now the single signal-safe accessor (never allocates, returns existing TLS or nullptr), replacing the old split between `current()` (lazy-allocating) and `currentSignalSafe()` (non-allocating). `currentSignalSafe()` is removed; all 39 call sites across 16 files were renamed to `current()`.
`initCurrentThread() / initCurrentThreadSignalSafe()` are the allocating variants and now return the `ProfiledThread*` (or `nullptr`) instead of `void`, so callers can check the result instead of asserting.
`ProfiledThread::isThreadKeyValid()` is a new, cheap validity check backed by the ThreadLocal's own key state.
`JVMSupport::initialize() / isInitialized()` now check `isThreadKeyValid()` for real, closing a previously-stubbed gap ("Add ProfiledThread key checking here in next PR").
`Profiler::init()` is new — called from `JavaProfiler_init0` before `VM::initProfilerBridge()` — and runs `checkState()` under `_state_lock` before calling `initCurrentThreadSignalSafe()`, closing a startup race where native TLS init could otherwise run ahead of a failed profiler state check.
JNI hot paths in javaApi.cpp (`filterThreadAdd0/filterThreadRemove0`) now return early on a null `current()` instead of relying on a debug-only assert; a new `initOrGetCurrentThread()` helper covers the early-startup TLS race in` initializeContextTLS0`.
Native-thread wrappers in libraryPatcher_linux.cpp were updated to consume the new `initCurrentThread()` return value with explicit null checks.
**Motivation**:
The old TLS implementation duplicated logic that ThreadLocal already provides (key creation, lazy allocation, signal-safe clear-then-free ordering), and left two known gaps: JVMSupport never actually validated the ProfiledThread key, and native-thread/JNI call sites relied on assert (a no-op in release builds) instead of real null checks — both closed here.

**Additional Notes**:

**How to test the change?**:

Existing gtest suites (thread_teardown_safety_ut.cpp, ddprof_ut.cpp, hotspot_crash_protection_ut.cpp, jvmSupport_ut.cpp) were updated for the renamed/reshaped API, plus a new regression test, JvmSupportInitFailureTest.InitReturnsErrorAndLatchesErrorState, covering Profiler::init()'s error path when JVMSupport::initialize() fails.
**For Datadog employees**:
- [X] JIRA: [PROF-15272](https://datadoghq.atlassian.net/browse/PROF-15272)

[PROF-15272]: https://datadoghq.atlassian.net/browse/PROF-15272?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddprof-lib/src/main/cpp/threadLocalData.cpp
```diff
@@ -11,57 +11,34 @@
...
+    return nullptr;
+  }
+  ProfiledThread* tls = _current_thread.get();
+  if (tls == nullptr) {
+    int tid = OS::threadId();
+    tls = ProfiledThread::forTid(tid);
+    _current_thread.set(tls);
   }
+  return tls;
 }
```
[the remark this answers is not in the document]
zhengyu123: As comment mentioned, caller should block signals.

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -168,7 +192,10 @@ JavaCritical_com_datadoghq_profiler_JavaProfiler_filterThreadAdd0() {
 extern "C" DLLEXPORT void JNICALL
 JavaCritical_com_datadoghq_profiler_JavaProfiler_filterThreadRemove0() {
   ProfiledThread *current = ProfiledThread::current();
-  assert(current != nullptr);
+  if(current == nullptr) {
+    return;
+  }
```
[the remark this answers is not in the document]
zhengyu123: `filterThreadAdd0()` is responsible for creating TLS.

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -685,7 +713,8 @@ Java_com_datadoghq_profiler_OTelContext_readProcessCtx0(JNIEnv *env, jclass unus
 
 extern "C" DLLEXPORT jobject JNICALL
 Java_com_datadoghq_profiler_JavaProfiler_initializeContextTLS0(JNIEnv* env, jclass unused, jlongArray metadata) {
-  ProfiledThread* thrd = ProfiledThread::current();
+  // Assume this call is not on hot path
+  ProfiledThread* thrd = initOrGetCurrentThread();
   assert(thrd != nullptr);
```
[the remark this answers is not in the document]
zhengyu123: Pre-existing: suggested is a behavior change that is beyond this PR.

ddprof-lib/src/test/cpp/thread_teardown_safety_ut.cpp
```diff
@@ -129,13 +129,11 @@ static void *t02_body(void *) {
...
-      << "currentSignalSafe() must return null in the TLS-clear/delete window";
+      << "current() must return null in the TLS-clear/delete window";
 
   // release() with TLS already null must not double-free.
   ProfiledThread::release();
-  // Complete the simulated teardown: delete the object (mirrors what freeKey
-  // would do). Destructor is private so we need the test helper.
-  ProfiledThread::deleteForTest(detached);
+  EXPECT_EQ(nullptr, ProfiledThread::current());
   return nullptr;
```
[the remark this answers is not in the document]
zhengyu123: Fixed

ddprof-lib/src/main/cpp/threadLocalData.cpp
```diff
@@ -11,57 +11,34 @@
...
-  }
-  pthread_key_t key = _tls_key;
-  ProfiledThread *tls = (ProfiledThread *)pthread_getspecific(key);
-  if (tls != NULL) {
-    SignalBlocker blocker;
-    pthread_setspecific(key, NULL);
-    delete tls;
-  }
+  _current_thread.clear();
 }
```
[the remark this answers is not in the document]
zhengyu123: Pre-exist

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -68,6 +69,12 @@ class JniString {
 
 extern "C" DLLEXPORT jboolean JNICALL
 Java_com_datadoghq_profiler_JavaProfiler_init0(JNIEnv *env, jclass unused) {
+  Error error = Profiler::instance()->init();
+  if (error) {
+    throwNew(env, "java/lang/IllegalStateException", error.message());
+    return JNI_FALSE;
+  }
+
```
jbachorik, line 77: Could this be moved to `initProfilerBridge` ?
That function checks if the shared library can be proberly loaded and initialized and whether jvmlib is accesible (among other things). These checks look like they should prevent prematurely initializing profiler.

ddprof-lib/src/main/cpp/jvmSupport.cpp
```diff
@@ -95,7 +95,7 @@ int JVMSupport::asyncGetCallTrace(ASGCT_CallFrame *frames, int max_depth, void*
         return 0;
     }
 
-    AsyncSampleMutex mutex(ProfiledThread::currentSignalSafe());
+    AsyncSampleMutex mutex(ProfiledThread::current());
```
jbachorik, line 104: This is more of a musing - the code is equivalent to the old implementation, but I wonder what happens when `current()` returns null. Probably SEGFAULT, right?

ddprof-lib/src/main/cpp/threadLocalData.cpp
```diff
@@ -11,57 +11,35 @@
...
-  // This MUST happen here to prevent lazy allocation in signal handler
-  initTLSKey();
-
-  if (pthread_getspecific(_tls_key) != NULL) {
-    return; // Already initialized
-  }
+ProfiledThread* ProfiledThread::initCurrentThreadSignalSafe() {
+  SignalBlocker blocker;
+  return initCurrentThread();
+}
```
jbachorik: Can we always protect initing current thread by the signal blocker?
I checked the code and the only place where 'naked' `initCurrentThread()` is used itself uses `SignalBlocker`.

ddprof-lib/src/main/cpp/threadLocalData.h
```diff
@@ -108,32 +105,43 @@ class ProfiledThread : public ThreadLocalData {
...
+    ProfiledThread* pt = _current_thread.get();
+    _current_thread.set(nullptr);
+    return pt;
   }
   // Deletes a ProfiledThread returned by clearCurrentThreadTLS().
   // Needed because the destructor is private.
   static void deleteForTest(ProfiledThread *pt) { delete pt; }
 #endif
+  // initCurrentThread() and release() are not async-signal-safe: 
+  // must call outside of a signal handler with signal blocked
```
jbachorik: `must call` -> `must be called`

ddprof-lib/src/main/cpp/threadLocalData.h
```diff
@@ -108,32 +105,43 @@ class ProfiledThread : public ThreadLocalData {
...
+  // be interrupted by signals
+  static ProfiledThread* initCurrentThreadSignalSafe();
 
-  inline int tid() { return _tid; }
+  // Signal-handler friendly (no allocation): returns existing TLS or nullptr.
+  static inline ProfiledThread *current() {
+    if (!isThreadKeyValid()) {
+      return nullptr;
+    }
+    return _current_thread.get();
```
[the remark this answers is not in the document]
jbachorik: I don't know where it found `currentTid()` in this particular change but it raises a valid point. Can you add a comment about how the TLS would get installed on already running threads?

jbachorik (changes requested):
I have some comments where changes or clarification might be necessary

ddprof-lib/src/main/cpp/jvmSupport.cpp
zhengyu123: No. When `current == nullptr`, it results in failing to acquire the mutex (https://github.com/DataDog/java-profiler/blob/main/ddprof-lib/src/main/cpp/asyncSampleMutex.h#L23). In this case, it results in early return.

ddprof-lib/src/main/cpp/javaApi.cpp
zhengyu123: Unfortunately, no.

In theory, profiler can only call into Java after JVMTI::VMInit() callback, and `initProfilerBridge()` is called as early as in `Agent_OnLoad()` - this is the source of startup race :-(

ddprof-lib/src/main/cpp/threadLocalData.h
zhengyu123: Fixed

ddprof-lib/src/main/cpp/threadLocalData.cpp
zhengyu123: I think, at this point, all `initCurrentThread()` calls are protected by signal blocker, either by callers, or going through `initCurrentThreadSignalSafe()` - I wish there was a way to `assert` that. But let me know if you find any exceptions.

jbachorik: My point is that we don't need two paths - it's always protected, so we should have just one path with mandated protection.

ddprof-lib/src/main/cpp/javaApi.cpp
jbachorik: Hmm, in that case, should not we make same precautions for initializing profiler as we do in `initProfilerBridge` - at least sanity checks for the libjavaProfiler and libjvm which are done there. Otherwise we might be trying (and maybe even succeeding) to initialize profiler for something which is not supporting it and nothing will work.

ddprof-lib/src/main/cpp/threadLocalData.h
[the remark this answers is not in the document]
zhengyu123: `currentTid()` fallback to `OS::threadId()` if `current == nullptr`.

I have not added back priming in this PR, certainly can use JNI entries to initialize TLS. Added `TODO` comment for `initOrGetCurrentThread()`.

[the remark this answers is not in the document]
jbachorik: Worth keeping this in mind and tracking the fix in JIRA - IIRC, the `current()`, when used from things like allocation profiler, did lazily initialized `ProfiledThread` even for threads that were active before profiler attached. This will be gone now so we might be missing threads from profiles randomly.

ddprof-lib/src/main/cpp/threadLocalData.h
```diff
@@ -108,32 +105,43 @@ class ProfiledThread : public ThreadLocalData {
...
-  static ProfiledThread *currentSignalSafe(); // Signal-safe version that never allocates
-  static int currentTid();
+  // This version blocks signals, so that initialization cannot
+  // be interrupted by signals
+  static ProfiledThread* initCurrentThreadSignalSafe();
 
-  inline int tid() { return _tid; }
+  // Signal-handler friendly (no allocation): returns existing TLS or nullptr.
+  static inline ProfiledThread *current() {
+    if (!isThreadKeyValid()) {
```
rkennke, line 140: **`isThreadKeyValid()` sentinel doesn't cover the pre-construction and post-destruction states of the static `_current_thread`**

`current()` now gates purely on `isThreadKeyValid()`, i.e. `_current_thread._key != INVALID_KEY` (`INVALID_KEY == (pthread_key_t)-1`). Because `_current_thread` is a namespace-scope **static**, that sentinel is wrong in two windows:

1. **Before the `ThreadLocal` constructor runs.** Static storage is zero-initialized first, so `_key == 0`, and `0 != (pthread_key_t)-1` → `isThreadKeyValid()` returns `true` before the key is created. `current()` then calls `pthread_getspecific(0)` on an uncreated key (UB where keys don't start at 0).
2. **After the destructor runs.** `~ThreadLocal()` calls `pthread_key_delete(_key)` but never resets `_key` to `INVALID_KEY`, so `isThreadKeyValid()` still returns `true` afterward and a late `current()` (live thread / signal during `dlclose`/process teardown) does `pthread_getspecific` on a **deleted** key (POSIX: undefined).

This is a regression relative to the old implementation, which used a dedicated `_tls_key_initialized` flag (false until `pthread_key_create` succeeded) and **never deleted the key** — so `currentSignalSafe()` returned `nullptr` in the pre-init window and there was no post-delete window at all. Adopting `ThreadLocal` for `ProfiledThread` newly exposes both, in the exact startup/teardown-race territory this PR is hardening.

Reachability is admittedly low (window 1 needs a cross-TU static-init-order path — none exists today since dynamic initializers run at `dlopen` before any JNI/signal entry; window 2 needs a thread still profiling during teardown), so I'd rank this medium-low. But it's a real defect and cheap to close: set `_key = INVALID_KEY` in `~ThreadLocal()`, and/or make the validity check robust to the zero-initialized state. Root cause is in `threadLocal.h` (not in this diff), but this PR is what puts `ProfiledThread` behind it.

ddprof-lib/src/main/cpp/threadLocalData.cpp
zhengyu123: You mean moving `SignalBlocker` to caller? - that seems a quite duplicated code.

ddprof-lib/src/main/cpp/threadLocalData.h
```diff
@@ -108,32 +105,43 @@ class ProfiledThread : public ThreadLocalData {
...
+      return nullptr;
+    }
+    return _current_thread.get();
```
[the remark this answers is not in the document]
zhengyu123: Hmmm... I thought I fixed them :-(

ddprof-lib/src/main/cpp/threadLocalData.cpp
jbachorik: No.

My point is that we already have places where we call `initCurrentThread()` and manually put the signal blocker:
- https://github.com/DataDog/java-profiler/blob/d59a7c2f1f2bcafeefabfaa28c47c905117e6e93/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp#L204
- https://github.com/DataDog/java-profiler/blob/d59a7c2f1f2bcafeefabfaa28c47c905117e6e93/ddprof-lib/src/main/cpp/libraryPatcher_linux.cpp#L258

The only remaining place, where we are doing `initCurrentThread()` (not in tests) without signal blocker is https://github.com/DataDog/java-profiler/blob/d59a7c2f1f2bcafeefabfaa28c47c905117e6e93/ddprof-lib/src/main/cpp/perfEvents_linux.cpp#L181 - and, arguably, that one also should/could be run with signal blocker.

This leaves the tests as the sole users of `initCurrentThread()`, at which point I think it makes sense to make `initCurrentThread()` just use the signal blocker; and fix the places where we are putting the signal blocker manually

ddprof-lib/src/main/cpp/threadLocalData.h
zhengyu123: 1. The sample in `threadLocal.h` states, `ThreadLocal` variables should be declared `static` - which is true for all our `ThreadLocal`s, that means the constructors should be ran by linker, before profiler's `pthread_create()` interceptor and signal handler can run, so I don't see it can happen.
2. `~ThreadLocal()` is only called by c++ runtime, since the instances are static - I am not completely sure the scenario can happen. If so, then resetting `_key` does not really help, because it is a race. However, we can workaround it by not deleting the key, assume the profiler is immortal.

jbachorik: > assume the profiler is immortal

Yes, this is a valid assumptions, at least in real-life usage. We never stop/unload profiler unless JVM is going down. Might be used differently in tests, though - worthwhile to check.

zhengyu123: Okay, no longer deleting the keys.

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -68,6 +69,12 @@ class JniString {
 
 extern "C" DLLEXPORT jboolean JNICALL
 Java_com_datadoghq_profiler_JavaProfiler_init0(JNIEnv *env, jclass unused) {
+  Error error = Profiler::instance()->init();
```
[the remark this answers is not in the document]
zhengyu123: Added check that only allows `Profiler` to be initialized on regular Java thread.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
```diff
@@ -115,6 +140,11 @@ public static synchronized JavaProfiler getInstance(String libLocation, String s
         if (!result.succeeded) {
             throw new IOException("Failed to load Datadog Java profiler library", result.error);
         }
+
+        if (isVirtualThread(Thread.currentThread())) {
+            throw new IOException("Cannot intialize profiler on an virtual thread");
```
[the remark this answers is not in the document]
zhengyu123: There are only 2 paths lead to Profiler initialization:
- JavaProfiler.getInstance() - now checking calling thread is not an virtual thread
- VMInit callback - it is known that calling thread is *not* an virtual thread.

ddprof-lib/src/main/cpp/threadLocalData.h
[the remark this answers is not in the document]
zhengyu123: They are new methods after merging.

ddprof-lib/src/main/cpp/javaApi.cpp
zhengyu123: Added comment - `JVMSupport::initialize()`

ddprof-lib/src/main/cpp/threadLocalData.cpp
```diff
@@ -11,57 +11,44 @@
...
-
-void ProfiledThread::doInitTLSKey() {
-  pthread_key_create(&_tls_key, freeKey);
-  // Must be set AFTER pthread_key_create so signal handlers see a valid key.
-  // Store-release pairs with the acquire loads in currentSignalSafe() and release()
-  // to prevent hardware load-load reordering on weakly-ordered architectures (aarch64):
-  // a plain volatile write is not sufficient there.
-  __atomic_store_n(&_tls_key_initialized, true, __ATOMIC_RELEASE);
-}
+ThreadLocal<ProfiledThread*, nullptr, ProfiledThread::freeValue>  ProfiledThread::_current_thread;
```
rkennke, line 20: Suggest documenting why the plain (non-atomic) `_key` read in `current()`/`isThreadKeyValid()` is safe now that the old acquire/release flag is gone — it hinges on eager static init ordering, which is load-bearing and non-obvious:

```suggestion
// Namespace-scope static: the ctor (pthread_key_create) runs during library
// load — single-threaded, before profiling signals or the pthread_create
// interceptor are armed. So current()/isThreadKeyValid() may read _key without
// synchronization: the ctor's write happens-before any later thread/signal that
// reads it.
ThreadLocal<ProfiledThread*, nullptr, ProfiledThread::freeValue>  ProfiledThread::_current_thread;
```

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
rkennke: Typo nit: "intialize" → "initialize", "an virtual" → "a virtual".

```suggestion
            throw new IOException("Cannot initialize profiler on a virtual thread");
```

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -74,8 +74,12 @@ static CTimer ctimer;
 static CTimerJvmti ctimer_jvmti;
 
 void Profiler::onThreadStart(jvmtiEnv *jvmti, JNIEnv *jni, jthread thread) {
-  ProfiledThread::initCurrentThread();
-  ProfiledThread *current = ProfiledThread::current();
+  // JVMTI callback - outside signal handeler
```
rkennke: Typo nit: "handeler" → "handler".

```suggestion
  // JVMTI callback - outside signal handler
```

rkennke (commented):
Re-review follow-ups — non-blocking; posting inline suggestions for the trivial nits.

**Behavior change worth tracking (not blocking):** `current()` no longer lazily installs `ProfiledThread` TLS. A thread already running at attach that never hits a JVMTI callback (`ThreadStart`/`ClassLoad`/`ClassPrepare`/GC/`SampledObjectAlloc`) or a JNI entry point will now see null TLS in the sampler and drop wall/CPU samples as `SAMPLES_DROPPED_THREAD_LOCAL`, whereas the old lazy `current()` would have installed it on first touch. The added init calls cover the common paths; suggest tracking the remaining gap as a follow-up.

The virtual-thread `getInstance()` guard is fine to keep as-is — the only fully-correct alternative is running `init0()` on a throwaway platform thread (the JVM Thread-key scan only resolves on a carrier), which isn't worth the added complexity for that rare first-touch path.

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -93,7 +97,8 @@ void Profiler::onThreadStart(jvmtiEnv *jvmti, JNIEnv *jni, jthread thread) {
 }
 
 void Profiler::onThreadEnd(jvmtiEnv *jvmti, JNIEnv *jni, jthread thread) {
-  ProfiledThread *current = ProfiledThread::currentSignalSafe();
+  // Thread teardown: no point to intialize TLS if not yet initialized
```
rkennke: Typo nit: "intialize" → "initialize".

```suggestion
  // Thread teardown: no point to initialize TLS if not yet initialized
```

ddprof-lib/src/main/cpp/hotspot/jitCodeCache.cpp
```diff
@@ -19,11 +20,13 @@ void JNICALL JitCodeCache::CompiledMethodLoad(jvmtiEnv *jvmti, jmethodID method,
                                 jint map_length,
                                 const jvmtiAddrLocationMap *map,
                                 const void *compile_info) {
+  ProfiledThread::initCurrentThreadSignalSafe();
```
jbachorik, line 23: Could you please add a small explanatory comment why `initCurrentThreadSignalSafe` is used here? I assume to try and initialize the JIT threads, but it would be good to have this explicitly stated.

ddprof-lib/src/main/cpp/threadLocalData.cpp
```diff
@@ -11,57 +11,49 @@
...
+  if (!isThreadKeyValid()) {
+    return nullptr;
+  }
 
-  if (pthread_getspecific(_tls_key) != NULL) {
-    return; // Already initialized
+  ProfiledThread* cur = current();
+  if (cur == nullptr) {
+    SignalBlocker blocker;
+    return initCurrentThread();
```
jbachorik, line 43: Could we just inline the code from `initCurrentThread()` and call `initCurrentThreadSignalSafe()` everywhere? Are there places where we MUST have the form without signal blocker?

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -74,8 +74,12 @@ static CTimer ctimer;
 static CTimerJvmti ctimer_jvmti;
 
 void Profiler::onThreadStart(jvmtiEnv *jvmti, JNIEnv *jni, jthread thread) {
-  ProfiledThread::initCurrentThread();
-  ProfiledThread *current = ProfiledThread::current();
+  // JVMTI callback - outside signal handler
+  ProfiledThread* current = ProfiledThread::initCurrentThreadSignalSafe();
+  if (current == nullptr) {
+    return;
```
jbachorik, line 80: This should be very rare, right? Can we get diag counters for this? We are going to drop the thread from profiling eligibility so it would be good to have some data how frequently this would happen.

jbachorik (approved):
Looking good. I have just small questions/suggestions but they are not blocking.

zhengyu123:
> Re-review follow-ups — non-blocking; posting inline suggestions for the trivial nits.
> 
> **Behavior change worth tracking (not blocking):** `current()` no longer lazily installs `ProfiledThread` TLS. A thread already running at attach that never hits a JVMTI callback (`ThreadStart`/`ClassLoad`/`ClassPrepare`/GC/`SampledObjectAlloc`) or a JNI entry point will now see null TLS in the sampler and drop wall/CPU samples as `SAMPLES_DROPPED_THREAD_LOCAL`, whereas the old lazy `current()` would have installed it on first touch. The added init calls cover the common paths; suggest tracking the remaining gap as a follow-up.
> 
> The virtual-thread `getInstance()` guard is fine to keep as-is — the only fully-correct alternative is running `init0()` on a throwaway platform thread (the JVM Thread-key scan only resolves on a carrier), which isn't worth the added complexity for that rare first-touch path.

@rkennke 
- `current()` is replacement of `currentSignalSafe()` in old code - no allocation, signal-safe
- Lazy installation of `ProfiledThread`, now, is done via `initCurrentThreadSignalSafe()`, and the call is added to `ThreadStart`/`ClassLoad`/`ClassPrepare`/GC/`SampledObjectAlloc` and etc. JNI entry points, that are sure the callers are fully initialized `JavaThread`.
- `virtual-thread` guard is needed for resolving JVM Thread-key - if the caller thread is a virtual-thread, it cannot resolve native `JavaThread*` from `jthread`, the field only applies physical thread `jthread`

ddprof-lib/src/main/cpp/hotspot/jitCodeCache.cpp
zhengyu123: Add comment to `ProfiledThread::initCurrentThreadSignalSafe()`

ddprof-lib/src/main/cpp/threadLocalData.cpp
zhengyu123: - The call is not performance critical and optimized compiler can inline it as it sees fit - they are in the same compilation unit.
- `initCurrentThread()` is called by the caller that already has signal blocked.

ddprof-lib/src/main/cpp/profiler.cpp
zhengyu123: Yea, I don't see it happen unless OOM :-( I added the check to silence AI. We already have counter `SAMPLES_DROPPED_THREAD_LOCAL` to give a rough idea.

PR #642, opened 2026-07-08
chaos: vthread-context-cascade antagonist + glibc heap-corruption defaults

jbachorik:
**What does this PR do?**:
Adds a new `vthread-context-cascade` chaos antagonist that races tracer-style
context propagate/activate/restore semantics against virtual-thread carrier
churn, targeting the `ContextStorageMode.THREAD` stale-carrier
`DirectByteBuffer` use-after-free. Also defaults `MALLOC_CHECK_`/
`MALLOC_PERTURB_` for glibc-backed test runs (chaos reliability runner and
the shared `ddprof-test` Gradle test-task configuration), turning silent
heap corruption into an immediate, attributable `SIGABRT` instead of a
crash much later in an unrelated allocation.

**Motivation**:
While building the `vthread-context-cascade` reproducer it became clear
that this class of use-after-free bug corrupts glibc heap/tcache metadata
silently and only manifests as a crash far later, in an unrelated
allocation call — making it unlikely to be caught within a short chaos
run or CI test window. Enabling glibc's own heap-corruption checks by
default closes that gap cheaply, without needing a sanitizer build (which
can't run against a live JVM, see `reference_sanitizers_no_jvm`).

**Additional Notes**:
The `MALLOC_CHECK_`/`MALLOC_PERTURB_` defaults are skipped on musl (no
`MALLOC_CHECK_` support) and whenever a sanitizer/allocator already
replaces malloc via `LD_PRELOAD` (ASan/TSan/tcmalloc/jemalloc configs),
mirroring the existing `gmalloc` vs `tcmalloc`/`jemalloc` mutual
exclusivity in `chaos_check.sh`.

**How to test the change?**:
- `./gradlew :ddprof-stresstest:chaosJar` builds cleanly with the new antagonist registered.
- `./gradlew :build-logic:conventions:compileKotlin` compiles cleanly with the `ProfilerTestPlugin` change.
- Ran the antagonist standalone in a JDK21 container (`-Dddprof.debug.context.storage.mode=thread`, tuned virtual-thread scheduler flags, `MALLOC_CHECK_=3`/`MALLOC_PERTURB_` set) for several minutes; no crash reproduced yet in this window — the antagonist is being landed as a reliability-suite addition to keep exercising this race over longer soak runs, not as a guaranteed-repro smoke test.

**For Datadog employees**:

- [x] JIRA: [PROF-15360]

[PROF-15364]: https://datadoghq.atlassian.net/browse/PROF-15364?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[PROF-15360]: https://datadoghq.atlassian.net/browse/PROF-15360?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


rkennke:
Not a review, but one thing that I can't quickly see from looking at the code: does it trigger actual migrations of vthreads across carriers? Because that is the fault scenario that led to the heap corruption.

jbachorik:
@rkennke It is trying to - but it is very probabilistic. It is trying to hold back carrier threads so new ones have to be spawned, then let the old one die so the vthread will have to be remounted but DBB still point to now invalid memory.
TBH, I haven't really got any invalid access, so the test might not be completely successful in doing that :/

rkennke:
### J9 `MALLOC_CHECK_` skip: stale ticket reference + open question

The comment justifying the J9 skip in `ProfilerTestPlugin.kt` contradicts this PR's own narrative, and it's the shipped comment (not just the description) that carries the problem:

```kotlin
// ... skip on J9 (its own allocator usage trips a pre-existing, unrelated heap
// corruption that needs separate investigation — see PROF-15360).
```

Two issues:

1. **Stale ticket reference.** This cites PROF-15360 as an open "needs separate investigation" item, but the PR description says PROF-15360 is *closed* — and closed precisely because it wasn't heap corruption at all; the real root cause was the `isRawPointer()` debug-assert `SIGABRT` that this same PR fixes (PROF-15364). So the comment describes the fixed bug as an unrelated, still-open one.

2. **The stale justification is load-bearing.** It's the stated reason for skipping `MALLOC_CHECK_`/`MALLOC_PERTURB_` on J9 — a real behavioral decision. So this isn't cosmetic: it drives whether J9 test runs get heap-corruption checking at all.

That raises a substantive question: **if the only known J9 problem was PROF-15360 and it's fixed here, is the J9 skip still needed?** Either

- there's a *genuinely separate* J9 heap issue → then point the comment at a real **open** ticket with an accurate description; or
- there isn't → then the skip is suppressing `MALLOC_CHECK_` coverage on exactly the VM where the recent trouble surfaced, and should probably be dropped so J9 gets the same checking as glibc/HotSpot.

Could you reconcile the comment and confirm which case applies before merge?

rkennke:
> @rkennke It is trying to - but it is very probabilistic. It is trying to hold back carrier threads so new ones have to be spawned, then let the old one die so the vthread will have to be remounted but DBB still point to now invalid memory. TBH, I haven't really got any invalid access, so the test might not be completely successful in doing that :/

I see. If you have the time/nerve(/AI tokens) for it, you could perhaps try using a custom FJP where you can control the thread lifecycles. Probabilistic is fine, though, better than nothing.

jbachorik:
@rkennke That's a great idea. Thanks!

jbachorik:
>Two issues:
> 
> 1. **Stale ticket reference.** This cites PROF-15360 as an open "needs separate investigation" item, but the PR description says PROF-15360 is *closed* — and closed precisely because it wasn't heap corruption at all; the real root cause was the `isRawPointer()` debug-assert `SIGABRT` that this same PR fixes (PROF-15364). So the comment describes the fixed bug as an unrelated, still-open one.
> 
> 2. **The stale justification is load-bearing.** It's the stated reason for skipping `MALLOC_CHECK_`/`MALLOC_PERTURB_` on J9 — a real behavioral decision. So this isn't cosmetic: it drives whether J9 test runs get heap-corruption checking at all.
> 
> That raises a substantive question: **if the only known J9 problem was PROF-15360 and it's fixed here, is the J9 skip still needed?** Either
> 
> - there's a *genuinely separate* J9 heap issue → then point the comment at a real **open** ticket with an accurate description; or
> - there isn't → then the skip is suppressing `MALLOC_CHECK_` coverage on exactly the VM where the recent trouble surfaced, and should probably be dropped so J9 gets the same checking as glibc/HotSpot.
> 
> Could you reconcile the comment and confirm which case applies before merge?


Confirmed case 2. Traced the skip's history: the commit that added it (`0b03bba2f`) says its sole justification is "a pre-existing, unrelated glibc heap-corruption bug" - that's exactly PROF-15360, which is closed as invalid in favor of PROF-15364 (the `isRawPointer()` fix in this PR). There's no other ticket or commit citing a genuinely separate J9 issue. Dropped the skip in `1d054ccc3` - J9 now gets the same `MALLOC_CHECK_`/`MALLOC_PERTURB_` coverage as glibc/HotSpot>.

utils/run-chaos-harness.sh
```diff
@@ -0,0 +1,195 @@
...
+  echo "FAIL:no java on PATH (set CHAOS_JDK to have one downloaded)" >&2
+  exit 1
+fi
+echo "Using: $(java -version 2>&1 | head -1)"
+
+# --- ddprof jar --------------------------------------------------------------
+DDPROF_JAR=$(ls "${ROOT}/ddprof-lib/build/libs/ddprof-"*.jar 2>/dev/null | head -1)
+if [ -z "${DDPROF_JAR}" ] || [ ! -f "${DDPROF_JAR}" ]; then
+  echo "FAIL:no local ddprof jar found — build one first: ./gradlew :ddprof-lib:jar" >&2
+  exit 1
```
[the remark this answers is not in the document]
rkennke: Confirmed — this is a real regression, not a false positive. The chaos jobs only `needs: [get-versions]` and never build/fetch a ddprof jar, so with the Maven-snapshot fallback gone this exits at `FAIL:no local ddprof jar found` for the whole matrix. When fixing, also update the now-stale CI comments that still describe the removed fallback: `.gitlab/reliability/pr-child.gitlab-ci.yml:76-77` ("downloads ddprof from Maven snapshots when no local jar exists") and the wrapper note at line 4.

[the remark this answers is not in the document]
jbachorik: Fixed in fb6a1e880 — restored the Maven snapshot fallback (using `CURRENT_VERSION`) for when no local ddprof jar is present.

utils/run-chaos-harness.sh
```diff
@@ -0,0 +1,195 @@
...
+if [ ! -f "${PATCHED_AGENT}" ]; then
+  echo "FAIL:dd-java-agent patching failed" >&2
+  exit 1
+fi
+
+case $CONFIG in
+  profiler)
+    ENABLEMENT="-Ddd.profiling.enabled=true -Ddd.trace.enabled=false"
+    # @Trace is a no-op without the tracer, so trace-context is excluded here.
+    DEFAULT_ANTAGONISTS="thread-churn,alloc-storm,vthread-churn,classloader-churn,vthread-context-cascade,bounded-pool,context-hop,consumer-group,hidden-class-churn,direct-memory,weakref-wave,dump-storm"
```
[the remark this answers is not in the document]
jbachorik: Fixed in fb6a1e880 — dropped `vthread-context-cascade` from the `profiler`-only default antagonist list, since it's driven entirely by `@Trace`-annotated methods that are no-ops without the tracer.

utils/run-chaos-harness.sh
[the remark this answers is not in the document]
jbachorik: I also added guardrails for agents to make this kind of regression less possible or at least very visible.

jbachorik:
Root-caused the reliability crashes reported in `hs_err_1.log`: bisecting the chaos-harness antagonists down to `vthread-context-cascade` alone reproduces the heap corruption every time (native-memory-tracking footer-canary abort in `os::free()`, downstream of a `JvmtiThreadState` free during `VTMS_vthread_end`).

Confirmed the crash requires the ddprof agent (the same antagonist run with no `-javaagent` completes cleanly), then traced it to `OtelContextStorage`'s `ContextStorageMode.THREAD` fallback — the exact use-after-free the class's own javadoc (and `vthread-context-cascade`'s javadoc) already documents: without carrier-scoped storage, a virtual thread's context buffer stays pinned to whichever carrier was mounted at first write; once that carrier's native `ProfiledThread` is freed on thread exit, a later write from a migrated virtual thread lands in freed memory and corrupts the heap. The `[WARN] ddprof: carrier-scoped OTEL context storage is unavailable...` line is present in the crashed run's log, confirming `THREAD` mode was active.

The harness was never passing `--add-exports java.base/jdk.internal.misc=ALL-UNNAMED`, so `CARRIER` mode (the actual fix, already shipped) never activated in reliability testing. Added the export flag in `utils/run-chaos-harness.sh`; re-ran the isolated antagonist for 90s afterward — no WARN, `CARRIER` mode active, completes cleanly (`RC=0`).

Worth calling out: `vthread-context-cascade` did exactly what it was designed for — it caught the memory-safety issue its own doc comments predicted for the `THREAD` fallback. Production is unaffected (dd-trace-java already grants this export at runtime); this was purely a reliability-harness gap.

jbachorik:

jbachorik:

utils/run-chaos-harness.sh
```diff
@@ -0,0 +1,255 @@
...
+esac
+ANTAGONISTS="${ANTAGONISTS_OVERRIDE:-${DEFAULT_ANTAGONISTS}}"
+
+case $ALLOCATOR in
+  gmalloc)
+    # Turn silent heap corruption (e.g. a stale-pointer UAF write into a freed
+    # chunk) into an immediate, attributable SIGABRT instead of a crash much
+    # later in an unrelated allocation. Only meaningful against glibc's own
+    # malloc, not the LD_PRELOAD-replaced allocators below.
+    export MALLOC_CHECK_=3
```
kaahos, line 161: for quite recent glibc (>= 2.34; the CI images use Ubuntu 22.04/glibc 2.35), malloc checking is disabled in the main C library and requires preloading libc_malloc_debug (using `LD_PRELOAD`). from [glibc heap-consistency documentation](https://sourceware.org/glibc/manual/2.34/html_node/Heap-Consistency-Checking.html):
> Like in the case of mcheck, one would need to preload the libc_malloc_debug library to enable MALLOC_CHECK_ functionality. Without this preloaded library, setting MALLOC_CHECK_ will have no effect.

I think we should try to detect and preload libc_malloc_debug for the glibc 2.34+, or fail setup clearly if it cannot be found, so that the intended corruption detection is guaranteed to be active, wdyt?

utils/run-chaos-harness.sh
```diff
@@ -0,0 +1,258 @@
...
+  echo "FAIL:chaos.jar unavailable" >&2
+  exit 1
+fi
+
+# Patch dd-java-agent.jar with the locally built ddprof contents so the
+# agent's (relocated) profiler classes match the version under test.
+PATCHED_AGENT="${WORK_DIR}/dd-java-agent-patched.jar"
+DD_AGENT_JAR="${DD_AGENT_JAR}" DDPROF_JAR="${DDPROF_JAR}" OUTPUT_JAR="${PATCHED_AGENT}" \
+  "${ROOT}/utils/patch-dd-java-agent.sh"
+if [ ! -f "${PATCHED_AGENT}" ]; then
```
kaahos: since the script uses `set +e`, we might check the patcher's exit status directly instead of inferring success from the existence of the jar? `patch-dd-java-agent.sh` creates the JAR before completing validation, so a validation failure can leave an output file which would bypass this check.

```suggestion
if ! DD_AGENT_JAR="${DD_AGENT_JAR}" DDPROF_JAR="${DDPROF_JAR}" OUTPUT_JAR="${PATCHED_AGENT}" \
  "${ROOT}/utils/patch-dd-java-agent.sh"; then
```

ddprof-lib/src/test/cpp/frame_ut.cpp
```diff
@@ -179,3 +174,15 @@ TEST(FrameTypeIsRawPointerTest, FalseWithOnlyBit30SetOnNegative) {
...
     EXPECT_FALSE(FrameType::isRawPointer(negativeWithBit30));
 }
+
+TEST(FrameTypeIsRawPointerTest, FalseForRawAsgctBciWithBit30SetButNoEncodedMarker) {
+    // Regression test: a raw, unencoded ASGCT BCI (bit 20 / ENCODED_MASK not set)
+    // that incidentally has bit 30 set must NOT be reported as a raw pointer.
+    // Such values occur on non-HotSpot VMs (e.g. J9), which never call encode()
+    // and therefore never set ENCODED_MASK; only encode(..., rawPointer=true)
+    // (HotSpot-only, per its own assert) is allowed to set RAW_POINTER_MASK.
+    int rawAsgctBciWithBit30 = 1 << 30;
```
kaahos, line 200: Hotspot is off by default and since `isRawPointer()` requires Hotspot to be on, the test always passes. I'd recommend forcing Hotspot to be activated.

```suggestion
    int rawAsgctBciWithBit30 = 1 << 30;
    VMHotspotGuard hotspot(true);
```

kaahos (commented):
I've started to take a look at it and left some comments!

utils/run-chaos-harness.sh
jbachorik: Fixed in 9ef1d899a: detect glibc >= 2.34 and preload libc_malloc_debug for MALLOC_CHECK_ to take effect, or fail setup clearly if it can't be found.

utils/run-chaos-harness.sh
jbachorik: Fixed in 9ef1d899a: check patch-dd-java-agent.sh's exit status directly instead of inferring success from JAR presence.

ddprof-lib/src/test/cpp/frame_ut.cpp
jbachorik: Fixed in 9ef1d899a: added VMHotspotGuard hotspot(true); to FalseForRawAsgctBciWithBit30SetButNoEncodedMarker (frame_ut.cpp:184), matching the suggested diff.

rkennke (commented):
## Review

Overall this is careful, well-commented work — the antagonist and runner reflect a lot of hard-won operational knowledge (documented job IDs, hysteresis, keep-alive reasoning). My main concerns are scope and blast radius, not correctness of the core antagonist.

### Key issues

**1. Scope creep: the `frame.h` / `frame_ut.cpp` change is unrelated to this PR.**
The `isRawPointer` gating change has nothing to do with the vthread antagonist or malloc defaults. The production change itself is *correct* — `encode()` always sets `ENCODED_MASK` (frame.h:63) and `RAW_POINTER_MASK` is only ever set alongside it, so requiring `ENCODED_MASK` before trusting bit 30 is sound, and the new regression test is good. But bundling a production behavior change into a "chaos tooling" PR hides it from reviewers who'd otherwise scrutinize the packing logic. Recommend splitting it into its own PR.

**2. Two tests were deleted with no explanation** (`frame_ut.cpp`): `FrameTypeEncodeTest.EncodedValuesArePositive` and `FrameTypeDecodeTest.DecodedTypeIsInValidRange`. Both look like they'd still pass and both cover behavior orthogonal to the `isRawPointer` change. Deleting passing tests silently reduces coverage — if they were redundant or wrong, please say so; otherwise keep them.

**3. `MALLOC_CHECK_`/`MALLOC_PERTURB_` blast radius in `ProfilerTestPlugin.kt`.** This applies to *every* `ddprof-test` task, not just the chaos suite.
- **Non-reproducible failures:** `MALLOC_PERTURB_` is `(1..255).random()` (and `$((RANDOM % 255 + 1))` in the shell) and the value is never logged. When a run aborts on corruption you can't reproduce it with the same perturb byte. Consider logging the chosen value (Gradle test output + the runner's stdout).
- **Surfacing latent issues suite-wide** is presumably intended, but worth calling out in the PR body — this can turn previously-green unrelated tests red.
- Minor: the guard is `!testEnv.containsKey("LD_PRELOAD")`. If `LD_PRELOAD` is set in the process environment rather than the task's `testEnv` map, the guard won't see it. Confirm the sanitizer/tcmalloc configs all populate `testEnv`.

### Smaller notes

- **`MemoryGovernor` `System.gc()` race:** two threads can both observe `critical == false`, both set it, and both fire `System.gc()`. The comment promises a "one-shot nudge"; it's approximately-once, not once. Harmless, but the code is slightly weaker than the comment claims.
- **`paceCalls.incrementAndGet()` on every hot-loop iteration** (MemoryGovernor) is a single contended `AtomicLong` shared across all antagonist threads — a cache-line hotspot that perturbs the very allocation timing you're stressing. A per-thread counter would avoid it if it ever matters.
- **`fastCarrierScheduler` is `ThreadPoolExecutor(0, Integer.MAX_VALUE, …, SynchronousQueue)`** — unbounded platform threads, bounded only indirectly by `MAX_STALE_RACERS = 1024`. Worth confirming ~1024 OS carrier threads stay under the container thread/RSS ceilings on aarch64.
- **`start()` hard-fails on a VT-capable JDK without `--add-opens`** (throws, crashing the harness). Intentional and documented, and `run-chaos-harness.sh` passes the flag — but it's a tight coupling: any alternate invocation path that forgets the flag hard-crashes at startup. Acceptable given it's test-only.
- **`run-chaos-harness.sh` `find … -name a -o -name b`** relies on the implicit `-print` attaching to the whole OR expression. It works, but wrapping in `\( … \)` makes intent explicit and robust against a future added clause.
- **`stopGracefully` joins the three driver threads unconditionally.** Safe here because a `start()` that threw is never registered in `Main`'s running list, but a null-guard or comment would prevent a shutdown NPE from masking the real failure.

### Correctness / tests / security
- **Correctness:** the one production change (`frame.h`) is verified correct against the bit layout; semaphore acquire/release pairing in both antagonists correctly releases on the reflective-failure path (no permit leaks).
- **Test coverage:** the new `isRawPointer` regression test is good; the two unexplained deletions are the gap (#2).
- **Conventions:** comment style, copyright headers, and Jira-only-in-PR-body all respected.
- **Security:** test/CI tooling only; the `dtdg.co` download and Maven snapshot fallback are unchanged from the prior script.

### Bottom line
The chaos antagonist and runner rework are solid. Before merge I'd (1) split out the `frame.h` change, (2) restore or justify the two deleted `frame_ut.cpp` tests, and (3) log the `MALLOC_PERTURB_` value so glibc-detected aborts are reproducible. The rest is optional polish.

jbachorik:
Ouch, the `frame.h/cpp` was supposed to go to a different PR. Will fix!

rkennke:
Thanks for the quick turnaround — the review items look resolved:

- Deleted `frame_ut.cpp` tests restored ✅
- `MALLOC_PERTURB_` value now logged (Gradle plugin + runner) so a corruption abort is reproducible ✅
- Polish commit: exactly-once GC nudge (CAS), per-thread pacing counter, bounded fast-carrier pool, null-guarded `stopGracefully` ✅
- P1 Maven-snapshot fallback + profiler-only cascade exclusion fixed ✅
- glibc 2.34+ `libc_malloc_debug` preload, patcher exit-status check, and HotSpot-forcing in the `isRawPointer` regression test all addressed ✅

I'll hold the approval until the `frame.h`/`frame_ut.cpp` split lands (per your note that they belong in a different PR) — that'll just be a quick delta re-review to confirm those two files are gone. The `isRawPointer` change itself is correct, so it's an easy approve wherever it ends up.

jbachorik:
@rkennke The `frame*` changes are gone from this PR

rkennke (approved):
Looks good now, thank you!

PR #644, opened 2026-07-08
Reference chains for surviving live-heap samples via a bounded manual heap walk (PROF-15341)

jbachorik:
**What does this PR do?**:

This PR adds reference chains for live-heap leak candidates (PROF-15341).
`LivenessTracker` tracks class populations across GC generations. A class
with a sustained growth trend becomes a leak candidate. For each candidate,
a background BFS thread walks the heap. It records the referrer chain from a
GC root to a live instance. The profiler writes each chain as a
`datadog.ReferenceChain` JFR event.

Key properties:

- The walk uses JVMTI heap calls only. It reads no raw oop. JVMTI applies the active collector's own barriers. This keeps the walk correct on every collector, including ZGC with concurrent relocation.
- The walk runs in small, budgeted passes. Each pass resumes from a persisted frontier. Each pass is abortable.
- A PID controller keeps each pass under the pause-time target. Two leaky buckets cap total safepoint cost and total CPU cost.
- The subsystem is off by default. Set `referencechains=true` to enable it.

```mermaid
flowchart LR
    A[LivenessTracker leak candidates] --> B[marker-tag representative objects]
    B --> C[BFS pass: roots + static sweep + expand + descend walks]
    C --> D[chain found and pruned]
    D --> E[chain cached and re-stamped into each JFR dump]
```

Main design decisions. The full detail lives in the design documents listed
under "Additional Notes":

- **Bounded manual walk.** Each pass makes one batched `FollowReferences` call. One BFS level is expanded per safepoint operation. No pass walks the whole heap at once.
- **Static-field roots.** Objects held only by static fields are invisible to root enumeration. A chunked, resumable sweep covers them. The sweep runs only when the loaded-class set changed.
- **Candidate-scoped reach.** An open chase gets bounded descend walks before breadth-first work. One prong walks the candidate threads' `ThreadLocalMap` subgraphs. One prong walks root-attached static holders.
- **Canary chase.** Each candidate gets a marker tag on its representative object. The walk records the chain at the first match. A chase that makes no progress backs off exponentially. A chase with no progress for 30 passes is abandoned.
- **Rotation.** A growing `static final` collection is already visited by the walk. The search stays open while a watched class grows. Rotation re-expands the flagged containers and picks up new elements.
- **Termination.** A search ends by completion, wall-clock TTL, frontier cap, or stuck chase. The reason is reported in a `datadog.ReferenceChainAbandoned` event. Restarts need a live leak signal and pain-budget headroom.
- **OOM ramp.** The projected time to OOM drives an exponential ramp of pause target, cadence, and budget. Liveness tracking admission rises to 100% during the ramp.
- **Admission boost.** During a chase, the candidates' allocating threads are admitted to liveness tracking at 100%. The default is 10% subsampling. The boost only adds admissions. It never drops one.

**Motivation**:

PROF-15341: give live-heap samples that survive long enough to look
leak-shaped an actual referrer chain, not just "this object is still alive",
without regressing safepoint-pause behavior on low-pause collectors.

**Additional Notes**:

Design documents shipped with this PR:

- `doc/reference-chains-design.md`: why bounded BFS-from-roots was chosen over the alternatives.
- `doc/reference-chains-collection-summary.md`: the four components. These are detection signal, resumable walk, latency budgets, and rotation.
- `doc/architecture/LiveHeapReferenceChains.md`: the full architecture. It covers the frontier and tag lifecycle, triggering, and termination.
- `doc/architecture/LiveHeapReferenceChains-Implementation.md`: the as-built implementation reference. It covers the pass structure, root discovery, candidate-scoped reach, canary tagging, rotation tiers, and pacing arithmetic.
- `doc/architecture/ReferenceChains-SignalsExplained.md`: a guided tour of the scheduling side. It covers cadence, the leak-signal gate, the pain budget, the PID controller, canary backoff, the OOM ramp, and the abort path.

Other notes:

- The tuning defaults are round, unbenchmarked placeholders. JMH cannot tune them: the cost is STW pause time per pass plus CPU on a native background thread, and the feature only activates after leak-detection hysteresis clears over many GC epochs. The shipped `utils/` tooling measures the real quantities directly: per-pass safepoint durations from JFR and the tracker pass-cost telemetry (see `LiveHeapReferenceChains-Implementation.md` §11). A coarse JMH on/off A/B remains useful as an end-to-end throughput regression guard.

**How to test the change?**:

- `./gradlew :ddprof-lib:gtestDebug`: the native unit suite. It covers the walk engine, frontier table, target selection, rotation, pacing, JFR roundtrip, argument parsing, and the leak-candidate ranking.
- `./gradlew :ddprof-test:testDebug --tests "*referencechains.*"`: the Java suite:
  - `ExternalProcessReferenceChainTest`: separate-process end-to-end. A child JVM runs a leaking app. The test asserts the chain's leaf class. It also asserts the chain re-emits into three fresh JFR dumps. One case appends to an already-discovered `static final` collection and asserts the chain for a later element.
  - `ThreadLocalLeakReferenceChainTest`: separate-process. The leak is reachable only through a worker thread's `ThreadLocalMap`.
  - `LeakTagCorrelationReferenceChainTest`: leak-tag correlation, plus rotation under pod-like tight pass windows.
  - `AggressiveLeakReferenceChainTest`: the search gate opens on heap-wide growth with no single candidate. It stays closed on a flat heap floor.
  - `ReferenceChainTrackingTest`: in-process coverage for the walk engine, target selection, and abandonment reporting.
  - `ReferenceChainJfrParserTest`: JFR-side parsing.
  - `ReferenceChainTestSeamsTest`: C++ seam exposure.
- Repro and chaos tooling: `utils/run-refchains-repro.sh` with `ddprof-stresstest`'s `ReferenceChainLeakDemo`; `utils/sweep-refchains-*.sh` with `utils/refchains-report*.py`; `utils/run-chaos-harness.sh` with `ReferenceChainLeakAntagonist`.

**For Datadog employees**:

- [x] JIRA: [PROF-15341]

[PROF-15341]: https://datadoghq.atlassian.net/browse/PROF-15341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbachorik:
@codex review

jbachorik:
@codex review

jbachorik:
@arp review

PR #647, opened 2026-07-10
fix(profiler): guard line-number-table copy against stale jmethodID

jbachorik:
**What does this PR do?**:
Guards the JVMTI line-number-table copy in `Lookup::fillJavaMethodInfo` (`flightRecorder.cpp`) with `SafeAccess::isReadableRange()` before `memcpy`-ing it, instead of dereferencing the pointer unconditionally. Also bounds `line_number_table_size` before it's trusted to compute a byte range: sizes above `MAX_LINE_NUMBER_TABLE_ENTRIES` (65535, the JVM spec's max `code_length`) or negative are rejected outright, since a corrupted pointer from a stale jmethodID is just as likely to be paired with a corrupted size. Adds a `LINE_NUMBER_TABLE_UNREADABLE` counter for observability when either guard trips.

**Motivation**:

Production crash report: SIGSEGV in `__memcpy_evex_unaligned_erms`, inside `Lookup::resolveMethod` (inlined `fillJavaMethodInfo`), during `FlightRecorder::dump`. Confirmed on stock HotSpot JDK 11.

`fillJavaMethodInfo` calls `GetLineNumberTable()` on a `jmethodID` captured asynchronously at sample time, which can point at an already-unloaded class by the time the dump thread processes it (the TOCTOU race already documented a few lines above this call, re: `GetMethodDeclaringClass`). The sibling `class_name`/`method_name`/`method_sig` strings returned by the preceding JVMTI calls are already probed with `SafeAccess::isReadableRange` for exactly this reason (#537 / `ef13aa29f`), but the line-number-table pointer and its paired size were copied/trusted unguarded. This PR closes that gap.

**Additional Notes**:
- Reproducer (`lineNumberTableCopy_ut.cpp`) isolates the copy step from JVMTI/JNI (impractical to fake a live JVM in a plain gtest) by exercising the identical copy pattern against a real `mmap`'d-then-`munmap`'d page, standing in for "GetLineNumberTable() returned a bad pointer" regardless of the precise reason. Verified independently outside gtest too: the unguarded pattern segfaults (exit 139) and the guarded pattern exits cleanly.
- Also adds `JMethodIDInvalidationStressTest.java`, a JUnit stress test that races class-unload churn (via disposable classloaders + a dedicated GC-pressure thread) against concurrent `dump()`/`resolveMethod()` calls, and asserts via native whitebox counters that a stale jmethodID was actually observed and safely guarded during the run — not just that the JVM didn't crash. The counter-delta check is a JUnit assumption (skip, not fail) since hitting the race within the window is host/JVM-discretionary.
- This branch also carries an unrelate addition to `AGENTS.md` (a short "CI / Automation Script Changes" guardrails section) picked up from other work on this branch; it is not part of the PROF-15385 fix itself but the gap was identified during the reviews and the addition is cheap/small enough not to disturb the main PR.

**How to test the change?**:
- `./gradlew :ddprof-lib:gtestDebug_lineNumberTableCopy_ut` — 3 tests: unguarded pattern crashes as expected (`EXPECT_DEATH`), guarded pattern skips the copy safely when the source is unmapped/oversized, guarded pattern still copies correctly for valid data.
- `./gradlew testDebug -Ptests=JMethodIDInvalidationStressTest` — stress test racing class-unload churn against profiler dump/resolve.

**For Datadog employees**:
- [x] JIRA: [PROF-15385]

[PROF-15385]: https://datadoghq.atlassian.net/browse/PROF-15385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/counters.h
+++ b/ddprof-lib/src/main/cpp/counters.h
@@ -74,7 +74,7 @@
   X(UNWINDING_TIME_JVMTI, "unwinding_ticks_jvmti")                             \
   X(CALLTRACE_STORAGE_DROPPED, "calltrace_storage_dropped_traces")             \
   X(LINE_NUMBER_TABLES, "line_number_tables")                                  \
-  X(LINE_NUMBER_TABLE_UNREADABLE, "line_number_table_unreadable")             \
+  X(LINE_NUMBER_TABLE_UNREADABLE, "line_number_table_unreadable")              \
   X(REMOTE_SYMBOLICATION_FRAMES, "remote_symbolication_frames")                \
   X(REMOTE_SYMBOLICATION_LIBS_WITH_BUILD_ID, "remote_symbolication_libs_with_build_id") \
   X(REMOTE_SYMBOLICATION_BUILD_ID_CACHE_HITS, "remote_symbolication_build_id_cache_hits") \
--- a/ddprof-lib/src/main/cpp/counters.h
+++ b/ddprof-lib/src/main/cpp/counters.h
@@ -74,6 +74,7 @@
   X(UNWINDING_TIME_JVMTI, "unwinding_ticks_jvmti")                             \
   X(CALLTRACE_STORAGE_DROPPED, "calltrace_storage_dropped_traces")             \
   X(LINE_NUMBER_TABLES, "line_number_tables")                                  \
+  X(LINE_NUMBER_TABLE_UNREADABLE, "line_number_table_unreadable")             \
   X(REMOTE_SYMBOLICATION_FRAMES, "remote_symbolication_frames")                \
   X(REMOTE_SYMBOLICATION_LIBS_WITH_BUILD_ID, "remote_symbolication_libs_with_build_id") \
   X(REMOTE_SYMBOLICATION_BUILD_ID_CACHE_HITS, "remote_symbolication_build_id_cache_hits") \
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -47,6 +47,16 @@
 static const char *const SETTING_RING[] = {NULL, "kernel", "user", "any"};
 static const char *const SETTING_CSTACK[] = {NULL, "no", "fp", "dwarf", "lbr"};
 
+// JVM spec SS4.7.3 caps a method's bytecode (code_length) at 65535 bytes (u2),
+// so a well-formed LineNumberTable can never have more entries than that.
+// Used to sanity-bound line_number_table_size before it drives the byte-count
+// passed to SafeAccess::isReadableRange()/memcpy(): if GetLineNumberTable()
+// returns a corrupted pointer for a stale jmethodID (see the TOCTOU race
+// documented in fillJavaMethodInfo below), the paired out-param size is just
+// as likely to be corrupted, and an implausible size should be rejected
+// before it is trusted to compute a byte range.
+static const jint MAX_LINE_NUMBER_TABLE_ENTRIES = 65535;
+
 // Compute a non-negative event duration from TSC timestamps.  Unsigned u64
 // subtraction wraps to a near-2^64 value when end < start, which can happen if
 // the thread migrates cores between the two TSC reads and the per-core counters
@@ -360,7 +370,8 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
       // the JVMTI-allocated memory immediately. This keeps _ptr valid even
       // after the underlying class is unloaded.
       void *owned_table = nullptr;
-      if (line_number_table_size > 0) {
+      if (line_number_table_size > 0 &&
+          line_number_table_size <= MAX_LINE_NUMBER_TABLE_ENTRIES) {
         size_t bytes = (size_t)line_number_table_size * sizeof(jvmtiLineNumberEntry);
         // GetLineNumberTable() is called on the same possibly-stale jmethodID
         // that GetMethodDeclaringClass/GetClassSignature/GetMethodName above
@@ -383,6 +394,14 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
           line_number_table = nullptr; // make sure the invalid address is not used for jvmti->Deallocate
           Counters::increment(LINE_NUMBER_TABLE_UNREADABLE);
         }
+      } else if (line_number_table_size > MAX_LINE_NUMBER_TABLE_ENTRIES) {
+        // A corrupted size out-param alongside a corrupted pointer is exactly
+        // as plausible as the corrupted-pointer case above (both come from
+        // the same GetLineNumberTable() call on the same stale jmethodID);
+        // an implausible entry count means the pointer can't be trusted for
+        // Deallocate() either, so treat it the same as the unreadable case.
+        line_number_table = nullptr;
+        Counters::increment(LINE_NUMBER_TABLE_UNREADABLE);
       }
       jvmtiError dealloc_err = jvmti->Deallocate((unsigned char *)line_number_table);
       if (dealloc_err != JVMTI_ERROR_NONE) {
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -394,12 +394,14 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
           line_number_table = nullptr; // make sure the invalid address is not used for jvmti->Deallocate
           Counters::increment(LINE_NUMBER_TABLE_UNREADABLE);
         }
-      } else if (line_number_table_size > MAX_LINE_NUMBER_TABLE_ENTRIES) {
+      } else if (line_number_table_size != 0) {
         // A corrupted size out-param alongside a corrupted pointer is exactly
         // as plausible as the corrupted-pointer case above (both come from
         // the same GetLineNumberTable() call on the same stale jmethodID);
-        // an implausible entry count means the pointer can't be trusted for
-        // Deallocate() either, so treat it the same as the unreadable case.
+        // an implausible entry count -- including a negative one, since this
+        // is a signed jint and a corrupted value can fall on either side of
+        // zero -- means the pointer can't be trusted for Deallocate() either,
+        // so treat it the same as the unreadable case.
         line_number_table = nullptr;
         Counters::increment(LINE_NUMBER_TABLE_UNREADABLE);
       }
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -411,9 +411,9 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
         line_number_table = nullptr;
         Counters::increment(LINE_NUMBER_TABLE_UNREADABLE);
       }
-      jvmtiError dealloc_err = jvmti->Deallocate((unsigned char *)line_number_table);
-      if (dealloc_err != JVMTI_ERROR_NONE) {
-        TEST_LOG("Unexpected error while deallocating linenumber table: %d", dealloc_err);
+      if (line_number_table != nullptr) {
+        jvmtiError dealloc_err = jvmti->Deallocate((unsigned char *)line_number_table);
+        assert(dealloc_err == JVMTI_ERROR_NONE && "Unexpected error while deallocating linenumber table");
       }
       if (owned_table != nullptr) {
         mi->_line_number_table = std::make_shared<SharedLineNumberTable>(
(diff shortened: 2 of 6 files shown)


jbachorik:
@codex review

ddprof-test/src/test/java/com/datadoghq/profiler/memleak/JMethodIDInvalidationStressTest.java
```diff
@@ -0,0 +1,306 @@
...
+              + "counter-delta assertion below would be meaningless without this, since it "
+              + "can't tell 'no stale jmethodID was hit' apart from 'churn never ran at all' "
+              + "(e.g. a regression in generateChurnClassBytecode or IsolatedClassLoader).");
+
+      long skippedDelta = after.getOrDefault("jmethodid_skipped_count", 0L)
+          - before.getOrDefault("jmethodid_skipped_count", 0L);
+      long unreadableLineTableDelta = after.getOrDefault("line_number_table_unreadable", 0L)
+          - before.getOrDefault("line_number_table_unreadable", 0L);
+
+      assertTrue(skippedDelta > 0 || unreadableLineTableDelta > 0,
```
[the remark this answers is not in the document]
jbachorik: Fixed in ee4081b1a — switched the counter-delta check to `Assumptions.assumeTrue`, so a run where the churn window doesn't happen to race unload against `resolveMethod`/`fillJavaMethodInfo` is reported as **skipped**, not **failed**. The deterministic checks (crash survival, non-empty dump, churn actually ran) stay as hard assertions.

ddprof-lib/src/test/cpp/lineNumberTableCopy_ut.cpp
```diff
@@ -0,0 +1,189 @@
...
+    if (orig_segvHandler != nullptr) {
+      orig_segvHandler(signo, siginfo, context);
+    }
+  }
+}
+
+class LineNumberTableCopyTest : public ::testing::Test {
+protected:
+  void SetUp() override {
+    orig_segvHandler = OS::replaceSigsegvHandler(lineNumberTableSegvHandler);
```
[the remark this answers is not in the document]
jbachorik: Fixed in [60f1f83](https://github.com/DataDog/java-profiler/pull/647/commits/60f1f839facc0a145054bc31b293d9adce478a05)

jbachorik:
Added commit 95a349859: gates `FrameType::isRawPointer()` on `VM::isHotspot()`.

Unrelated root cause to the line-number-table fix above, found while investigating a J9 debug-lane CI abort in `GetLineNumberTableLeakTest`: the raw-pointer bci bit (bit 30) only means "jmethodID slot replaced by a VMMethod*" on HotSpot. On OpenJ9, an unencoded, VM-supplied bci can coincidentally have bit 30 set, which was misread as our raw-pointer encoding and routed into `JVMSupport::resolve()`, asserting/aborting on any non-HotSpot VM.

Fixing in the PR because the failure was virtually blocking getting the clean CI run.

Reproduced on the exact CI toolchain (JDK21 build JDK, Semeru 8.0.462/OpenJ9 0.53.0 test JDK): unpatched code crashed 1/2 iterations of `GetLineNumberTableLeakTest`; with the fix, 15/15 passed cleanly.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -360,14 +370,38 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
...
+            memcpy(owned_table, line_number_table, bytes);
+          } else {
+            TEST_LOG("Failed to allocate %zu bytes for line number table copy", bytes);
+          }
         } else {
-          TEST_LOG("Failed to allocate %zu bytes for line number table copy", bytes);
+          line_number_table = nullptr; // make sure the invalid address is not used for jvmti->Deallocate
+          Counters::increment(LINE_NUMBER_TABLE_UNREADABLE);
         }
+      } else if (line_number_table_size > MAX_LINE_NUMBER_TABLE_ENTRIES) {
```
rkennke: `line_number_table_size` is a signed `jint`, so a corrupted negative out-param value falls through *both* branches: it fails `> 0` above and it isn't `> MAX_LINE_NUMBER_TABLE_ENTRIES` here, so `line_number_table` stays non-null and gets handed to `Deallocate()` a few lines down — the same untrusted-pointer deref this `> MAX` branch was added to prevent. If the size is corrupt enough to go negative, the paired pointer is equally untrustworthy. Suggest normalizing so any non-zero, out-of-range size hits the reject path, e.g. `else if (line_number_table_size != 0)`. Low severity (telemetry showed the positive-size path), but cheap to close while we're here.

ddprof-lib/src/main/cpp/counters.h
```diff
@@ -74,6 +74,7 @@
   X(UNWINDING_TIME_JVMTI, "unwinding_ticks_jvmti")                             \
   X(CALLTRACE_STORAGE_DROPPED, "calltrace_storage_dropped_traces")             \
   X(LINE_NUMBER_TABLES, "line_number_tables")                                  \
+  X(LINE_NUMBER_TABLE_UNREADABLE, "line_number_table_unreadable")             \
```
rkennke: Nit: the trailing `\` here is a couple columns short of the neighbouring `X(...)` entries — align it to keep the macro block tidy.

rkennke (commented):
Took a thorough pass — the core fix is correct, minimal, and consistent with the #537 string-probe hardening; the death-test's volatile-sink-vs-DCE handling and the dual SIGSEGV/SIGBUS registration are nice touches. Two small things inline, plus one question:

The `isRawPointer` HotSpot gating looks like a genuinely separate defect from the line-number-table memcpy crash — different symptom, different VM (OpenJ9 vs stock HotSpot), different file, and the title/JIRA (PROF-15385) only mention the memcpy. Was there a reason to fold it in here rather than split it into its own PR? Both trace back to stale-jmethodID handling so I'm fine either way — just want to make sure the J9 fix gets its own visibility/traceability.

jbachorik:
@rkennke ~The reason for folding in the unrelated fix was that I was not able to get clean CI pass for this change. And jumping one more hoop to have the rather straight-forward fix felt a bit heavy. I might have gotten impatient but I was chasing weird CI failures in a kind of circular mode the whole day :(
If you feel strongly about this I can extract just the raw-pointer fix to a a PR and then wait with this PR until the previous PR gets merged.~

Actually, extracted to a separate PR (https://github.com/DataDog/java-profiler/pull/658) for posterity

ddprof-lib/src/main/cpp/flightRecorder.cpp
jbachorik: Fixed in a37e7ecfe.

ddprof-lib/src/main/cpp/counters.h
jbachorik: Fixed in a37e7ecfe.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -360,14 +370,40 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
...
+        // much as to those calls, and crash telemetry already showed those
+        // sibling calls returning JVMTI_ERROR_NONE with unmapped string
+        // pointers despite the spec saying the returned array should be a
+        // fresh, caller-owned allocation. Nothing about this call guarantees
+        // it is exempt from the same failure mode, so probe before copying
+        // rather than assume the pointer is safe to dereference.
+        if (SafeAccess::isReadableRange(line_number_table, bytes)) {
+          owned_table = malloc(bytes);
+          if (owned_table != nullptr) {
+            memcpy(owned_table, line_number_table, bytes);
```
zhengyu123: Unfortunately, it is still a race - we may need to add a `safeCopy()` to `safeAccess`

zhengyu123 (changes requested):

ddprof-lib/src/main/cpp/flightRecorder.cpp
jbachorik: Why? A valid jmethodId should return a copy of line number table, hence subsequent invalidation should not have any effect on the obtained copy. If an original invalid jmethodid returns garbage, we try to validate, if we are lucky the range is not readable and we skip. If the range is readable, we treat it as a line number, which might give us garbage but should not crash (the memory is readable, right?)

IMO, we should try and get a reproducer for this and file JVMTI bug because the call should return JVMTI error when called on invalid jmethodid.

Is there another possibility or a race I don't see here?

zhengyu123: Actually, I am not convinced that we need to copy - it does not appear returned line number table has any lifecycle associated with `jmethodID`, because spec says you do need to call `JVMTI->Deallocate()` to release the table.

If it is not, there is a gap between `SafeAccess::isReadableRange()` check and `memcpy()` - that is the race. 

If line number table can indeed be corrupted, that how to guard `JVMTI->Deallocate()` calls? or risk leaking memory.

jbachorik: You're right, and I'd overstated the mechanism initially.

Per the JVMTI spec, a *validly* returned `GetLineNumberTable()` buffer is a fresh, caller-owned allocation fully decoupled from the `jmethodID`'s lifetime, so a spec-compliant JVM can't invalidate it later out from under us. The actual race is narrower: the pointer is already corrupted/stale (the same class-unload TOCTOU documented above for `GetMethodDeclaringClass`/`GetClassSignature`/`GetMethodName`), it happens to alias some other live, mapped memory at `isReadableRange()` check-time, and that aliased mapping changes before the `memcpy` runs. Still a real race, just via a different mechanism than a legitimately-valid buffer going stale later.

`SafeAccess::safeCopy()` already exists in this codebase (`safeAccess.h`/`.cpp`, already used a few lines below for `Symbol` byte copies) - no new primitive needed. It closes the gap by fault-protecting each word read as it happens instead of trusting a point-in-time `isReadableRange()` probe before an unguarded `memcpy`. Swapped this `isReadableRange`+`memcpy` for `malloc`+`safeCopy()`, updated `lineNumberTableCopy_ut.cpp` to cover the new copy path, and verified locally (all 3 relevant gtest cases pass).

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -360,14 +370,46 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
...
+        // as plausible as the corrupted-pointer case above (both come from
+        // the same GetLineNumberTable() call on the same stale jmethodID);
+        // an implausible entry count -- including a negative one, since this
+        // is a signed jint and a corrupted value can fall on either side of
+        // zero -- means the pointer can't be trusted for Deallocate() either,
+        // so treat it the same as the unreadable case.
+        line_number_table = nullptr;
+        Counters::increment(LINE_NUMBER_TABLE_UNREADABLE);
       }
       jvmtiError dealloc_err = jvmti->Deallocate((unsigned char *)line_number_table);
```
zhengyu123: If the table is invalid, will this `jvmti->Deallocate()` call safe?

jbachorik: It would have been set to `nullptr` at https://github.com/DataDog/java-profiler/pull/647/changes#diff-9a6a575bc5c4e9a0980683bb7bb698b45a23bdb480b6a250ae1a2a8eb9374b01R397

`Deallocate` should be ok with `nullptr` but I can add a safety check, just in case.

rkennke (approved):
Looks good to me, thank you!

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -360,18 +370,50 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
...
+        // zero -- means the pointer can't be trusted for Deallocate() either,
+        // so treat it the same as the unreadable case.
+        line_number_table = nullptr;
+        Counters::increment(LINE_NUMBER_TABLE_UNREADABLE);
       }
-      jvmtiError dealloc_err = jvmti->Deallocate((unsigned char *)line_number_table);
-      if (dealloc_err != JVMTI_ERROR_NONE) {
-        TEST_LOG("Unexpected error while deallocating linenumber table: %d", dealloc_err);
+      if (line_number_table != nullptr) {
+        jvmtiError dealloc_err = jvmti->Deallocate((unsigned char *)line_number_table);
```
zhengyu123, line 415: I would suggest that bypassing `jvmti->Deallocate()` if `safeCopy()` fails, because the call is likely to crash if `line_number_table` is inaccessible - leak probably is *not* worst case :-)

jbachorik: If `safecopy` has failed, the `line_number_table` is `nullptr`. We won't get here.

zhengyu123: Ah, okay.

zhengyu123 (approved):

jbachorik:
Thanks for the reviews!

PR #648, opened 2026-07-10
feat(profiler): native malloc/socket stack capture with profiler-hook frame skipping

jbachorik:
**What does this PR do?**:

Adds native stack capture for malloc (`BCI_NATIVE_MALLOC`) and native-socket (`BCI_NATIVE_SOCKET`) samples, stripping profiler-internal hook frames so only real caller frames remain — across all cstack modes (fp/dwarf/vm/vmx). Also renames the JFR event to `datadog.NativeMemoryAllocation`.

- Surfaces silent hook-symbol mark-resolution failures via new diagnostic counters and a warning log instead of silently no-opping (`markAsyncProfilerHook` now returns `bool`; `NativeSocketSampler::start()` checks it).
- Adds a counter for when the `MARK_JAVA_PROFILER` hook boundary is never found in a callchain, so a fully-discarded/empty native stack is observable rather than silent.
- Fixes `convertNativeTrace`'s `MARK_JAVA_PROFILER` reset to gate on the immutable `skip_hook_prefix` parameter instead of the mutable `skipping` flag, so a second boundary-marked frame later in the same callchain is still treated as a reset point (matches `walkVM`'s per-frame dispatch).
- Restores the `MARK_COMPILER_ENTRY` dispatch branch in `walkVM` that had been dropped without updating the `comp_task` flag/docs that still reference it.
- Drops the now-redundant `is_marked` field on `NativeFrameResolution` in favor of a computed `mark != 0` accessor, and updates the stale doc comments describing per-mark dispatch semantics.
- Introduces a single `isHookPrefixedSample()` helper (BCI and EventType overloads) used at all 4 call sites that previously hardcoded the "is this a malloc/socket sample" predicate independently.
- Adds stack-trace assertions to `NativeSocketStackTraceTest` proving a real caller frame (not a profiler-internal hook frame) is present after unwinding.
- Renames the `profiler.Malloc` JFR event to `datadog.NativeMemoryAllocation` and moves it into the Datadog/Profiling JFR category.

**Motivation**:

Native malloc/socket samples currently include profiler-internal hook frames (and, in some cstack modes, drop the real caller entirely), which pollutes symbolication and makes the resulting stacks hard to read.

**Additional Notes**:

The restored `MARK_COMPILER_ENTRY` dispatch is currently unreachable: `_features.comp_task` is never set to `1` anywhere in the codebase (no CLI flag wires it), so this fix restores doc/dead-code consistency but doesn't change observable behavior. Revisiting that separately.

**How to test the change?**:

`./gradlew assembleDebugJar` builds clean; `./gradlew compileTestJava` compiles clean. The new/updated assertions in `NativeSocketStackTraceTest` are parameterized across `vm`/`vmx`/`dwarf`/`fp` cstack modes and cover the fixed code paths.

**For Datadog employees**:
- [x] JIRA: [PROF-15461](https://datadoghq.atlassian.net/browse/PROF-15461)

[PROF-15461]: https://datadoghq.atlassian.net/browse/PROF-15461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbachorik:
@codex review

rkennke:
Before we start reviewing, I think the PR title/description need to be brought in line with what actually lands on `main` when this merges. Three asks:

1. **Re-scope the description to the whole effort, not just the last commit.** The current text reads as a follow-up that "closes the gaps before the branch merges," but nothing here is on `main` yet — the entire feature lands at once: the `datadog.NativeMemoryAllocation` JFR event rename, native frame capture for `BCI_NATIVE_MALLOC`/`BCI_NATIVE_SOCKET` under default cstack, the hook-prefix-skip logic (FP/DWARF + `walkVM`), *and* the review fixups. The description should let a reviewer understand all of that, not just the fixup commit.

2. **Reconsider the `fix(...)` prefix.** Net effect on `main` is *introducing* native malloc/socket stack capture, so `feat(profiler):` is more accurate for the changelog. The "fix" framing is only true relative to intermediate commits on the branch.

3. **Update stale terminology.** The tip commit renames `MARK_ASYNC_PROFILER` → `MARK_JAVA_PROFILER` across the code, but the description still refers to `MARK_ASYNC_PROFILER` throughout — it's now stale against the branch's own final commit.

The bullet-by-bullet breakdown itself is great; this is just a re-scope + rename, not a rewrite.

jbachorik:
Sorry, wanted to click on 'Update branch' and did 'Ready for review' before it was actually ready :/

jbachorik:
@rkennke The PR should be ready now

rkennke:
Thanks for the JIRA link and the event-rename bullet. Two of the three asks are still open though, and both are quick — here's concrete text so it's copy-paste:

**Ask #2 — title prefix.** Change:
> `fix(profiler): complete malloc/native-socket hook-prefix-skip stackwalking`

to something like:
> `feat(profiler): native malloc/socket stack capture with profiler-hook frame skipping`

(net effect on `main` is introducing the feature, not fixing a shipped regression).

**Ask #3 — stale marker name.** The tip commit renames `MARK_ASYNC_PROFILER` → `MARK_JAVA_PROFILER` in the code, but the description still says `MARK_ASYNC_PROFILER` in three bullets (the counter bullet, the `convertNativeTrace` reset bullet). A find-replace of `MARK_ASYNC_PROFILER` → `MARK_JAVA_PROFILER` across the description fixes it.

**Ask #1 — framing (partially done).** The new event-rename bullet helps, but the opening still reads as a follow-up. Suggest reframing the intro from:
> "Fixes up gaps in the malloc/native-socket hook-prefix-skip stackwalking logic introduced earlier on this branch ... which was incomplete"

to something that leads with what the PR delivers, e.g.:
> "Adds native stack capture for malloc (`BCI_NATIVE_MALLOC`) and native-socket (`BCI_NATIVE_SOCKET`) samples, stripping profiler-internal hook frames so only real caller frames remain — across all cstack modes (fp/dwarf/vm/vmx). Also renames the JFR event to `datadog.NativeMemoryAllocation`."

...then keep the existing bullet list as the detailed breakdown. Same content, just scoped to what lands on `main` rather than to the intermediate commits.

Happy to start the review as soon as #2 and #3 are in.

jbachorik:
@rkennke 🤞 all PR desc changes are in

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -681,18 +683,15 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
...
-                if (mark == MARK_ASYNC_PROFILER && event_type == MALLOC_SAMPLE) {
-                    // Skip all internal frames above malloc_hook functions, leave the hook itself
+            if (resolution.is_marked()) {
+                if (resolution.mark == MARK_JAVA_PROFILER &&
+                    isHookPrefixedSample(event_type)) {
+                    // Discard frames captured above the malloc/socket hook boundary,
+                    // excluding the hook's own frame, and resume from the real
+                    // caller above it — mirrors the FP/DWARF skip-prefix logic in
+                    // Profiler::convertNativeTrace.
                     depth = 0;
```
rkennke, line 700: **Observability asymmetry (worth fixing).** The FP/DWARF path (`convertNativeTrace`) increments `NATIVE_TRACE_HOOK_PREFIX_NOT_FOUND` when the boundary is never reached, but `walkVM` has no equivalent. A malloc/socket sample under vm/vmx that never hits `MARK_JAVA_PROFILER` silently returns whatever frames it collected. Since the whole PR is about making these silent failures observable, consider mirroring the counter here.

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -481,6 +509,11 @@ int Profiler::convertNativeTrace(int native_frames, const void **callchain,
     }
   }
 
+  if (skipping) {
+    // The hook-boundary (MARK_JAVA_PROFILER) frame was never found in the
+    // callchain; every frame was discarded and the sample has no native stack.
+    Counters::increment(NATIVE_TRACE_HOOK_PREFIX_NOT_FOUND);
```
rkennke, line 525: **Edge that still escapes this counter.** If `skipping` is still true but a *non-*`MARK_JAVA_PROFILER` mark is hit first, the function `return depth`s (0) earlier without reaching this increment — so one "empty hook-prefixed stack" path stays unobservable. Realistically the hook boundary is the first mark, so this is minor, but if the counter is meant to mean "hook-prefixed sample produced no stack," it's incomplete.

ddprof-lib/src/main/cpp/nativeSocketSampler.cpp
```diff
@@ -27,6 +30,30 @@
...
 
+// Marks the hook wrapper's own symbol as MARK_JAVA_PROFILER so native call-stack
+// unwinding (Profiler::convertNativeTrace) can recognize the boundary between
+// profiler-internal frames and the real caller, mirroring MallocHooker::initialize().
+// Resolved by address (not by symbol-name predicate) because these are mangled
+// C++ static member functions, unlike malloc_hook's extern "C" free functions.
+// Returns false if the symbol could not be resolved/marked, in which case the
+// hook boundary is never recognized and every socket sample's native stack
+// comes back empty (see NATIVE_TRACE_HOOK_PREFIX_NOT_FOUND).
+static bool markAsyncProfilerHook(void* fn_addr) {
```
rkennke: **Stale name post-rename.** This marks `MARK_JAVA_PROFILER`, but the function is still called `markAsyncProfilerHook`. Suggest `markJavaProfilerHook` for consistency with the rename that landed on this branch.

ddprof-lib/src/main/cpp/codeCache.h
```diff
@@ -55,7 +55,7 @@ enum Mark {
     MARK_VM_RUNTIME = 1,
     MARK_INTERPRETER = 2,
     MARK_COMPILER_ENTRY = 3,
-    MARK_ASYNC_PROFILER = 4, // async-profiler internals such as native hooks.
+    MARK_JAVA_PROFILER = 4, // async-profiler internals such as native hooks.
```
rkennke: Comment still says `async-profiler internals` after the rename to `MARK_JAVA_PROFILER`. Minor, but worth updating so the name and comment agree.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -681,18 +683,15 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
...
-                } else if (mark == MARK_COMPILER_ENTRY && features.comp_task && vm_thread != NULL) {
+                } else if (resolution.mark == MARK_COMPILER_ENTRY && features.comp_task && vm_thread != NULL) {
```
rkennke, line 701: **Dead-code note for the future.** As the PR acknowledges, `features.comp_task` is never set, so this branch is unreachable. Just flagging that the new `goto dwarf_unwind` below also drops the native compiler-entry frame that the old fall-through would have `fillFrame`'d after the pseudo JIT frame — whoever re-enables `comp_task` should know the native frame is no longer emitted.

ddprof-lib/src/main/cpp/profiler.h
```diff
@@ -390,20 +404,25 @@ class alignas(alignof(SpinLock)) Profiler {
...
-      const char* method_name;            // Resolved method name 
+      const char* method_name;            // Resolved method name
     };
     int bci;                            // BCI_NATIVE_FRAME_REMOTE or BCI_NATIVE_FRAME
-    bool is_marked;                     // true if this is a marked C++ interpreter frame (stop processing)
-    NativeFrameResolution(const char* name, int bci_type, bool marked)
-      : method_name(name), bci(bci_type), is_marked(marked) {}
-    NativeFrameResolution(unsigned long packed, int bci_type, bool marked)
-      : packed_remote_frame(packed), bci(bci_type), is_marked(marked) {}
+    char mark;                         // the Mark value for this frame, 0 if unmarked
```
rkennke: Nit: `char mark` holds `Mark` enum values (0..5). Fine for the equality/nonzero comparisons here, but consider noting the value domain (or using the enum type) for the next reader.

ddprof-lib/src/main/cpp/nativeSocketSampler.cpp
```diff
@@ -386,6 +413,18 @@ Error NativeSocketSampler::start(Arguments &args) {
     TEST_LOG("NativeSocketSampler::start interval_ticks=%ld tsc_freq=%llu",
              init_interval, (unsigned long long)TSC::frequency());
 #endif
+    bool hooks_marked = markAsyncProfilerHook((void*)&NativeSocketSampler::send_hook);
+    hooks_marked &= markAsyncProfilerHook((void*)&NativeSocketSampler::recv_hook);
+    hooks_marked &= markAsyncProfilerHook((void*)&NativeSocketSampler::write_hook);
+    hooks_marked &= markAsyncProfilerHook((void*)&NativeSocketSampler::read_hook);
+    if (!hooks_marked) {
```
rkennke, line 420: Coverage gap (acceptable): nothing exercises this failure branch or the new `NATIVE_HOOK_MARK_RESOLVE_FAILED` / `NATIVE_TRACE_HOOK_PREFIX_NOT_FOUND` counters. Genuinely hard to unit-test, so just noting it rather than requesting a test.

rkennke (commented):
Reviewed the net diff against `main`. The logic is sound — I traced both unwinding paths and the central fix is real: on `main`, `walkVM`'s `if (resolution.is_marked) break;` ran *before* the mark-dispatch chain, making the whole `MARK_*` dispatch dead code (every marked frame terminated the scan). The new per-mark dispatch with `else { break; }` fixes that while keeping VM_RUNTIME/INTERPRETER as terminators. `convertNativeTrace` gating the reset on the immutable `skip_hook_prefix` is correct and now matches `walkVM`. The socket test is a genuine positive+negative assertion (real caller present, no `*_hook` leak), parameterized across cstack modes.

One intentional behavior change worth flagging in the description: the `walkVM` MARK_JAVA_PROFILER path now does `depth = 0; goto dwarf_unwind;`, which skips the `fillFrame` that previously emitted the hook frame — so the hook frame is now *dropped* for malloc under vm/vmx too (not just socket). Deliberate and consistent, just wire-visible for existing malloc consumers.

Inline notes below — all minor; #1 and #3 are the ones I'd actually action.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
jbachorik: Fixed in ce3a04774.

ddprof-lib/src/main/cpp/profiler.cpp
jbachorik: Fixed in ce3a04774.

ddprof-lib/src/main/cpp/nativeSocketSampler.cpp
jbachorik: Fixed in ce3a04774.

ddprof-lib/src/main/cpp/codeCache.h
jbachorik: Fixed in ce3a04774.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
jbachorik: Confirmed, no code change here — this is a note for future maintainers, not an actionable defect (features.comp_task is currently never set, so the MARK_COMPILER_ENTRY branch is dead code). Before this PR, a marked frame fell through to a shared fillFrame(frames[depth++], frame_bci, method_name) call after the mark-dispatch if/else, so the compiler-entry branch got both the pseudo JIT frame *and* the native compiler-entry frame itself. The new code replaces that shared fallthrough with `goto dwarf_unwind` at the end of the mark-dispatch chain, so if comp_task is re-enabled, only the pseudo JIT frame will be emitted going forward. Leaving as documented for future readers.

ddprof-lib/src/main/cpp/profiler.h
jbachorik: Fixed in 8e3bb2570 — documented the mark field's value domain (0, or one of the Mark enum constants in codeCache.h) and noted why it's kept as char rather than the enum type (it's packed into RemoteFramePacker's 3-bit field).

ddprof-lib/src/main/cpp/jfrMetadata.cpp
```diff
@@ -303,8 +303,8 @@ void JfrMetadata::initialize(
               << field("name", T_STRING, "Name")
               << field("count", T_LONG, "Count"))
 
-          << (type("profiler.Malloc", T_MALLOC, "malloc")
-                  << category("Java Virtual Machine", "Native Memory")
+          << (type("datadog.NativeMemoryAllocation", T_MALLOC, "Native Memory Allocation")
```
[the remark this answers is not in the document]
rkennke: Checked the compatibility risk directly. `profiler.Malloc` did ship — it's been in releases v_1.42.0 through the current v_1.47.0 (introduced by #398). So this is a real rename of a released public JFR type name with no alias.

However, I verified the primary consumer doesn't depend on it: I cloned `DataDog/profiling-backend` and grepped its production default branch (`prod`, @35d179d, 2026-07-23). There is **no reference to `profiler.Malloc`** anywhere — not the literal name, not a `profiler.*` prefix/wildcard match, not a `startsWith("profiler")` consumer. The new name `NativeMemoryAllocation` is also absent. The only `Malloc` hit is an unrelated comment in `perfetto_config.proto`.

Combined with the feature being opt-in (`nativemem=`), the practical blast radius in the backend is effectively nil, so the hard rename looks safe for the consumer that matters here.

Caveat on scope: I only checked profiling-backend@prod. Other consumers (dd-trace-java, the agent, or dashboards/monitors defined outside code) weren't checked — if any of those query `profiler.Malloc` by name, they'd still break. If there's any chance of that, a compatibility alias would be cheap insurance; otherwise this looks fine to proceed.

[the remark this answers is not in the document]
jbachorik: This was never enabled. It required passing in the profiler arg and the integration from dd-trace-java was simply not there.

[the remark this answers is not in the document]
rkennke: Correcting my earlier comment: I suggested a "compatibility alias would be cheap insurance" — that's wrong, because **JFR has effectively no alias mechanism**. An event's name lives only in the metadata as a 1:1 mapping to a numeric type id, and event records carry just the numeric id (`buf->putVar64(T_MALLOC)`); consumers resolve name→id from each chunk's metadata and filter by name. There's no "also-known-as" field. The only way to expose two names is to declare a second type and **emit every sample twice** — which doubles malloc event volume (each with its own stacktrace ref) and risks double-counting in any consumer that reads both. So an alias is not cheap; the realistic choices are "keep the old name" or "hard rename." (Note: the type *label* — the 3rd arg — is human-facing only and can change freely; only the name string is the breaking part.)

Given that, and that I found **no consumer referencing `profiler.Malloc`** — verified in profiling-backend@prod (full clone + grep) and dd-trace-java (code search, controls confirm the repo is indexed) — I don't think an alias is worth building. **The hard rename looks fine to proceed.**

Same scope caveat as before: dashboards/monitors defined outside these repos weren't checked, but with the two code consumers clear, the practical risk is negligible.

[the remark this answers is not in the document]
rkennke: Agreed — and that's a stronger argument than my consumer grep. "Never enabled" means no recording ever actually contained a `profiler.Malloc` event (the `nativemem` arg was never passed, and the dd-trace-java integration to pass it wasn't there), which is also exactly why the dd-trace-java search came back empty. So there's nothing to break regardless of who queries the name. Resolved from my side — no alias needed, rename is fine.

rkennke (approved):
Ok, looks good to me. Thanks!

PR #660, opened 2026-07-14
Optimize SafeAccess:safeCopy() implementation in assembly

zhengyu123:
**What does this PR do?**:
Reimplement `SafeAccess::safeCopy()` in assembly.

**Motivation**:
Current implementation of SafeAccess::safeCopy() is extremely expensive:

- Double reads to detect fault

- No reads can inline

Optimized the implementation in assembly, similar to SafeAccess::safefetch()

- No false detection. The old code inferred faults from two sentinel values matching — a heuristic. The assembly version catches the actual hardware fault via the signal handler, so it's exact.

- No over-read. The byte-granular copy only ever touches [src, src+len). The old code did 4-byte aligned loads with front-fixup and up-to-3-byte over-reads, requiring careful page-boundary reasoning (all now unnecessary).

- Sanitizer-clean. Assembly loads aren't ASan-instrumented, so faulting reads on unmapped/PROT_NONE pages don't trip the sanitizer.

**Additional Notes**:

**How to test the change?**:
Passed existing tests for old implementation

**For Datadog employees**:
- [X] JIRA: [PROF-15402](https://datadoghq.atlassian.net/browse/PROF-15402)

[PROF-15402]: https://datadoghq.atlassian.net/browse/PROF-15402?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbachorik:
🗿 🤖 🔴

**File:** `ddprof-lib/src/main/cpp/safeAccess.cpp` (line 263)

**[Sphinx Review — MEDIUM]** The range-based fault redirect for safecopy_impl depends on link-time symbol adjacency (safecopy_impl immediately followed by safecopy_cont) with no static or build-time check enforcing that layout, unlike the exact-address match used for safefetch32/64.

**Suggestion:** Add a one-time startup assertion (e.g. in profiler init or a static constructor) that safecopy_cont's address is strictly greater than safecopy_impl's, so any future toolchain regression fails loudly at startup rather than silently losing fault protection.

jbachorik:
🗿 🤖 🔴

**File:** `ddprof-lib/src/test/cpp/safefetch_ut.cpp`

**[Sphinx Review — LOW]** No test exercises a fault partway through a multi-byte safeCopy (a readable prefix followed by an unreadable byte), so the newly-documented 'up to len-1 bytes already written to dst' partial-write behavior is unverified.

**Suggestion:** Add a test that maps two adjacent pages, unmaps the second, places src straddling the boundary with a known prefix length, and asserts both that safeCopy returns false and that dst[0..prefix) matches the readable bytes.

jbachorik (approved):
LGTM, assuming the small nits in the review are addressed.

PR #662, opened 2026-07-16
Share object across fuzz targets & increase build timeout

edznux-dd:
**What does this PR do?**:

Shared object are now reused instead of being recompiled for every target.
Also bumping the timeout to 1h30 (which should be plenty enough now).

The fuzz jobs are async & daily, having larger timeout shouldn't be noticeable for end users.

**Motivation**:

Fuzz tests generation where timing out.

**Additional Notes**:

**How to test the change?**:

**For Datadog employees**:
- [ ] JIRA: [JIRA-XXXX]


kaahos (approved):
looks good to me!

PR #664, opened 2026-07-16
feat: add core TaskBlock event recording support

kaahos:
**What does this PR do?**:

Adds the core TaskBlock event emission machinery. Recognized blocked intervals are recorded synchronously at block exit, with native APIs, JFR metadata, counters, and wall precheck integration.

**Motivation**:

Blocked threads should not need to receive wall-clock signals just to represent blocking time. TaskBlock events let the profiler report known blocking intervals directly.

**Additional Notes**:

This PR adds the generic recording path only. Concrete JVM and native blocking producers are added in follow-up PRs.

**How to test the change?**:

Automated coverage includes TaskBlock recorder unit tests, Java API tests, wall precheck behavior, disabled-state behavior, and combined wall-clock mitigation tests.

- `./.claude/commands/build-and-summarize testDebug`

**For Datadog employees**:

- [ ] If this PR touches code that signs or publishes builds or packages, or handles credentials of any kind, I've requested a security review (run the `dd:platform-security-review` skill, or file a request via the [PSEC review form](https://datadoghq.atlassian.net/jira/software/c/projects/PSEC/forms/form/direct/7861446195161534/37715)).
- [x] JIRA: [[JIRA-14931](https://datadoghq.atlassian.net/browse/PROF-14931)]


kaahos:
@codex review

ddprof-lib/src/main/cpp/event.h
```diff
@@ -58,7 +58,7 @@ class ExecutionEvent : public Event {
   OSThreadState _thread_state;
   ExecutionMode _execution_mode;
   u64 _weight;
-  u32 _call_trace_id;
+  u64 _call_trace_id;
```
jbachorik: 🗿 🤖 🔴

**[Sphinx Review — LOW]** ExecutionEvent::_call_trace_id is widened from u32 to u64, but the field is dead code — it is never read or written anywhere in the codebase (recordExecutionSample/recordMethodSample take call_trace_id as an explicit parameter, not from the event object). ExecutionEvent is stack-constructed once per CPU/wall-clock sample (itimer.cpp, ctimer_linux.cpp, perfEvents_linux.cpp, wallClock.cpp), so this unrelated change grows a hot-path struct for no functional reason.

**Suggestion:** Either remove the unused _call_trace_id field from ExecutionEvent entirely, or if it must stay for a future use, revert it to u32 in this PR since nothing here requires the wider type — keep the TaskBlock diff scoped to TaskBlock-related changes.

ddprof-lib/src/main/cpp/taskBlockRecorder.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+  if (!activity) {
+    Counters::increment(TASK_BLOCK_DROPPED_ROTATION);
+    return false;
+  }
+  if (!recording_enabled || !exited) {
+    profiler->leaveTaskBlockActivity();
+    return false;
+  }
+  if (!snapshot.context_eligible) {
```
jbachorik, line 67: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** finishTaskBlockAtExit() increments TASK_BLOCK_SKIPPED_TRACE_CONTEXT when the block-run's context-window eligibility check fails, but that counter (and its wc-prefixed sibling) is documented and used elsewhere to mean 'an active distributed-trace span suppressed the interval'. Conflating context-window crossing with trace-context presence under one counter name makes it impossible to distinguish the two causes from telemetry.

**Suggestion:** Add a distinct counter (e.g. TASK_BLOCK_SKIPPED_CONTEXT_WINDOW) for the context-window-crossing rejection path in finishTaskBlockAtExit, separate from the trace-context rejection in taskBlockPassesBasicEligibility, and cover both with a targeted counter assertion in the context-window test.

ddprof-lib/src/main/cpp/threadFilter.cpp
```diff
@@ -306,6 +319,45 @@ ThreadFilter::Slot* ThreadFilter::activeSlotForId(SlotID slot_id,
...
+    if (slot == nullptr) {
+        return false;
+    }
+    entry.slot = slot;
+    entry.lifecycle_generation = slot->lifecycleGeneration();
+    entry.recording_epoch = slot->recordingEpoch();
+    return true;
+}
+
+ThreadFilter::SlotID ThreadFilter::ensureCurrentThreadSlot(ProfiledThread* current) {
```
jbachorik, line 416: 🗿 🤖 🔴

**[Sphinx Review — LOW]** ThreadFilter::ensureCurrentThreadSlot(ProfiledThread*) is a new public method with no call sites anywhere in the codebase; javaApi.cpp still uses its own separate, duplicate static helper ensureCurrentThreadFilterSlot() for the same purpose.

**Suggestion:** Either wire javaApi.cpp's ensureCurrentThreadFilterSlot() call sites to use this new ThreadFilter::ensureCurrentThreadSlot() (removing the duplicate local helper), or remove the unused method if it was scaffolding for a follow-up PR.

ddprof-lib/src/main/cpp/threadFilter.h
```diff
@@ -224,18 +247,27 @@ class ThreadFilter {
...
+            resetSampledBlockGeneration();
+            disableUnownedBlockedFallback();
             *generation_out = generation;
             return true;
         }
-        inline void clearActiveBlockRun(OSThreadState state) {
+        inline void publishActiveBlockRun(OSThreadState state) {
+            active_block_state.store(state, std::memory_order_release);
+        }
+        inline void clearActiveBlockRun(OSThreadState) {
```
jbachorik: 🗿 🤖 🔴

**[Sphinx Review — HIGH]** Slot::clearActiveBlockRun() (invoked by ThreadFilter::exitBlockedRun/snapshotAndExitBlockedRun on every normal owned-block exit) resets unowned-blocked sampling counters but never re-enables unowned_blocked_fallback_enabled, which was disabled by tryPrepareActiveBlockRun() when the block run began. Only full lifecycle-reset paths (registerThread, refreshSlotForRecording, unregisterThreadLocked, resetRegistrationsLocked, clearActive, resetSlotRunState) call enableUnownedBlockedFallback(); resetSlotRunState currently has no production caller. As a result, once a thread completes even one owned TaskBlock/blockEnter interval, its slot's weighted unowned-blocked fallback sampling is disabled for the rest of that registration.

**Suggestion:** Call enableUnownedBlockedFallback() (or at least re-set unowned_blocked_fallback_enabled to 1) inside clearActiveBlockRun() so that ending an owned block restores eligibility for unowned-blocked weighted fallback sampling, matching the behavior at slot registration/reset.

jbachorik (changes requested):
🗿 🤖 🔴

ddprof-lib/src/main/cpp/wallClock.cpp
```diff
@@ -104,39 +103,29 @@ static inline WallPrecheckResult prepareWallPrecheck(ProfiledThread* current,
...
+  if (registry->activeOwnedBlockGeneration(entry, block_generation)) {
+    // Arm only after recordSample succeeds. A skipped JFR write must leave the
+    // run eligible so the next signal retries instead of losing its only stack.
+    result.owned_block_slot = slot;
+    result.owned_block_generation = block_generation;
+    return result;
+  }
+  if (!slot->unownedBlockedFallbackEnabled()) {
     return result;
   }
```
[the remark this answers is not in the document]
kaahos: addressed in [ea7f6db](https://github.com/DataDog/java-profiler/pull/664/commits/ea7f6dbf9a15ab1670ca2192e17174f61b698e13).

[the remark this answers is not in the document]
kaahos: true, but let's keep it that way for now.

ddprof-lib/src/main/cpp/event.h
kaahos: Nice catch, I've removed it in [8cc548a](https://github.com/DataDog/java-profiler/pull/664/commits/8cc548ab5904db346337271d71640fe1ed6415e3).

ddprof-lib/src/main/cpp/threadFilter.cpp
kaahos: addressed in [8cc548a](https://github.com/DataDog/java-profiler/pull/664/commits/8cc548ab5904db346337271d71640fe1ed6415e3).

ddprof-lib/src/main/cpp/taskBlockRecorder.cpp
kaahos: addressed in [8cc548a](https://github.com/DataDog/java-profiler/pull/664/commits/8cc548ab5904db346337271d71640fe1ed6415e3).

ddprof-lib/src/main/cpp/threadFilter.h
kaahos: addressed in [8cc548a](https://github.com/DataDog/java-profiler/pull/664/commits/8cc548ab5904db346337271d71640fe1ed6415e3).

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -417,14 +395,16 @@ Java_com_datadoghq_profiler_JavaProfiler_blockEnter0(
   if (current == nullptr) {
     return 0;
   }
-  ThreadFilter *tf = Profiler::instance()->threadFilter();
-  if (!tf->registryActive()) {
+  if (ContextApi::snapshot().spanId != 0) {
```
jbachorik: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** blockEnter0 now unconditionally computes a full Context snapshot (ContextApi::snapshot(), which touches numContextAttributes(), the OTel record's valid flag, and copies up to DD_TAGS_CAPACITY tag values) before checking the two cheap boolean gates (profiler->taskBlockEnabled() and tf->enabled()) that would otherwise let the call return immediately when tracking is off. The previous version checked tf->registryActive() (a single atomic load) first.

**Suggestion:** Check the cheap profiler->taskBlockEnabled()/tf->enabled() gates first and only call ContextApi::snapshot() once tracking is confirmed active, matching the ordering used elsewhere (cheapest check first).

jbachorik: 🗿 🤖 🔴

**[Sphinx Review — LOW]** No test in this PR would detect flipping the spanId != 0 check to == 0, or removing it entirely, in blockEnter0().

**Suggestion:** Add a test that calls blockEnter with trace context set via setTraceContext and verifies the token is 0 (rejection), and/or a test exercising the spanId == 0 acceptance path.

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -450,6 +430,60 @@ Java_com_datadoghq_profiler_JavaProfiler_blockExit0(
   }
 }
```
jbachorik, line 502: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** beginTaskBlock0 calls JVMSupport::isPlatformThread(env, thread) -- which performs a JNI GetVersion() call and, on JNI 19+, an actual JNI down-call to IsVirtualThread() -- unconditionally before checking the much cheaper profiler->isRunning()/taskBlockEnabled()/tf->unfilteredWallTrackingActive() flags that would otherwise reject the call immediately. When the feature is disabled or the profiler isn't running, every beginTaskBlock() call still pays for the JNI virtual-thread check for no benefit.

**Suggestion:** Reorder the checks so the cheap atomic reads (profiler->isRunning(), profiler->taskBlockEnabled(), tf->unfilteredWallTrackingActive()) run first, and only call JVMSupport::isPlatformThread() once those pass -- mirroring how endTaskBlock0 already short-circuits on decodeBlockRunToken() before calling isPlatformThread().

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -450,6 +430,60 @@ Java_com_datadoghq_profiler_JavaProfiler_blockExit0(
...
 
+extern "C" DLLEXPORT jlong JNICALL
+Java_com_datadoghq_profiler_JavaProfiler_beginTaskBlock0(
```
jbachorik, line 505: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** beginTaskBlock0/endTaskBlock0 take an explicit jthread argument used for stack-walking and virtual-thread classification, but never verify it corresponds to the JNI environment's current thread (the same thread whose ProfiledThread/ThreadFilter slot is being mutated). There is no test exercising a mismatched-thread call.

**Suggestion:** Either drop the jthread parameter and always resolve it internally via JNIEnv so the two identities cannot diverge, or add an explicit check before using thread for stack capture/classification.

ddprof-lib/src/main/cpp/jvmSupport.cpp
```diff
@@ -16,16 +16,38 @@
...
 volatile JVMSupport::JMethodIDLoadStats JVMSupport::jmethodID_load_state = JVMSupport::No_loaded;
 Mutex JVMSupport::_initialization_lock;
 
-
 // This method must be called after JVM has been properly initialized, e.g. after JVMTI::VMinit()
 // callback.
 // Currently, there are two paths lead to this call
 // - JVMTI::VMInit() callback (vmEntry.cpp)
 // - JavaProfiler.getInstance() via JNI down call - JVM must have been initialized
+bool JVMSupport::isPlatformThread(JNIEnv* jni, jthread thread) {
```
jbachorik, line 38: 🗿 🤖 🔴

**[Sphinx Review — HIGH]** isPlatformThread() indexes into JNINativeInterface_ at a hardcoded HotSpot-only offset (IS_VIRTUAL_THREAD_INDEX = 234) whenever GetVersion() >= JNI 19, with no VM::isHotspot()/isOpenJ9()/isZing() check anywhere on the call path (jvmSupport.cpp or its callers in javaApi.cpp). Every other vendor-specific branch in this same file (walkJavaStack, initExecution) is explicitly gated on VM vendor; this one is not.

**Suggestion:** Gate the function-table indexing behind VM::isHotspot() (returning true/conservatively-platform on other vendors, or routing to a vendor-specific detection mechanism), matching the pattern already used by JVMSupport::walkJavaStack/initExecution.

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -771,6 +780,92 @@ void Profiler::recordQueueTime(int tid, QueueTimeEvent *event) {
   _locks[lock_index].unlock();
 }
 
+Profiler::TaskBlockRecordResult Profiler::recordTaskBlock(
```
jbachorik, line 792: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** Profiler::recordTaskBlock() re-implements the lock-index selection, JVMTI GetStackTrace/copyJvmtiFrames, and call-trace-storage insertion sequence that Profiler::recordJVMTISample() already implements a few hundred lines above, instead of extracting a shared private helper. The duplicated code also silently omits the JDK21+ virtual-thread continuation-boundary synthetic frame fixup present in recordJVMTISample.

**Suggestion:** Extract the shared stack-capture-and-store logic into a private Profiler method that both recordJVMTISample and recordTaskBlock call with their respective arguments.

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -771,6 +780,92 @@ void Profiler::recordQueueTime(int tid, QueueTimeEvent *event) {
...
+    return false;
+  }
+  return true;
+}
+
+void Profiler::leaveTaskBlockActivity() {
+  _task_block_inflight.fetch_sub(1, std::memory_order_release);
+}
+
+void Profiler::beginTaskBlockRotation() {
```
jbachorik: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** beginTaskBlockRotation() spins on std::this_thread::yield() with no timeout, deadline, or exponential backoff, and is invoked synchronously from Profiler::stop() and Profiler::dump(). A single missed decrement anywhere in the tryEnterTaskBlockActivity()/leaveTaskBlockActivity() pairing (now or in a future edit) makes stop()/dump() hang forever with no diagnostic.

**Suggestion:** Add a bounded wait (deadline-based) with a fallback path that logs/counts a timeout and proceeds, similar to other drain-with-timeout patterns already used elsewhere in Profiler::stop().

ddprof-lib/src/main/cpp/taskBlockRecorder.cpp
```diff
@@ -0,0 +1,85 @@
...
+  u64 min_ticks = g_min_task_block_ticks.load(std::memory_order_acquire);
+  if (min_ticks == 0) min_ticks = computeMinTaskBlockTicks();
+  return end_ticks > start_ticks && end_ticks - start_ticks >= min_ticks;
+}
+
+bool recordTaskBlockAtExit(ProfiledThread* current, ThreadFilter* thread_filter,
+                           jthread thread, int start_depth, u64 block_token,
+                           ThreadFilter::SlotID slot_id, u64 generation,
+                           u64 blocker, u64 unblocking_span_id) {
+  u64 start_ticks = 0;
```
jbachorik, line 30: 🗿 🤖 🔴

**[Sphinx Review — LOW]** recordTaskBlockAtExit() re-validates slot_id/generation against ThreadFilter::tokenSlotId(block_token)/tokenGeneration(block_token), but its only caller (endTaskBlock0) computes slot_id/generation by decoding that same block_token immediately beforehand, so the check can never be false.

**Suggestion:** Either remove the redundant re-decode/check, or have recordTaskBlockAtExit decode the token itself and drop the separate parameters.

ddprof-lib/src/main/cpp/taskBlockRecorder.cpp
```diff
@@ -0,0 +1,85 @@
...
+      generation != ThreadFilter::tokenGeneration(block_token)) {
+    return false;
+  }
+
+  return finishTaskBlockAtExit(
+      current, thread_filter, thread, start_depth, block_token, start_ticks,
+      context, blocker, unblocking_span_id);
+}
+
+bool finishTaskBlockAtExit(ProfiledThread* current,
```
jbachorik, line 41: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** finishTaskBlockAtExit() manually pairs tryEnterTaskBlockActivity()/leaveTaskBlockActivity() across several early-return branches, while recordTaskBlockIfEligible() in the same header uses the RAII TaskBlockActivity guard for the same lifecycle. The manual path directly gates an unbounded (no-timeout) spin-wait in Profiler::beginTaskBlockRotation(), called from stop()/dump(); a future added early return that forgets the matching leave() call would hang teardown indefinitely instead of failing loudly.

**Suggestion:** Have finishTaskBlockAtExit() use the same TaskBlockActivity RAII guard as recordTaskBlockIfEligible() (or wrap it in a small scope-guard) so every future early return releases automatically, and/or add a bounded timeout/log-and-proceed fallback to beginTaskBlockRotation() so a stray leak degrades instead of hanging stop()/dump() forever.

ddprof-lib/src/main/cpp/taskBlockRecorder.cpp
```diff
@@ -0,0 +1,85 @@
...
+  ThreadFilter::SlotID slot_id = ThreadFilter::tokenSlotId(block_token);
+  u64 generation = ThreadFilter::tokenGeneration(block_token);
+  ThreadFilter::SlotID current_slot = current->filterSlotId();
+  if (current_slot < 0) {
+    current_slot = thread_filter->slotIdByTid(current->tid());
+  }
+  BlockRunSnapshot snapshot;
+  bool exited = current_slot == slot_id &&
+      thread_filter->snapshotAndExitBlockedRun(slot_id, generation, &snapshot);
+
```
jbachorik, line 59: 🗿 🤖 🔴

**[Sphinx Review — LOW]** If the calling thread's filterSlotId() (or its tid-based fallback lookup) ever diverges from the slot_id encoded in an outstanding, still-valid block_token, finishTaskBlockAtExit skips snapshotAndExitBlockedRun entirely, even though ProfiledThread::taskBlockExit() has already consumed and zeroed the thread's own token, leaving the ThreadFilter slot's owned block run active/unclearable via the normal exit path.

**Suggestion:** Add a regression test that forces current->filterSlotId() to diverge from the token encoded slot_id and assert the ThreadFilter slot is still cleared (or the mismatch is at least counted/logged) rather than silently leaked.

ddprof-lib/src/main/cpp/threadFilter.h
```diff
@@ -35,6 +37,14 @@ enum class BlockRunOwner : int {
     NATIVE = 3,
 };
 
+struct BlockRunSnapshot {
```
jbachorik, line 41: 🗿 🤖 🔴

**[Sphinx Review — LOW]** BlockRunSnapshot exposes owner, generation, and active fields that are populated by snapshotBlockRun() but never consumed by the sole production caller (taskBlockRecorder.cpp's finishTaskBlockAtExit, which reads only active_state and context_eligible).

**Suggestion:** Either trim BlockRunSnapshot to the two fields actually consumed today, or add a short comment noting which follow-up consumer needs owner/generation/active.

ddprof-lib/src/main/cpp/threadFilter.h
```diff
@@ -224,18 +247,27 @@ class ThreadFilter {
                                          std::memory_order_release);
                 return false;
             }
-            u32 generation = block_generation.fetch_add(1, std::memory_order_acq_rel) + 1;
+            u64 generation = block_generation.load(std::memory_order_relaxed);
+            if (generation == kMaxBlockRunGeneration) {
```
jbachorik, line 256: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** Slot::block_generation is a monotonic per-slot counter that is never reset across thread unregister/reuse (clearActiveBlockRun/enableUnownedBlockedFallback/resetSlotRunState do not touch it). Once it reaches kMaxBlockRunGeneration (2^53), tryPrepareActiveBlockRun() permanently refuses every future block-run entry on that slot, and the failure is indistinguishable from ordinary contention/context-window rejection (no dedicated counter fires), so the degradation is silent and permanent for the process lifetime.

**Suggestion:** Increment a dedicated counter (e.g. TASK_BLOCK_GENERATION_SATURATED) in the kMaxBlockRunGeneration branch so the condition is at least observable, and consider whether the slot generation can be safely reset when the slot is fully unregistered rather than only on process exit.

ddprof-lib/src/main/cpp/threadFilter.h
jbachorik: 🗿 🤖 🔴

**[Sphinx Review — LOW]** Slot::clearActiveBlockRun(OSThreadState) takes a parameter that is completely unused in the function body (it used to feed resetSampledRun(state) before the once-per-run sampling mechanism was replaced by sampled_block_generation). All 10 call sites still pass either OSThreadState::UNKNOWN or OSThreadState::RUNNABLE, giving the false impression that the argument affects behavior.

**Suggestion:** Drop the now-unused OSThreadState parameter from clearActiveBlockRun() and update all call sites.

ddprof-lib/src/main/cpp/wallClock.cpp
```diff
@@ -104,39 +103,29 @@ static inline WallPrecheckResult prepareWallPrecheck(ProfiledThread* current,
...
-      incrementSuppressedSampledRun();
-      result.suppress = true;
-      return result;
-    }
-    // Arm only after the MethodSample has been successfully recorded. If the
-    // JFR write is skipped due to lock contention, the next signal must retry
-    // instead of losing the only stack for this blocked run.
-    result.slot_to_arm = slot;
-    result.state_to_arm = active_block_state;
+  ThreadEntry entry{current->tid(), slot, slot->lifecycleGeneration(),
```
jbachorik, line 120: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** prepareWallPrecheck() calls isOwnedBlockSuppressionCandidate() and, on failure, immediately calls activeOwnedBlockGeneration() with the same ThreadEntry. Both funnel into the private ThreadFilter::ownedBlockGeneration() helper, which re-reads and re-validates the same slot fields a second time for the very common case where the block is active but has not yet been sampled. This redundant computation runs inside the async-signal-handler hot path on every wall-clock signal for a wallprecheck-armed thread.

**Suggestion:** Expose a single ThreadFilter method that classifies the owned-block state once so prepareWallPrecheck() only reads/validates the slot's owned-block payload one time per signal instead of calling ownedBlockGeneration() twice with different require_sampled values.

jbachorik (changes requested):
🗿 🤖 🔴

ddprof-lib/src/main/cpp/javaApi.cpp
kaahos: addressed in [cbcfd2d](https://github.com/DataDog/java-profiler/pull/664/commits/cbcfd2dfa).

ddprof-lib/src/main/cpp/javaApi.cpp
kaahos: addressed in [cbcfd2d](https://github.com/DataDog/java-profiler/pull/664/commits/cbcfd2dfa).

ddprof-lib/src/main/cpp/jvmSupport.cpp
kaahos: addressed in [cbcfd2d](https://github.com/DataDog/java-profiler/pull/664/commits/cbcfd2dfa).

ddprof-lib/src/main/cpp/taskBlockRecorder.cpp
kaahos: addressed in [cbcfd2d](https://github.com/DataDog/java-profiler/pull/664/commits/cbcfd2dfa).

ddprof-lib/src/main/cpp/taskBlockRecorder.cpp
kaahos: addressed in [cbcfd2d](https://github.com/DataDog/java-profiler/pull/664/commits/cbcfd2dfa).

ddprof-lib/src/main/cpp/javaApi.cpp
kaahos: addressed in [cbcfd2d](https://github.com/DataDog/java-profiler/pull/664/commits/cbcfd2dfa).

ddprof-lib/src/main/cpp/javaApi.cpp
kaahos: addressed in [cbcfd2d](https://github.com/DataDog/java-profiler/pull/664/commits/cbcfd2dfa).

ddprof-lib/src/main/cpp/profiler.cpp
kaahos: addressed in [cbcfd2d](https://github.com/DataDog/java-profiler/pull/664/commits/cbcfd2dfa).

ddprof-lib/src/main/cpp/taskBlockRecorder.cpp
kaahos: addressed in [cbcfd2d](https://github.com/DataDog/java-profiler/pull/664/commits/cbcfd2dfa).

ddprof-lib/src/main/cpp/wallClock.cpp
kaahos: addressed in [cbcfd2d](https://github.com/DataDog/java-profiler/pull/664/commits/cbcfd2dfa).

ddprof-lib/src/main/cpp/threadFilter.h
kaahos: addressed in [cbcfd2d](https://github.com/DataDog/java-profiler/pull/664/commits/cbcfd2dfa).

ddprof-lib/src/main/cpp/profiler.cpp
kaahos: Fixed in ebdd16f71: extracted `Profiler::captureJVMTIFrames()` (declared in profiler.h, implemented in profiler.cpp) covering the JVMTI GetStackTrace/copyJvmtiFrames sequence plus the JDK21+ virtual-thread continuation-boundary fixup. Both `recordJVMTISample` and `recordTaskBlock` now call this shared helper; each keeps its own distinct failure-branching/put() semantics, which genuinely differ between the two callers. Note: I traced recordTaskBlock's only call path (endTaskBlock0/beginTaskBlock0 in javaApi.cpp) and confirmed it rejects non-platform threads via JVMSupport::isPlatformThread, so the VT continuation branch is currently unreachable from that path -- the fix is still correct dedup/future-proofing, but not independently integration-testable at the TaskBlock layer today.

ddprof-lib/src/main/cpp/threadFilter.h
kaahos: Fixed in ebdd16f71: added a THREAD_REGISTRY_BLOCK_GENERATION_SATURATED counter incremented in tryPrepareActiveBlockRun's saturation branch, and reset block_generation to 0 at every genuine slot-occupancy-change point (registerThread's reused-slot branch, resetRegistrationsLocked) alongside the existing lifecycle_generation bump -- traced every consumer of block_generation/lifecycle_generation to confirm resetting is safe: both the token-based (exitBlockedRun/snapshotAndExitBlockedRun) and entry-snapshot-based (ownedBlockGeneration) validation paths already have an independent invalidation mechanism (active_block_owner clearing, lifecycle_generation bump) that fires on slot reuse regardless of the numeric generation value. Also added a THREAD_REGISTRY_UNREGISTER_ACTIVE_BLOCK_RUN canary counter in unregisterThreadLocked to make the one residual precondition (no slot reuse while a run is genuinely still active) independently observable rather than just assumed. New tests: SaturatedGenerationIsCountedAndRecoversAfterSlotReuse, UnregisterWhileBlockRunActiveIsCounted (threadFilter_ut now 44 tests, up from 42).

ddprof-lib/src/main/cpp/threadFilter.h
kaahos: Fair point that within #664 alone, owner/generation/active are genuinely unused -- finishTaskBlockAtExit only reads active_state/context_eligible. The fields aren't hypothetical though: #665 (queued after this one) already reads snapshot.active/owner/generation in monitorBlockEnter() to detect JVMTI-owned reentrant monitor blocks (threadFilter.cpp:629-632), and #666 reads snapshot.active/owner similarly -- but I can't guarantee those land unchanged, so that's not a reason to keep unused fields in #664 by itself.

kaahos: Trimmed in 672a6f5d4: BlockRunSnapshot now only has active_state/context_eligible, matching finishTaskBlockAtExit's actual usage. owner/generation will be reintroduced in #665 when monitorBlockEnter() actually needs them.

PR #669, opened 2026-07-17
feat(nativemem): categorized native-memory accounting — first cut

rkennke:
## What

Adds a `NativeMem` facility to measure the **profiler's own native memory usage** — currently we only observe whole-process RSS. Per-category live gauge, moving-window **average**, and a **precise per-category peak**, wired through the existing counter/JFR path.

## Why

We measure the agent's native footprint only as an RSS delta (~150–200 MB) with no breakdown. This gives an in-process, attributable, categorized number we can trend, and a foundation for later pinpointing which code is responsible.

## How

- **`NativeMem`** (`nativeMem.h` / `nativeMem.cpp`): a `NativeMemCategory` enum whose per-category live gauges **partition the total** — each backing allocation belongs to exactly one category, so no double-counting.
- **Always-on**, independent of the `COUNTERS` build flag: `record()` is a single relaxed atomic add. Most sites are off the signal path (malloc/new is not async-signal-safe anyway); the exception is **CALLTRACE**, whose arena allocates inside the sampling signal handler (via `OS::safeAlloc`), so `record()` is kept async-signal-safe there.
- **Precise per-category peak**: `record()` maintains each category's high-water mark at allocation time via a relaxed CAS (common path is load+compare; CAS only on a new peak; frees skip it). Spikes between sample ticks are still captured.
- **`sample()`** refreshes the moving-window averages and the observed total, ticked once per JFR chunk finish.
- **Exposure**: totals mirror into `NATIVE_MEM_{LIVE,AVG,MAX}_BYTES` (JFR `T_DATADOG_COUNTER` + JNI debug counters); per-category `native_mem_{live,avg,max}_bytes.<category>` values reuse the same event format — no new event type.

## Total peak: bracketed, no shared hotspot

A *precise* total peak would need a single global atomic hit by every allocation (a contention hotspot). Instead the total peak is bracketed:
- `NATIVE_MEM_MAX_BYTES` = **upper bound** = sum of the precise per-category peaks.
- `native_mem_max_observed_total_bytes` = **lower bound** = the largest instantaneous total seen at a sampling tick.

## Instrumented categories

Tagged via `record()` at their backing alloc/free sites (precise, always-on):
- **CALLTRACE** — `LinearAllocator` chunks (call-trace arena) + per-shard calltrace buffers.
- **DICTIONARY** — both implementations: the older per-key-malloc `Dictionary` (symbols/packages — root/overflow `DictTable`s + key strings, whose size is recovered via `strlen` at bulk-free), and the arena `StringDictionary` (arena chunks + root/overflow `SBTable`s; keys live inside chunks so they're not counted separately). Counted unconditionally, independent of the diagnostic counters' offset gate. Single category — a per-role split, if ever wanted, belongs in a nested dimension.
- **THREAD_LOCAL** — `ProfiledThread` per thread.
- **JFR_BUFFERS** — the `Recording` object (embeds the JFR buffer array).
- **LINE_TABLES** — malloc'd JVMTI line-number table copies.
- **PERF** — perf ring `mmap` (2 × page_size).
- **THREAD_FILTER** — `ChunkStorage` chunks (bounded) + free-list array.

Gauge-mirrored (`NativeMem::setLive()` at a recompute site):
- **NATIVE_SYMBOLS** — the profiler's per-native-library symbol tables used to symbolicate native frames (async-profiler's `CodeCache`; not the JVM code cache). Sourced from `CodeCache::memoryUsage()`, which is now accurate on `main` (#677 merged); the gauge is refreshed at dump/stop, the same cadence as the existing `CODECACHE_NATIVE_SIZE_BYTES` counter.

**Not yet covered (reads 0):**
- **The long tail** — scattered `new`/`malloc` across the remaining files → best captured by malloc interception, not hand-tagging.

(`CONTEXT` was removed: the OTel context record is embedded in `ProfiledThread`, already counted under `THREAD_LOCAL`; a separate category would double-count or read a permanent 0.)

So the total is still a **subset** of the RSS delta until interception lands, but now covers the major consumers.

## Avoiding double-counting

Call-trace bytes appear at three layers (`safeAlloc` mmap, `LINEAR_ALLOCATOR_BYTES` reserved chunks, `CALLTRACE_STORAGE_BYTES` used-within-chunks); summing counters would double/triple-count. `NativeMem` counts **backing memory once per category**; the existing reserved/used/waste counters remain an independent **nested** diagnostic dimension. The dictionary tagging follows the same rule — e.g. arena keys are inside chunks and are not counted separately.

## Performance

Designed to be off the hot path. Measured in operations, not observed as a regression in this workload:

- **Per sample (signal handler): zero in the common case.** A sample's `CallTraceStorage::put` normally just bump-allocates within an existing arena chunk and touches no `record()`. `record()` runs only when the call-trace arena grows a new chunk (8 MiB apart), and there it's a lock-free relaxed atomic add + high-water update — async-signal-safe.
- **Per backing allocation: ~1–2 relaxed atomic ops.** `record()` is one relaxed `fetch_add`; on allocation it also does a relaxed load + compare, with a CAS only when a new per-category peak is set (rare, since the peak is monotonic). This sits next to a `malloc`/`calloc`/`mmap`/`new` that dominates the cost, so it's negligible. These sites (dictionary inserts, chunk/table growth, thread/lib/recording creation) are not per-sample.
- **Per JFR chunk finish: a fixed, tiny cost.** `sample()` is O(categories × window) ≈ 9 × 64 folds; `writeNativeMem()` emits ~28 counter events. This runs once per chunk rotation (seconds–minutes), off the sampling path.
- **One extra pass at dictionary `clear()`:** the per-key free now also does a `strlen` to recover the freed size, so `clear()` cost is O(sum of key lengths) rather than O(key count). `clear()` already walks every key; this is a marginal constant-factor increase, at dictionary rotation only.

No new locks, no shared global counter on the hot path (the total peak is bracketed precisely to avoid one), and the always-on atomics are relaxed. `sample()`/`writeNativeMem()` run within the existing recording flush.

## Roadmap

1. **This PR** — facility + precise per-category max (bracketed total) + the major consumers, including DICTIONARY.
2. **Malloc interception** — capture the untagged long tail so categories sum toward the RSS delta, plus a reconciliation check to quantify what's still unattributed.

## Testing

- `nativeMem_ut` (7 tests): live tracking + total partition, single-sample avg==live, moving average, max high-water, precise per-category max catching an inter-sample spike (with the bracket), `setLive` gauge + peak retention, category names. (The non-negative invariant is enforced by an assert in `record()` rather than a dedicated clamp test.)
- **Lifecycle invariants** for both dictionary implementations (`dictionary_ut`, `stringDictionary_ut`): accounting grows on insert, returns exactly to the construction baseline after `clear()`, and to zero after destruction — directly proving inc/dec pairing (incl. the strlen-at-free path and arena chunk growth).
- **THREAD_LOCAL lifecycle** (`thread_teardown_safety_ut`, Linux): `NM_THREAD_LOCAL` returns to baseline after both `release()` and thread-exit teardown, proving the `forTid`/`freeValue` pairing.
- Full `:ddprof-lib:gtestDebug` suite green (**358 cross-platform tests, 0 failures** locally; Linux-only perf/thread-teardown coverage runs in CI).

### DIFF
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -1780,24 +1780,37 @@ void Recording::writeLogLevels(Buffer *buf) {
 }
 
 void Recording::updateNativeMemStats() {
-  // Fold the per-category live gauges into their moving-window averages and
-  // running peaks. This is a sampled max: peaks that rise and fall entirely
-  // between two chunk finishes are not observed. A spike-accurate max would
-  // track the high-water mark at allocation time.
+  // Refresh the moving-window averages and the observed total peak. Per-category
+  // peaks are maintained precisely at allocation time, so they are not sampled
+  // here; the total peak is bracketed instead (see writeNativeMem).
   NativeMem::sample();
 
   // Mirror the totals into the flat counter table so they flow out through the
   // existing counter path (JFR T_DATADOG_COUNTER events and the JNI debug
-  // counters). Per-category values are emitted separately by writeNativeMem().
+  // counters). NATIVE_MEM_MAX_BYTES carries the upper bound on the total peak
+  // (sum of precise per-category peaks); the observed lower bound and the
+  // per-category values are emitted by writeNativeMem().
   Counters::set(NATIVE_MEM_LIVE_BYTES, NativeMem::liveTotal());
   Counters::set(NATIVE_MEM_AVG_BYTES, NativeMem::avgTotal());
   Counters::set(NATIVE_MEM_MAX_BYTES, NativeMem::maxTotal());
 }
 
 void Recording::writeNativeMem(Buffer *buf) {
-  // Emit live/avg/max per category as counter events, reusing the counter event
-  // format with a "<metric>.<category>" name so they land alongside the totals
-  // without needing a dedicated event type or a slot in the counter table.
+  // Emit native-memory stats as counter events, reusing the counter event format
+  // so they land alongside the totals without needing a dedicated event type or
+  // a slot in the counter table.
+  auto emit = [&](const char *label, long long value) {
+    int start = buf->skip(1);
+    buf->putVar64(T_DATADOG_COUNTER);
+    buf->putVar64(_start_ticks);
+    buf->putUtf8(label);
+    buf->putVar64(value);
+    writeEventSizePrefix(buf, start);
+    flushIfNeeded(buf);
+  };
+
+  // Per-category live/avg/max, named "<metric>.<category>". The max here is the
+  // precise per-category peak tracked at allocation time.
   for (int c = 0; c < NM_NUM_CATEGORIES; c++) {
     NativeMemCategory cat = (NativeMemCategory)c;
     const char *name = NativeMem::categoryName(cat);
@@ -1812,15 +1825,14 @@ void Recording::writeNativeMem(Buffer *buf) {
     for (const auto &m : metrics) {
       char label[64];
       snprintf(label, sizeof(label), "%s%s", m.prefix, name);
-      int start = buf->skip(1);
-      buf->putVar64(T_DATADOG_COUNTER);
-      buf->putVar64(_start_ticks);
-      buf->putUtf8(label);
-      buf->putVar64(m.value);
-      writeEventSizePrefix(buf, start);
-      flushIfNeeded(buf);
+      emit(label, m.value);
     }
   }
+
+  // The total peak is bracketed: NATIVE_MEM_MAX_BYTES already carries the upper
+  // bound (sum of precise per-category peaks); here we also emit the observed
+  // lower bound (largest instantaneous total seen at a sampling tick).
+  emit("native_mem_max_observed_total_bytes", NativeMem::maxTotalObserved());
 }
 
 void Recording::writeCounters(Buffer *buf) {
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -1794,7 +1794,7 @@ void Recording::updateNativeMemStats() {
   // Mirror the totals into the flat counter table so they flow out through the
   // existing counter path (JFR T_DATADOG_COUNTER events and the JNI debug
   // counters). NATIVE_MEM_MAX_BYTES carries the upper bound on the total peak
-  // (sum of precise per-category peaks); the observed lower bound and the
+  // (sum of precise per-category peaks); the observed sampled total and the
   // per-category values are emitted by writeNativeMem().
   Counters::set(NATIVE_MEM_LIVE_BYTES, NativeMem::liveTotal());
   Counters::set(NATIVE_MEM_AVG_BYTES, NativeMem::avgTotal());
@@ -1806,6 +1806,13 @@ void Recording::writeNativeMem(Buffer *buf) {
   // so they land alongside the totals without needing a dedicated event type or
   // a slot in the counter table.
   auto emit = [&](const char *label, long long value) {
+    // Clamp to 0 before encoding: the value is serialized as an unsigned varint
+    // (putVar64), so a negative live gauge would emit a huge value and corrupt
+    // the counter stream. avg/max are already non-negative; live is clamped
+    // here to match sample()/liveTotal().
+    if (value < 0) {
+      value = 0;
+    }
     int start = buf->skip(1);
     buf->putVar64(T_DATADOG_COUNTER);
     buf->putVar64(_start_ticks);
@@ -1835,9 +1842,9 @@ void Recording::writeNativeMem(Buffer *buf) {
     }
   }
 
-  // The total peak is bracketed: NATIVE_MEM_MAX_BYTES already carries the upper
-  // bound (sum of precise per-category peaks); here we also emit the observed
-  // lower bound (largest instantaneous total seen at a sampling tick).
+  // NATIVE_MEM_MAX_BYTES already carries the upper bound on the total peak (sum
+  // of precise per-category peaks); here we also emit the largest observed
+  // sampled total (a non-atomic per-category sum; approximate).
   emit("native_mem_max_observed_total_bytes", NativeMem::maxTotalObserved());
 }
 
@@ -2142,8 +2149,12 @@ void FlightRecorder::stop() {
   if (rec != nullptr) {
     // NULL first, deallocate later
     _rec = nullptr;
-    NativeMem::record(NM_JFR_BUFFERS, -(long long)sizeof(Recording));
+    // Decrement AFTER delete: ~Recording() runs finishChunk(), which emits the
+    // native-memory counters for the final chunk. The Recording buffers are
+    // still live during that serialization, so account the free only once it
+    // has actually happened.
     delete rec;
+    NativeMem::record(NM_JFR_BUFFERS, -(long long)sizeof(Recording));
   }
 }
 
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -12,6 +12,7 @@
 #include "context.h"
 #include "context_api.h"
 #include "counters.h"
+#include "nativeMem.h"
 #include "dictionary.h"
 #include "flightRecorder.inline.h"
 #include "incbin.h"
@@ -810,7 +811,9 @@ off_t Recording::finishChunk(bool end_recording, bool do_cleanup) {
   // dictionary) will reflect the previous serialization. That is, some level of
   // familiarity with the code base will be required to use this diagnostic
   // information for now.
+  updateNativeMemStats();
   writeCounters(_buf);
+  writeNativeMem(_buf);
 
   // Keep a simple stats for where we failed to unwind
   // For the sakes of simplicity we are not keeping the count of failed unwinds which would also be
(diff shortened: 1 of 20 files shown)


ddprof-lib/src/main/cpp/nativeMem.cpp
```diff
@@ -0,0 +1,112 @@
...
+    long long v = load(_live[c]);
+    // A category's live bytes are never negative under correct pairing (asserted
+    // in record()). This clamp is a release-mode safety net: should an accounting
+    // bug slip past the assert under NDEBUG, it keeps a negative from skewing the
+    // window average and total rather than propagating garbage.
+    if (v < 0) {
+      v = 0;
+    }
+    _window[c][_window_pos] = v;
+    total += v;
```
kaahos, line 51: here `total` might not necessarily be the lower bound in the case where we have:
- read category A (read in `v`)
- free category A
- allocate category B (read in `v`)

`total` would be `vA + vB` while it should be `max(vA, vB)` or something like that I guess?

ddprof-lib/src/test/cpp/dictionary_ut.cpp
kaahos: it should be updated to 2026 I think.

ddprof-lib/src/test/cpp/stringDictionary_ut.cpp
```diff
@@ -1,13 +1,42 @@
 #include <gtest/gtest.h>
```
kaahos, line 6: Datadog copyright header should be added here.

ddprof-lib/src/main/cpp/threadLocalData.h
kaahos: `NM_THREAD_LOCAL` is not decremented here while it is incremented in `forTid()` meaning that it after this call and the release of `pt`, it would have an incorrect value.

ddprof-lib/src/main/cpp/perfEvents_linux.cpp
kaahos, line 907: `NM_PERF` accounts for the per-thread ring mappings, but this allocation is not included. is it intentional? I am not sure this should be included but, if `NativeMem` measures nominal requested/reserved allocation sizes, this calloc and its matching free should have paired `NM_PERF` accounting I presume. what do you think?

kaahos (commented):
Thank you for the work, I've left some comments! let me know what you think.

rkennke:
Thanks all — worked through every comment. Pushed as four commits (`957e13a13` and the three before it).

**Correctness**
- **Negative → `putVar64(u64)` corruption (Copilot ×3):** clamp per-category live to 0 at the emit boundary (`liveTotal()` and `writeNativeMem`), matching `sample()`. The `record()` invariant asserts non-negative in debug; this guards the release path where the assert is stripped.
- **`NM_JFR_BUFFERS` decrement before `delete rec` (codex):** moved after `delete rec`. `~Recording()` runs `finishChunk()`, which emits the counters for the final chunk while the buffers are still live, so the free is now accounted only once it happens.
- **Calltrace resize OOM partial-swap (Copilot):** rewrote as two-phase — allocate all `CONCURRENCY_LEVEL` buffers first, and only swap/free/account if all succeed. A mid-loop failure now leaves the profiler entirely unchanged (also fixes the pre-existing `_max_stack_depth = 0` partial-state bug).

**Coverage / completeness**
- **`NATIVE_SYMBOLS` stale on stop-without-dump (codex):** extracted `updateNativeLibMemStats()` and call it from both `dump()` and `stop()` (before `_jfr.stop()`), so the final chunk reflects native-symbol memory on the stop path too.
- **`native_libs.memoryUsage()` called 3× (Copilot):** the new helper reads it once.
- **perf `_events` array not counted (@kaahos):** you're right — added paired `NM_PERF` inc/dec around the `calloc`/`free` in `start()` (it previously counted only the ring mappings).

**`NM_THREAD_LOCAL` not decremented at `release()` (@kaahos):** it *is* balanced — the decrement lives in `freeValue()`, which is reached both via `release()` → `ThreadLocal::clear()` → the cleanup callback `F(p)`, and via the pthread-key destructor on thread exit. `forTid()` is the only allocation site and `freeValue()` the only free site. Added two lifecycle tests in `thread_teardown_safety_ut` that assert `NM_THREAD_LOCAL` returns to baseline after both `release()` and thread-exit, to make that explicit and guard against regressions.

**`total` not a strict lower bound (@kaahos):** correct — the per-category sum in `sample()` isn't an atomic snapshot, so it can drift above/below a true instantaneous total. Reworded the docs to call it an approximate sampled total; `maxTotal()` (sum of precise per-category peaks) remains the authoritative ceiling.

**Nits (@kaahos):** `dictionary_ut.cpp` → `2025, 2026`; added the Datadog header to `stringDictionary_ut.cpp`.

Also updated the PR description (test count 8→7 with the invariant now asserted rather than clamp-tested; noted the new THREAD_LOCAL coverage). Cross-platform suite is green locally (358); the Linux-only perf and thread-teardown additions run in CI.

rkennke:
Both addressed in `63a203c70`:

- **`deleteForTest()` bypasses the decrement (Copilot):** right — it's the UNIT_TEST stand-in for freeValue()'s delete, so it now mirrors the `NM_THREAD_LOCAL` decrement. Prevents the cross-test leak.
- **`~ThreadFilter()` decrements `_free_list` before the unique_ptr frees it (Copilot):** reordered to account after the free — `_free_list.reset()` before the decrement, and the chunk `delete` before its decrement — same "account the free once it has happened" principle as the JFR_BUFFERS fix.

Cross-platform suite green (358); the Linux-only bits run in CI.

rkennke:
Two Copilot points here:

**`profiler.cpp:1747` — `CODECACHE_RUNTIME_STUBS_SIZE_BYTES` = `native_libs.memoryUsage()`.** Correct, and it's a pre-existing copy-paste that this PR only *relocated* into `updateNativeLibMemStats()` (the value is unchanged from `main`). The actual fix — sourcing it from `JitCodeCache::_runtime_stubs` under its lock — is **#681**, kept separate so it doesn't collide with this PR's refactor. Not duplicating it here.

**`flightRecorder.cpp:1802` — `updateNativeMemStats()` doesn't refresh `NM_NATIVE_SYMBOLS` before `sample()`.** True: `NATIVE_SYMBOLS` is a gauge refreshed at `dump()`/`stop()` (where `Libraries::refresh()` re-scans the library set), so between those points a chunk-finish samples its last value. That's the **same cadence as the existing `CODECACHE_NATIVE_SIZE_BYTES` counter** — which is likewise only set at dump/stop and re-emitted per chunk — so this isn't a regression, just the gauge's nature. Making it per-chunk-fresh would require `Recording::updateNativeMemStats()` to reach `Libraries` (a layering it doesn't currently have) or re-scan libraries every chunk; I'd do that as a separate follow-up if per-chunk freshness is wanted, and it would apply equally to the existing CODECACHE counter. The record()-tracked categories (CALLTRACE, DICTIONARY, THREAD_LOCAL, …) are always live per chunk; only the two gauge-style ones (NATIVE_SYMBOLS, CODECACHE) have this cadence.

rkennke:
Reordered the calltrace-buffer decrement to after `free(prev)` in the latest commit, matching the other sites (`FlightRecorder::stop`, `~ThreadFilter`). Functionally equivalent here (the free doesn't read the counter), done for consistency.

rkennke:
Reordered the `NM_PERF` decrement to after `free(_events)` (capturing the old size first, since `_max_events` is overwritten by the realloc) — consistent with the calltrace-buffer and `~ThreadFilter` sites. `perfEvents_linux.cpp` is Linux-only so it's validated by CI rather than the local macOS build.

rkennke:
Reordered the `NM_THREAD_LOCAL` decrement to after `delete pt` in `freeValue()`, consistent with the other decrement sites (calltrace resize, perf `_events`, `~ThreadFilter`). `sizeof` is a compile-time constant so nothing is lost by deleting first. Verified by the THREAD_LOCAL lifecycle tests (release + thread-exit both return the gauge to baseline).

rkennke:
Swept all remaining `NativeMem::record` decrements to run after the `free()` — StringArena chunks, SBTables, and the Dictionary key strings + overflow tables (capturing `strlen+1` before the key free, since it reads the string). Now uniform with the calltrace/perf/thread-local/ThreadFilter sites. Dictionary/StringDictionary lifecycle tests still green (balance returns to baseline).

jbachorik (approved):
Nice work!
Can you plz consider adding some smoke tests? Mybe extending the existing tests to assert the presence of the NM accounting counters and that they are sane?

ddprof-lib/src/test/cpp/dictionary_ut.cpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025 Datadog, Inc
+ * Copyright 2025, 2026 Datadog, Inc
```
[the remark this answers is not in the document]
rkennke: Fixed in 9ebd61591 — copyright line is now `Copyright 2025, 2026, Datadog, Inc.` (comma + trailing period), matching the tree convention. The nativeMem.{h,cpp} and nativeMem_ut.cpp headers in the same commit were switched to the short-form copyright + SPDX identifier.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -1776,6 +1785,69 @@ void Recording::writeLogLevels(Buffer *buf) {
...
+      value = 0;
+    }
+    int start = buf->skip(1);
+    buf->putVar64(T_DATADOG_COUNTER);
+    buf->putVar64(_start_ticks);
+    buf->putUtf8(label);
+    buf->putVar64(value);
+    writeEventSizePrefix(buf, start);
+    flushIfNeeded(buf);
+  };
```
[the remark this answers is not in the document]
rkennke: `-DCOUNTERS` is defined unconditionally in every build preset: it is part of `commonLinuxCompilerArgs` (build-logic `ConfigurationPresets.kt:81`), which is composed into all Linux and macOS release/debug variants plus the asan/tsan/fuzzer configs. There is no shipping or test build that compiles without it, so `Counters::set()` / `writeCounters()` always emit these totals in practice — the same guarantee the existing counter-based tests rely on (e.g. `ContextCpuTest` asserts `calltrace_storage_traces`/`_bytes` unconditionally, and the new smoke test asserts `native_mem_*` the same way).

Adding a separate emission path in `writeNativeMem()` for a `COUNTERS`-disabled build would be dead code for an unsupported configuration, so leaving as-is. (Note the NativeMem gauges themselves — record/sample/live/max — are not COUNTERS-gated; only the export into the flat counter table is.)

PR #670, opened 2026-07-17
Fix native OOM from unbounded CallTraceHashTable expansion

jbachorik:
## Problem

Under live-heap profiling (`memory=...:l`) with a long-running, unrotated recording, the profiler can exhaust native address space and crash with `os::commit_memory ... failed; error='Not enough space' (errno=12)` / `pthread_create failed (EAGAIN)`. The failure is **`vm.max_map_count` (VMA) exhaustion**, not RAM: a captured memory map showed tens of thousands of 8 MB (and 8 MB-multiple) anonymous regions.

## Root cause

`CallTraceHashTable` grows its backing `LongHashTable` by doubling capacity, allocating each generation from a `LinearAllocator` whose chunks are a fixed 8 MB (`CALL_TRACE_CHUNK`). Once a table's byte size exceeds one chunk (capacity ≥ ~512K entries), `LinearAllocator::alloc()` can never place it: the bump-allocator's inner loop only fits `offs + size <= _chunk_size`, so an oversized request falls through to `getNextChunk()`, which allocates a fresh 8 MB chunk that *also* can't fit it — looping and allocating chunks until `mmap` fails.

The active table only grows this large between `processTraces()` rotations (which reset it to the initial capacity). In a single start-to-end JFR recording no rotation happens, so under the high distinct-trace volume that live-heap sampling sustains, the table crosses the one-chunk threshold and the allocator runs away.

## Fix

`expandTableIfNeeded()` now refuses to expand when the next generation's `getSize()` would exceed `LinearAllocator::maxAllocatableSize()` (one chunk minus its header). Instead of an unbounded allocation loop, `put()` keeps working on the current table at a higher load factor; the next rotation resets it. The bound is derived from the chunk size, so it stays correct if `CALL_TRACE_CHUNK` changes. A new `calltrace_storage_expansion_skipped` counter records when the cap engages.

This also includes a related per-thread native leak fix in `LivenessTracker`: its subsampling RNG `ThreadLocal`s (`gen`/`dis`/`skipped`) were reclaimed only via pthread-key destructors, which never fire when a JNI-attached thread detaches on a reused OS thread. They are now released explicitly from `Profiler::onThreadEnd()`, matching how the CPU/wall engines and `ProfiledThread` are already torn down.

## Validation

Reproducer (`renaissance akka-uct -r 20` under `memory=64:l`) OOM'd at ~90 s / 13 iterations before the fix. With the fix it completes all 20 iterations with peak VMA count in the hundreds (was pinned at the 65530 limit). Compiles clean; verified the change is compatible with the existing `expandTableIfNeeded`/overflow-guard unit tests by inspection (the new check sits alongside the existing guards and only engages for tables that exceed one chunk).

### DIFF
--- a/ddprof-lib/src/main/cpp/callTraceHashTable.cpp
+++ b/ddprof-lib/src/main/cpp/callTraceHashTable.cpp
@@ -46,13 +46,13 @@ private:
   volatile u32 _size;
   u32 _padding2[15];
 
+public:
   static size_t getSize(u32 capacity) {
     size_t size = sizeof(LongHashTable) +
                   (sizeof(u64) + sizeof(CallTraceSample)) * capacity;
     return (size + OS::page_mask) & ~OS::page_mask;
   }
 
-public:
   LongHashTable(LongHashTable *prev = nullptr, u32 capacity = 0,
                 u32 slot_base = 0, bool should_clean = true)
     : _prev(prev), _padding0(nullptr), _capacity(capacity),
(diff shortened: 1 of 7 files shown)


zhengyu123:
Could you rebase the PR? seems it has conflicts.

zhengyu123 (approved):
LGTM

ddprof-lib/src/main/cpp/livenessTracker.h
kaahos: nit: copyrights should be updated to 2026?

ddprof-lib/src/main/cpp/callTraceHashTable.cpp
kaahos: nit: copyrights should be updated

kaahos (approved):
apart from the missing copyrights, it looks good to me!

ddprof-lib/src/main/cpp/livenessTracker.h
jbachorik: @datadog fix this

PR #677, opened 2026-07-21
fix(codecache): make memoryUsage() accurate and live

rkennke:
## What

Makes `CodeCache::memoryUsage()` — which drives the `CODECACHE_NATIVE_SIZE_BYTES` counter — report the profiler's actual native-symbol memory, instead of a stale, heavily-undercounted approximation.

(`CodeCache` here is the profiler's per-native-library **symbol table** used to symbolicate native frames — async-profiler's naming, not the JVM's JIT code cache.)

## Why

The old formula `_capacity * sizeof(CodeBlob*) + _count * sizeof(NativeFunc)` was wrong in several ways:

- **Blob array undercounted ~3×** — counted `sizeof(CodeBlob*)` (a pointer, 8 B) where the array actually holds `CodeBlob` (three pointers, 24 B).
- **Symbol names undercounted** — approximated as a fixed `sizeof(NativeFunc)` (4 B) per symbol, ignoring the name-string length that dominates each allocation (`aligned_alloc(sizeof(NativeFunc) + 1 + strlen(name))`).
- **Missing entirely** — the DWARF unwind table (`_dwarf_table`, can be large).
- **Stale** — `CodeCacheArray` cached the sum at `add()` time via `_used_memory`, so it never reflected a library's later symbol growth.

## How

- `CodeCache::memoryUsage()` recomputes accurately: full blob array (`sizeof(CodeBlob)`), each symbol's real name allocation (new `NativeFunc::allocSize()`), this cache's own name, and the DWARF table.
- The **build-id string is intentionally excluded**: the background library refresher (`Libraries::updateBuildIds`) frees and replaces `_build_id` on already-published caches under `_build_id_lock`, which dump does not hold, so dereferencing it here (`strlen`) would race → use-after-free. It is negligible (~tens of bytes per library) next to the symbol tables, so excluding it keeps this read lock-free at no meaningful accuracy cost.
- `CodeCacheArray::memoryUsage()` sums the live per-library values instead of a cached total. The array is append-only, so iterating the published prefix is safe alongside concurrent `add()`s. The now-dead `_used_memory` field and its `add()`-time accumulation are removed.
- The lock-free read is only valid for caches registered in a `CodeCacheArray` (their `_blobs`/DWARF fields are fixed once published). This is now **enforced**: a `_published` flag is set by `CodeCacheArray::add()`, and `add()`/`expand()`/`setDwarfTable()` assert `!_published` (debug builds). Standalone, continuously-mutated caches like `Libraries::_runtime_stubs` are never published and are exempt.
- `NativeFunc::create()` calls `NativeFunc::allocSize()` so the allocation size and `memoryUsage()`'s accounting share one formula and can't drift.
- `memoryUsage()` is only called at dump time, so the `O(symbols)` recompute is off any hot path.

## Testing

- New `codeCache_ut` (5 tests): per-symbol name length tracked exactly; longer name costs more; empty-cache size is exactly blob-array + own-name (`EXPECT_EQ`); DWARF table term counted; and `CodeCacheArray::memoryUsage()` sums the per-library values (the aggregation `Profiler::dump()` actually consumes).
- Full `:ddprof-lib:gtestDebug` suite green — **354 tests, 0 failures** (rebased on latest `main`).

## Note

Independent of the native-memory accounting work in #669; that PR's `NM_NATIVE_SYMBOLS` gauge mirrors this `memoryUsage()`, so it benefits automatically once both land, but neither depends on the other to build.

### DIFF
--- a/ddprof-lib/src/main/cpp/codeCache.cpp
+++ b/ddprof-lib/src/main/cpp/codeCache.cpp
@@ -116,6 +116,14 @@ void CodeCache::copyFrom(const CodeCache& other) {
   _count = other._count;
   _blobs = new CodeBlob[_capacity];
   memcpy(_blobs, other._blobs, _count * sizeof(CodeBlob));
+  // The memcpy above copied each CodeBlob's _name *pointer*, so without this the
+  // two caches would share name allocations and both destructors would free
+  // them (double-free / use-after-free). Give this copy its own name strings.
+  for (int i = 0; i < _count; i++) {
+    if (_blobs[i]._name != nullptr) {
+      _blobs[i]._name = NativeFunc::create(other._blobs[i]._name, _lib_index);
+    }
+  }
 
   // A copy is a fresh, not-yet-registered cache.
   _published.store(false, std::memory_order_relaxed);
--- a/ddprof-lib/src/main/cpp/codeCache.cpp
+++ b/ddprof-lib/src/main/cpp/codeCache.cpp
@@ -144,23 +144,29 @@ CodeCache::~CodeCache() {
   free(_build_id);  // Free build-id memory
 }
 
-long long CodeCache::memoryUsage() {
+long long CodeCache::memoryUsage() const {
   // The blob array: _capacity entries of CodeBlob.
   long long total = (long long)_capacity * sizeof(CodeBlob);
 
   // This cache's own name, plus each symbol's name string. Each is a
   // variable-length allocation whose size depends on the name length
-  // (see NativeFunc::allocSize).
+  // (see NativeFunc::allocSize). Both the name pointers and the array are fixed
+  // once the library is published (the same read the symbolication fast path
+  // does), so this is safe to read at dump time without extra synchronization.
   total += (long long)NativeFunc::allocSize(_name);
   for (int i = 0; i < _count; i++) {
     total += (long long)NativeFunc::allocSize(_blobs[i]._name);
   }
 
-  // The DWARF unwind table and the build-id string, when present.
+  // The DWARF unwind table, when present (length only — no pointer deref).
   total += (long long)_dwarf_table_length * sizeof(FrameDesc);
-  if (_build_id != nullptr) {
-    total += (long long)strlen(_build_id) + 1;
-  }
+
+  // The build-id string is intentionally NOT counted here: the background
+  // library refresher (Libraries::updateBuildIds) frees and replaces _build_id
+  // on already-published caches under _build_id_lock, which dump does not hold,
+  // so dereferencing it here would race. It is negligible (~tens of bytes per
+  // library) next to the symbol tables, so excluding it costs no meaningful
+  // accuracy while keeping this read lock-free.
 
   return total;
 }
--- a/ddprof-lib/src/main/cpp/codeCache.cpp
+++ b/ddprof-lib/src/main/cpp/codeCache.cpp
@@ -9,13 +9,14 @@
 #include "os.h"
 #include "safeAccess.h"
 
+#include <cassert>
 #include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
 #include <sys/mman.h>
 
 char *NativeFunc::create(const char *name, short lib_index) {
-  size_t size = align_up(sizeof(NativeFunc) + 1 + strlen(name), sizeof(NativeFunc*));
+  size_t size = allocSize(name);
   NativeFunc *f = (NativeFunc *)aligned_alloc(sizeof(NativeFunc*), size);
   f->_lib_index = lib_index;
   f->_mark = 0;
@@ -69,6 +70,8 @@ CodeCache::CodeCache(const char *name, short lib_index,
   _capacity = INITIAL_CODE_CACHE_CAPACITY;
   _count = 0;
   _blobs = new CodeBlob[_capacity];
+
+  _published = false;
 }
 
 void CodeCache::copyFrom(const CodeCache& other) {
@@ -113,6 +116,9 @@ void CodeCache::copyFrom(const CodeCache& other) {
   _count = other._count;
   _blobs = new CodeBlob[_capacity];
   memcpy(_blobs, other._blobs, _count * sizeof(CodeBlob));
+
+  // A copy is a fresh, not-yet-registered cache.
+  _published = false;
 }
 
 CodeCache::CodeCache(const CodeCache &other) {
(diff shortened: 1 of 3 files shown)


rkennke:
Thanks for the reviews — addressed in cf512dde5:

**Build-id data race (codex P2 / Copilot)** — Good catch. `memoryUsage()` no longer reads `_build_id`. The background refresher (`Libraries::updateBuildIds`) calls `setBuildId()` (free + replace) on already-published caches under `_build_id_lock`, which dump doesn't hold, so `strlen(_build_id)` could race that free → use-after-free. It's set at most once per lib and is only ~tens of bytes (a hash), negligible next to the symbol tables, so it's excluded rather than synchronized — keeping the read lock-free. The rest of the sum is safe: blob names are fixed once a library is published (same read the symbolication fast path does), and the DWARF contribution reads only the length scalar (no deref).

**`memoryUsage()` should be `const` (Copilot)** — Done, on both the declaration and definition.

**Use `reinterpret_cast` for int→pointer in tests / drop unused `<cstring>` (Copilot)** — Both done.

Re: the benchmark outliers (`reactors` +7.3%, `fj-kmeans` +4.4% on one commit; `future-genetic` -3.9% the other) — these look like run-to-run noise: they're near/within the reported variance bands and point in opposite directions across commits, and `memoryUsage()` runs only at dump time (off any hot path), so a real few-percent runtime effect isn't plausible from this change.

rkennke:
Re: the PR-description / code mismatch on build-id (Copilot @ codeCache.cpp:169) — went with option (b): updated the PR description so it no longer lists build-id as counted, and added an explicit note that it's intentionally excluded to avoid the `updateBuildIds` free/replace race (matching the header and code comments). Description and implementation now agree.

Also noting the benchmark run for cf512dde5 came back "within expected boundaries — all within run-to-run noise," which lines up with the earlier assessment that the outliers on prior commits were noise.

zhengyu123 (approved):
Looks good overall.

Unrelated to this PR:

Doubled cost at the call site — profiler.cpp:1739-1741 
```
Counters::set(CODECACHE_NATIVE_SIZE_BYTES, native_libs.memoryUsage());
Counters::set(CODECACHE_RUNTIME_STUBS_SIZE_BYTES,
              native_libs.memoryUsage());
```

rkennke:
Thanks for the review!

On the unrelated note (`profiler.cpp:1739-1741` calling `native_libs.memoryUsage()` per counter) — that's already addressed in #669, which factors those lines into a `updateNativeLibMemStats()` helper that reads `memoryUsage()` once and reuses it. (#677 doesn't touch `profiler.cpp`, so nothing to change here.)

Separately, while there: `CODECACHE_RUNTIME_STUBS_SIZE_BYTES` is set to the same `native_libs.memoryUsage()` as `CODECACHE_NATIVE_SIZE_BYTES` — looks like it may be intended to report a distinct quantity (the runtime-stubs cache). I've left the value as-is in #669 to avoid changing that counter's meaning without knowing the intent; happy to follow up separately if it should differ.

rkennke:
Thanks for the thorough Sphinx pass — all seven addressed (`cc839e1f0` and the commit before it):

**MEDIUM**
- **No test for `CodeCacheArray::memoryUsage()`** — added `ArrayMemoryUsageSumsLibraries`: builds an array, `add()`s two populated libs, asserts the aggregate equals the sum of per-lib values.
- **Safety comment over-claims (`_runtime_stubs`)** — scoped it: the lock-free read is valid only for caches registered in a `CodeCacheArray`; `_runtime_stubs` is continuously mutated and unpublished, and must not be read this way without its lock. Reworded the comment accordingly.
- **`allocSize()` duplicates `create()`'s formula** — `create()` now calls `allocSize()`, single source of truth.
- **`CodeCacheArray::memoryUsage()` comment contradicts reality** — reworded; dropped the "symbols added after registration" claim (adds happen pre-publish).
- **Unenforced published-immutability invariant** — added a `_published` flag set by `CodeCacheArray::add()`; `add()`/`expand()`/`setDwarfTable()` now `assert(!_published)`, so a future post-publish mutation fails loudly in debug. `_runtime_stubs` (never published) is exempt.

**LOW**
- **DWARF term untested** — `MemoryUsageCountsDwarfTable` populates a table via `setDwarfTable()` and checks the `length * sizeof(FrameDesc)` contribution.
- **`MemoryUsageCountsFullBlobArray` only lower-bounds** — tightened to exact `EXPECT_EQ` (blob array + own name).

Full `gtestDebug` green (354); the `_published` asserts don't fire anywhere in the suite, confirming the pre-publish invariant holds on exercised paths (JVM integration paths covered by CI). PR description updated to match.

rkennke:
Good catch — fixed in the latest commit. My comments named `Libraries::_runtime_stubs` (which turns out to be an unused/vestigial member); the real continuously-mutated cache is `JitCodeCache::_runtime_stubs`, guarded by `JitCodeCache::_stubs_lock` (`add()` under `lock()`, `findRuntimeStub()` under `lockShared()`). Comments now reference the correct symbol and lock.

ddprof-lib/src/main/cpp/codeCache.cpp
```diff
@@ -157,6 +198,11 @@ void CodeCache::expand() {
 
 void CodeCache::add(const void *start, int length, const char *name,
                     bool update_bounds) {
+  // Symbols are added while parsing a library, before it is registered into a
+  // CodeCacheArray. Adding after publication would race memoryUsage() (which
+  // reads _blobs/_count lock-free at dump time). Unpublished standalone caches
+  // (e.g. _runtime_stubs) are exempt — they are never published.
+  assert(!_published && "add() on a published CodeCache races memoryUsage()");
```
jbachorik: Is `_published` guaranteed observable here? From this [comment](https://github.com/DataDog/java-profiler/pull/677/changes#diff-255a19b5d713308003b48a9a9b54af308f4c09d68bd1671379eb2bea6a52bdfeR327) it seems it needs to deal with concurrency.

rkennke:
Good catch — made `_published` a `std::atomic<bool>` (release store in `markPublished()`, acquire loads in the asserts) in the latest commit. It's set sequenced-before `CodeCacheArray::add()`'s RELEASE publish of the pointer, so any thread that reaches a published cache (via the array's ACQUIRE load) observes `_published == true` — now with a proper happens-before and no formal data race (TSan-clean), rather than relying on a plain-bool read.

rkennke:
Fixed in the latest commit — `copyFrom()` now deep-copies each blob's name string (`NativeFunc::create()`), matching how build-id and the DWARF table were already deep-copied, so a copied `CodeCache` no longer shares — and double-frees — the originals' name allocations. Added `CopyIsDeepAndDoesNotDoubleFree` (copies a populated cache, destroys both) which aborts under the old shallow behavior. (Note `CodeCache` is only ever used by pointer today, so this path was unexercised — the fix is hardening; deleting the copy ops was the alternative, but deep-copy keeps the change surgical.)

jbachorik (approved):

PR #683, opened 2026-07-22
chore: address release script

kaahos:
**What does this PR do?**:

fixes `check_not_stuck` so that a missing release tag is treated as the expected state for a new release.

The check now uses `git show-ref --verify --quiet` with the complete tag reference and returns success explicitly when the tag does not exist or the release is a dry run.

**Motivation**:

With `set -e`, the previous `git rev-parse ... || return` propagated exit code 128 when the target tag did not exist. This aborted valid releases before tag creation.

**Additional Notes**:

Existing tags still trigger the intended “branch is stuck” failure. Using the complete `refs/tags/...` reference also ensures that only the exact tag is checked.

An end-to-end release dry run can only be executed after merging because minor-release workflows require remote `main`.

**How to test the change?**:

- `bash -n .github/scripts/release.sh`
- `git diff --check origin/main...HEAD`
- Verified `check_not_stuck` behavior:
  - Missing tag returns 0.
  - Existing tag returns 1 with the expected diagnostic.
  - Dry-run returns 0.

**For Datadog employees**:
- [ ] JIRA: [JIRA-XXXX]

### DIFF
--- a/.github/scripts/release.sh
+++ b/.github/scripts/release.sh
@@ -20,8 +20,13 @@ check_not_stuck() {
   local base=$1
   local branch=$2
 
-  [ -n "$DRYRUN" ] && return
-  git rev-parse "v_${base}" >/dev/null 2>&1 || return
+  if [ -n "$DRYRUN" ]; then
+    return 0
+  fi
+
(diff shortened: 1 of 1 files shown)


jbachorik (approved):
Thanks!

PR #696, opened 2026-07-29
Remove legacy DirectByteBuffer-based OTEP context storage (Phase 3)

rkennke:
## Summary

- Phase 3 of the OTEP context-storage migration: dd-trace-java PR #11899 has landed and migrated onto the all-native context write API (`setTraceContext`/`clearTraceContext`/`setContextValue`/`clearContextValue`), so java-profiler was the last consumer of the legacy DirectByteBuffer-based path (`ThreadContext`, `OtelContextStorage`, `ContextStorageMode`, `ContextSetter`, `ScopeStack`), which carried a virtual-thread use-after-free risk.
- Deletes the DBB conduit classes and their JNI backing (`initializeContextTLS0`), removes deprecated `JavaProfiler` APIs (`setContext`, `clearContext`, `setContextAttribute`, `getThreadContext`, `contextStorageMode`, etc.), and adds package-private test-only native read accessors (`testGetSpanId`, `testGetRootSpanId`, `testReadTraceId`, `testReadContextAttribute`, `testIsContextValid`) as the read oracle for tests.
- Ports all Java tests, JMH benchmarks, and the chaos harness onto the all-native API; replaces `ReapplyContextAntagonist`/`VirtualThreadContextCascadeAntagonist` (whose premise was the DBB carrier-cache UAF) with a new `ReapplyContextValueAntagonist` exercising equivalent contention against the all-native record.
- Adds `ContextValueCacheTest` to preserve the Dictionary-exhaustion/hash-collision coverage previously carried by the deleted `TagContextTest`.
- Rewrites `doc/architecture/TLSContext.md` and deletes the now-superseded phase 1/2 design docs.

## Test plan

- [x] `ddprof-lib` native (C++) and Java compile cleanly
- [x] `ddprof-test-tracer`, `ddprof-stresstest` (main/jmh/chaos source sets) compile cleanly
- [x] Targeted context tests pass: `AllNativeContextTest`, `AllNativeContextSamplingTest`, `ContextValueCacheTest`, `MaxContextSlotsTest`, `PrecheckTest`, `WallclockMitigationsCombinedTest`, `ThreadFilterSmokeTest`, `QueueTimeTest`, `ContendedWallclockSamplesTest`, `WriteStackTracesAfterClassUnloadTest`
- [x] Full `ddprof-test:testDebug` suite: 236 tests, 4 failures — all confirmed pre-existing/unrelated (thread-churn lifecycle timeouts, a flaky retried vtable-frame test, a wall-clock timing assertion), none touching context/tracing code
- [x] `ContextCombinedBenchmark` and `TracedParallelWork` JMH benchmarks execute without exceptions against the all-native API
- [x] Grep sweep confirms no dangling references to `ThreadContext`, `OtelContextStorage`, `ContextStorageMode`, `ContextSetter`, `ScopeStack` outside history


ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -1088,3 +1035,87 @@ Java_com_datadoghq_profiler_JavaProfiler_dumpContext(JNIEnv* env, jclass unused)
...
+Java_com_datadoghq_profiler_JavaProfiler_testGetSpanId0(JNIEnv* env, jclass unused) {
+  ProfiledThread* thrd = ProfiledThread::current();
+  if (thrd == nullptr) {
+    return 0;
+  }
+  OtelThreadContextRecord* record = thrd->getOtelContextRecord();
+  uint64_t beSpan;
+  memcpy(&beSpan, record->span_id, 8);
+  return (jlong)__builtin_bswap64(beSpan);
+}
```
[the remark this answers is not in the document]
rkennke: Fair point in general, but current usage is self-read-only: all callers (AllNativeContextTest) call these test-only accessors synchronously on the same thread immediately after the corresponding set/clear, with no concurrent writer ever in play. So there's no live bug today, and these aren't reachable from any non-test code path. If a future test ever reads another thread's record concurrently, we should gate on `valid` like the rest of the publication protocol at that point — leaving as-is for now.

ddprof-stresstest/src/chaos/java/com/datadoghq/profiler/chaos/Main.java
```diff
@@ -94,8 +92,8 @@ private static Antagonist create(String name) {
                 return new WeakRefWaveAntagonist();
             case "dump-storm":
                 return new DumpStormAntagonist();
-            case "reapply-context":
-                return new ReapplyContextAntagonist();
+            case "reapply-context-value":
```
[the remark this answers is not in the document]
rkennke: Confirmed — vthread-context-cascade targeted a use-after-free specific to the removed per-carrier-cached DBB conduit, which has no equivalent in the all-native design (each JNI call resolves the context record fresh via ProfiledThread::current(), so there's nothing left to go stale). No native-path equivalent exists to swap in, so I just dropped the removed antagonist name from run-chaos-harness.sh's profiler+tracer default list (and the stale comment referencing it). Fixed in 25e3f5ada.

ddprof-stresstest/src/chaos/java/com/datadoghq/profiler/chaos/ReapplyContextValueAntagonist.java
```diff
@@ -0,0 +1,114 @@
...
+                profiler.setContextValue(i, ROUTES[i % ROUTES.length]);
+            }
+            for (int i = 0; i < SLOT_COUNT; i += 2) {
+                profiler.clearContextValue(i);
+            }
+            for (int i = 0; i < SLOT_COUNT; i += 2) {
+                profiler.setContextValue(i, ROUTES[(i + ThreadLocalRandom.current().nextInt(ROUTES.length)) % ROUTES.length]);
+            }
+            profiler.clearTraceContext();
+        }
```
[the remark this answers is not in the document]
rkennke: Fixed: workerLoop now checks setContextValue's return and halts the process (Runtime.getRuntime().halt(1)) on false, rather than ignoring it. Note the first cut of this fix (throwing IllegalStateException) turned out to be insufficient on its own since it was thrown from a task submitted via pool.submit(...) with the Future discarded, so the exception would have died silently. Switched to execute() and an explicit halt so a failure actually produces the non-zero exit code that's this harness's only failure signal (per Main's class javadoc). Fixed in 25e3f5ada and e0b0e5bb4.

ddprof-test-tracer/src/main/java/com/datadoghq/profiler/context/ContextExecutor.java
```diff
@@ -25,11 +25,6 @@ protected <T> RunnableFuture<T> newTaskFor(Runnable runnable, T value) {
     protected void beforeExecute(Thread t, Runnable r) {
         super.beforeExecute(t, r);
         profiler.addThread();
```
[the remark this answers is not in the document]
rkennke: Good catch, confirmed as a real regression — removing the priming `setContext(0, 0)` call also removed its incidental side effect of clearing the previous task's context before `addThread()` re-enables wall-clock sampling. Fixed in 6fd6733f6 by calling `clearTraceContext()` before `addThread()` in `beforeExecute`.

ddprof-test/src/test/java/com/datadoghq/profiler/ContextValueCacheTest.java
```diff
@@ -0,0 +1,115 @@
...
+        assertNotNull(first);
+        ContextValueCache.Entry second = cache.resolve("hello");
+        assertSame(first, second, "an unchanged value must hit the cache, not re-register");
+    }
+
+    @Test
+    public void oversizedValueResolvesToNull() throws IOException {
+        loadLibrary();
+        char[] chars = new char[ContextValueCache.MAX_VALUE_BYTES + 1];
+        java.util.Arrays.fill(chars, 'x');
```
jbachorik: Nit: Would prefer importing`Arrays` instead of FQN

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
```diff
@@ -250,54 +223,10 @@ public void removeThread() {
...
-    }
-
     // ---- All-native context write API (OTEP #4947) --------------------------------------------
-    // Provisional names (subject to dd-trace-java coordination). These resolve the current carrier's
-    // OTEP record inside a single JNI call per operation — no cached DirectByteBuffer, so they are
-    // race-free under virtual-thread migration (see the design note and setTraceContext0 et al.).
-    // They coexist with the deprecated DirectByteBuffer path below; both write the same native
-    // record, and no thread uses both at once.
+    // Each of these resolves the current carrier's OTEP record inside a single JNI call per
+    // operation — no cached DirectByteBuffer, so they are race-free under virtual-thread migration
```
jbachorik: We might drop the references to the previous, no longer existing functionality. The migration is over and from now on only the new approach will be available.

ddprof-lib/src/main/java/com/datadoghq/profiler/OTelContext.java
```diff
@@ -235,7 +235,7 @@ public ProcessContext readProcessContext() {
      *                      element is null, the entire process context publish is
      *                      skipped - no context is published (a warning is logged) and
      *                      no exception is thrown. Order must match the indices used
-     *                      with {@link ThreadContext#setContextAttribute(int, String)}.
+     *                      with {@link JavaProfiler#setContextValue(int, CharSequence)}.
```
jbachorik: I noticed this only now - is there a particular reason why `CharSequence` and not string? The method will call `toString()` on the CharSequence anyway.

jbachorik (approved):
I have a pair of nits and one question, but nothing blocking.
Thanks!

ddprof-test/src/test/java/com/datadoghq/profiler/ContextValueCacheTest.java
rkennke: Fixed: switched to `import java.util.Arrays` in d3dc0a23f.

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
rkennke: Fair — the DBB references were leftover from before the removal and no longer apply. Trimmed the comment that mentioned it in d3dc0a23f.

ddprof-lib/src/main/java/com/datadoghq/profiler/OTelContext.java
rkennke: No particular reason — resolveContextValue immediately calls value.toString() before touching the cache either way, so CharSequence bought nothing. Switched setTraceContext/setContextValue to String in d3dc0a23f.

ddprof-lib/src/main/cpp/javaApi.cpp
```diff
@@ -1088,3 +1035,87 @@ Java_com_datadoghq_profiler_JavaProfiler_dumpContext(JNIEnv* env, jclass unused)
   ContextApi::get(spanId, rootSpanId);
   TEST_LOG("===> Context: tid:%lu, spanId=%lu, rootSpanId=%lu", OS::threadId(), spanId, rootSpanId);
 }
+
+// ---- Test-only reads of the current thread's OTEP record -----------------------------------
+// Each reads the current carrier's record directly via ProfiledThread::current(), with no
+// detach/attach (diagnostic-only, not on any signal-handler or hot write path).
+
+extern "C" DLLEXPORT jlong JNICALL
+Java_com_datadoghq_profiler_JavaProfiler_testGetSpanId0(JNIEnv* env, jclass unused) {
```
jbachorik, line 1044: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** testGetSpanId0, testGetRootSpanId0, testReadTraceId0, and testReadContextAttribute0 read OtelThreadContextRecord fields directly with no check of record->valid, while the sibling testIsContextValid0 (and every production reader in wallClock.cpp / ContextApi::get) gates on an acquire-load of valid before trusting the record's contents. A reader that races a write's detach/attach window on these accessors can observe a torn or half-updated record.

**Suggestion:** Have these test accessors check record->valid (with an acquire load, mirroring testIsContextValid0) before reading fields, or document explicitly that callers must serialize with testIsContextValid0 themselves before invoking them.

rkennke: Same question as Copilot raised earlier on this file (see my reply at https://github.com/DataDog/java-profiler/pull/696#discussion_r3673862365) — these are test-only accessors that AllNativeContextTest calls synchronously on the same thread immediately after the corresponding set/clear, with no concurrent writer ever in play, so there's no live bug today and they're not reachable from any non-test code path. Happy to add a doc comment noting the single-threaded-caller assumption if that's preferred, but didn't want to add an acquire-load check that has nothing to guard against in current usage.

zhengyu123:
1. Stale/dead carrier-scoping build logic left behind (build-logic, not touched by this PR)
build-logic/conventions/src/main/kotlin/com/datadoghq/profiler/ProfilerTestPlugin.kt still contains carrierExportJvmArgs() (and two call sites) which adds --add-exports=java.base/jdk.internal.misc=ALL-UNNAMED on JDK 21+ test JVMs, plus ~20 lines of doc comments referencing OtelContextStorage.Mode.CARRIER and "carrier-scoping tests." That flag existed solely to resolve jdk.internal.misc.CarrierThreadLocal for OtelContextStorage's carrier-scoped mode — and the only tests that exercised it, CarrierContextStorageTest and OtelContextStorageModeTest, are deleted by this PR. A repo-wide grep confirms CarrierThreadLocal/jdk.internal.misc now appears only in this dead build-logic path, ContextCombinedBenchmark's JMH jvmArgsAppend (also now pointless — that benchmark's carrier/thread @Param and contextStorageMode() call were correctly removed, but the --add-exports annotation was left behind), and utils/run-chaos-harness.sh's --add-exports line. None of these are functionally broken (an unused --add-exports flag is harmless), but it directly contradicts the PR description's claim of a "grep sweep confirms no dangling references... outside history" — the sweep evidently didn't cover build-logic/ or these two leftover flag sites. Worth a follow-up cleanup pass.

2. Test-coverage gap: custom context attributes → JFR field pipeline is no longer exercised end-to-end
The deleted TagContextTest.test() was a substantial integration test: it drove ContextSetter/setContextAttribute, then verified (a) per-tag-value sample weight distribution in real datadog.MethodSample JFR events, (b) that jdk.ActiveSetting events correctly enumerate the configured contextattribute names, and (c) that the dictionary_context_keys JFR counter matches. The PR's replacement, ContextValueCacheTest, intentionally only ports the Dictionary-exhaustion/hash-collision unit coverage (as the PR description states), and AllNativeContextSamplingTest only verifies span/root-span propagation into JFR — not custom attribute values. I couldn't find any other test in the diff or existing suite that verifies a value written via the new setContextValue/ContextSetter.offsetOf path actually surfaces as a correctly-named, correctly-weighted JFR field. This is a real (if non-blocking) reduction in end-to-end coverage for a code path that customers actually consume (custom span tags in profiles) — worth either an explicit acknowledgment that it's covered elsewhere, or a follow-up test.

3. reapply-context-value antagonist isn't wired into the default chaos harness run
The new ReapplyContextValueAntagonist — introduced specifically to preserve UAF-risk stress coverage for the record this PR touches — is correctly registered in Main.java's factory and documented in the chaos README.md, but utils/run-chaos-harness.sh's DEFAULT_ANTAGONISTS list for the profiler+tracer config was not updated to include it (the old reapply-context wasn't in that list either, so this isn't a regression, but given the PR's stated purpose — proving the UAF class of bug is gone — it's a missed opportunity to make that proof run by default rather than only via ANTAGONISTS_OVERRIDE).

4. Minor: ContextExecutor.beforeExecute rationale changed without full clarity on native semantics
The old comment justified priming context via "avoid race condition with wall clock signals" against lazily-initialized TLS; the new comment reframes it as clearing stale context from thread-pool reuse. That reframing is accurate and the fix is sound — but note that clearTraceContext0 (unlike setTraceContext0) never calls ContextApi::initializeContextTLS, so it does not actually set isContextInitialized the way the old DBB-priming path did. I traced through ContextApi::get/wallClock.cpp's hasKnownActiveTraceContext and confirmed both gate on that flag and both correctly report "no context" either way (the record is zeroed regardless), so this isn't a functional bug — just worth double-checking against the original race this code was written to avoid, since the mechanism providing the guarantee changed even though the observable behavior didn't.

rkennke:
Thanks for the thorough sweep — addressed in 0dfc51a34:

1. **Dead carrier-scoping build logic**: confirmed `CarrierThreadLocal`/`jdk.internal.misc` had no other referents. Removed `carrierExportJvmArgs()`/`testJvmMajorVersionFromRelease()` and both call sites from `ProfilerTestPlugin.kt`, the now-pointless `jvmArgsAppend` on `ContextCombinedBenchmark`, and the `--add-exports` line in `run-chaos-harness.sh`. Also generalized the stale doc comment in `ConfigurationPresets` that specifically called out carrier-scoping as its example.

2. **Test-coverage gap**: added `CustomContextAttributeSamplingTest`, porting the weight-distribution / `jdk.ActiveSetting` / `dictionary_context_keys` assertions from the deleted `TagContextTest.test()` onto the all-native `setContextValue` API. Confirms a value written through `setContextValue`/`ContextSetter` surfaces as a correctly-named, correctly-weighted `datadog.MethodSample` field. Ran it (and the rest of the `context` package) locally — all green.

3. **reapply-context-value not defaulted**: added it to both `DEFAULT_ANTAGONISTS` lists (it calls `JavaProfiler` directly, no tracer required, so it's not `profiler+tracer`-only like `trace-context`). It now runs by default rather than only via `ANTAGONISTS_OVERRIDE`.

4. **ContextExecutor semantics**: traced through it — you're right that `clearTraceContext0` never calls `initializeContextTLS`, unlike `setTraceContext0`/`setContextValue0`. But `hasKnownActiveTraceContext` (wallClock.cpp) doesn't gate on `valid`; it reads `span_id` directly after the `isContextInitialized` check, and `clearTraceContext0` always zeroes `span_id`. So for a reused worker (already initialized from its previous task) the span_id==0 check correctly reports "no context", and for a never-initialized thread the `isContextInitialized` check does the same. Observable behavior is correct in both cases — no code change, just wanted to confirm the trace-through explicitly since you flagged the mechanism change.

jbachorik (approved):
Looking good! Thanks for fixing all this!

ddprof-lib/src/main/java/com/datadoghq/profiler/JavaProfiler.java
```diff
@@ -320,7 +248,7 @@ public void clearContext() {
      *         non-negative {@code slotN} is {@code >= MAX_CONTEXT_SLOTS} (out of range)
      */
     public void setTraceContext(long rootSpanId, long spanId, long traceIdHigh, long traceIdLow,
-                                int slot0, CharSequence v0, int slot1, CharSequence v1) {
+                                int slot0, String v0, int slot1, String v1) {
         if (spanId == 0) {
```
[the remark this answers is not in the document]
jbachorik: All consumers are under our control. This is not important.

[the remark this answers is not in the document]
rkennke: Good catch — reverted in f80a9b387. dd-trace-java's `DatadogProfiler.setTraceContext` forwards `ProfilerContext#getOperationName()`/`getResourceName()` directly into this method, and both are declared `CharSequence` throughout `DDSpanContext`'s public span API (backed by allocation-free implementations like `UTF8BytesString`/`SubSequence` on the tracing hot path, not always `java.lang.String`). Narrowing to `String` would break that call site at compile time on the next ddprof-lib version bump.

`setContextValue` keeps its `String` param — dd-trace-java always passes an actual `String` there, so no compatibility concern on that one.

ddprof-lib/src/main/java/com/datadoghq/profiler/OTelContext.java
rkennke: Correction to my earlier reply here: there *is* a good reason after all, and I got it wrong. Copilot flagged (https://github.com/DataDog/java-profiler/pull/696#discussion_r3685390991) that dd-trace-java's `DatadogProfiler.setTraceContext` forwards `ProfilerContext#getOperationName()`/`getResourceName()` directly into this method, and those are declared `CharSequence` throughout `DDSpanContext`'s public span API — backed by allocation-free implementations (`UTF8BytesString`, `SubSequence`) specifically to avoid materializing a `String` on the tracing hot path. Narrowing to `String` would've broken that call site at compile time on the next ddprof-lib version bump, and runs counter to the tracer's own allocation-avoidance idiom for these fields.

Reverted `setTraceContext`'s `v0`/`v1` back to `CharSequence` in f80a9b387. Since the whole point of accepting `CharSequence` there is avoiding an eager `String`, I also made `ContextValueCache` actually honor that: it now has a `resolve(CharSequence)` overload that never calls `toString()` on a cache hit (the common case) — a small `contentHashCode()` replicates `String#hashCode()`'s documented polynomial algorithm so a `CharSequence` and a content-equal cached `String` key land in the same slot, and `String#contentEquals(CharSequence)` compares without allocating. `toString()` is only paid on a genuine miss, where `registerConstant0` needs a real `String` anyway. Added `charSequenceHitNeverCallsToString`/`charSequenceMissMaterializesStringExactlyOnce` to `ContextValueCacheTest` to pin that behavior.

`setContextValue` keeps its `String` param — dd-trace-java always passes an actual `String` there, so `String` is the honest, non-lossy type for that one.

ddprof-lib/src/main/java/com/datadoghq/profiler/ContextValueCache.java
```diff
@@ -98,6 +97,45 @@ Entry resolve(String value) {
...
+     * setTraceContext}) whose value may be a non-{@code String} implementation — dd-trace-java
+     * passes {@code UTF8BytesString}/{@code SubSequence} spans-name views specifically to avoid
+     * forcing a {@code String} allocation on the hot path. The cache-hit path here never calls
+     * {@code toString()}: {@link #contentHashCode} replicates {@link String#hashCode()}'s
+     * algorithm so a {@code CharSequence} and a content-equal cached {@code String} key land in
+     * the same slot, and {@link String#contentEquals(CharSequence)} compares without allocating.
+     * {@code toString()} is only paid on a genuine miss, where a {@code String} is needed anyway
+     * to hand to {@code registerConstant0} and to store as the entry's key.
+     */
+    Entry resolve(CharSequence value) {
```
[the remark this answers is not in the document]
rkennke: Applied, in 83c6d4ce8 — resolve(String) is now a private resolveString helper; resolve(CharSequence) is the sole entrypoint, with the existing instanceof String fast-path preserved so behavior and performance are unchanged.

PR #720, opened 2026-08-06
fix: don't disable profiler on Alpine/musl over missing libgcc_s.so.1

jbachorik:
**What does this PR do?**:
`checkState()` was failing profiler startup on any host missing `libgcc_s.so.1`, including musl/Alpine — but that failure mode (glibc's `pthread_exit`/`pthread_cancel` aborting via `__libc_fatal` when its private lazy `dlopen` of `libgcc_s` fails) is glibc-specific. musl never takes that code path, so a missing `libgcc_s.so.1` there was never actually dangerous. The check now only treats `prewarmUnwinder()` failure as fatal off musl (`!prewarmUnwinder() && !OS::isMusl()`), so Alpine/musl hosts without `libgcc_s.so.1` start normally.

Also rewrote the `prewarmUnwinder()` comment, which had drifted to attribute the eager load to only one of the two lazy-load issues it actually closes (J9's DWARF-unwinder signal-context dlopen, and the glibc pthread_exit path above) — it now documents both, and why only one of them gates fatality.

**Motivation**:
Distroless/hardened/Alpine images commonly ship without `libgcc_s.so.1`. The overly broad check meant the profiler refused to start on those musl hosts even though nothing there was actually at risk.

**Additional Notes**:
Also fixes `faultInjection_ut.cpp`'s `CheckStateSurfacesInjectedPrewarmUnwinderFailure` test, which assumed an injected `prewarmUnwinder()` failure always surfaces as `"Missing libgcc_s.so.1"` — no longer true on musl. Latent until now since `-PenableFaultInjection` isn't wired into CI yet, but fixed ahead of that.

**How to test the change?**:
- [x] Updated `faultInjection_ut.cpp` to assert the correct behavior on both glibc and musl.
- [x] No behavior change on glibc hosts (still fails fast on missing `libgcc_s.so.1`); musl now starts normally instead of refusing to start.

**For Datadog employees**:
- [x] JIRA: [PROF-15664]

[PROF-15664]: https://datadoghq.atlassian.net/browse/PROF-15664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/profiler.cpp
+++ b/ddprof-lib/src/main/cpp/profiler.cpp
@@ -845,12 +845,37 @@ void Profiler::writeHeapUsage(long value, bool live) {
 
 bool Profiler::prewarmUnwinder() {
 #ifdef __linux__
-  // J9 on aarch64 (and other JVMs) lazily loads libgcc_s.so.1 from its DWARF
-  // unwinder during stack walks. When that happens inside a signal handler
-  // frame, our dlopen_hook fires from signal context and tries to refresh the
-  // library list — Mutex::lock and malloc on a signal stack.  By forcing the
-  // load here, before any signal handler is installed, subsequent calls find
-  // libgcc_s already mapped and the lazy-load path never runs.
+  // Force libgcc_s.so.1 to load now and report whether that succeeded. This
+  // closes two separate lazy-load landmines that happen to share the same
+  // library and the same fix:
+  //
+  // 1. J9's DWARF unwinder can lazily dlopen libgcc_s.so.1 while walking a
+  //    stack (e.g. on aarch64). Our sampling stack walks run from a signal
+  //    handler, so that dlopen call can fire our PLT-patched dlopen_hook()
+  //    from signal context, where it does Mutex::lock/malloc — both
+  //    AS-unsafe. Other JVM-internal lazy loads (observed under Graal on
+  //    aarch64) can trigger the same path. Forcing the load here, before any
+  //    signal handler is installed, means the JVM's later resolve finds
+  //    libgcc_s already mapped and this lazy-load path never runs.
+  //
+  // 2. Separately, glibc's pthread_exit()/pthread_cancel() call
+  //    __pthread_unwind(), which invokes _Unwind_ForcedUnwind()
+  //    unconditionally. glibc does not hard-link libgcc_s into
+  //    libc/libpthread; it lazily resolves that symbol via its own private
+  //    __libc_dlopen(LIBGCC_S_SO) the first time a thread exits or is
+  //    cancelled — including our own worker/sampler threads, whose
+  //    start_routine_wrapper (libraryPatcher_linux.cpp) calls pthread_exit().
+  //    If that lazy load fails, glibc itself calls __libc_fatal("libgcc_s.
+  //    so.1 must be installed for pthread_exit to work\n"), aborting the
+  //    whole process — observed in production on hardened/distroless images
+  //    that ship without libgcc_s.so.1.
+  //    __libc_dlopen is glibc's private loader, not the public PLT-visible
+  //    dlopen, so unlike (1) this does not route through dlopen_hook().
+  //
+  // Whether a load failure here is actually fatal is decided by the caller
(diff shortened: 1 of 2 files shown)


jbachorik:
@codex review

ddprof-lib/src/test/cpp/faultInjection_ut.cpp
```diff
@@ -271,19 +278,26 @@ TEST_F(FaultInjectionTest, CheckStateSurfacesInjectedPrewarmUnwinderFailure) {
...
       // prewarmUnwinder() succeeded (non-injected, ~99% of calls) and fell
-      // through to the mocked JVMSupport::initialize() failure instead.
+      // through to the mocked JVMSupport::initialize() failure instead --
+      // or, on musl, an injected prewarmUnwinder() failure did the same
+      // because checkState() doesn't treat it as fatal there.
       EXPECT_STREQ("Profiler encountered fatal error", error.message());
+      if (!prewarmFailureIsFatal) {
+        sawInjectedFailure = true;
+      }
       sawNonInjectedPrewarm = true;
```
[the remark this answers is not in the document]
jbachorik: Fixed. The loop now reads the `FAULTS_INJECTED` counter before and after each `checkState()` call and uses the delta to decide whether that iteration actually took the injected `prewarmUnwinder()` path, instead of inferring it from the mocked error string (which is identical for both outcomes on musl). `sawInjectedFailure` and `sawNonInjectedPrewarm` are now set independently from that counter-derived boolean, so the test genuinely proves both the injected failure and the musl fall-through were exercised.

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -1325,9 +1350,12 @@ Error Profiler::checkState() {
...
   } else if (s == NEW) {
-    // Force libgcc_s to load now (idempotent dlopen) so the JVM's DWARF
-    // unwinder cannot lazy-load it later from signal context.
-    if (!prewarmUnwinder()) {
+    // prewarmUnwinder() closes a glibc-specific pthread_exit/pthread_cancel
+    // landmine (see its comment) by force-loading libgcc_s.so.1 up front.
+    // A failure only matters on glibc, where that landmine is real; musl
+    // never hits the code path that needs libgcc_s, so a missing library
+    // there is not a reason to fail profiler startup.
+    if (!prewarmUnwinder() && !OS::isMusl()) {
```
zhengyu123, line 1358: I would move `OS::isMusl()` check into `prewarmUnwinder`, where it already performs platform check, e.g. __linux__ guard.

zhengyu123 (changes requested):

ddprof-lib/src/main/cpp/profiler.cpp
jbachorik: But we want to prewarm regardless of whether this is musl or not, right? The original bug for prewarming was for J9 and was not limited to musl. We just want to skip the 'fail fast' for musl because the `pthread_exit` on musl has different implementation and does not need calls from libgcc_s

zhengyu123: My point is that, if musl does not need libgcc_s, why attempt to load it?

jbachorik: J9 might need it even on musl. This actually made me wonder if we should also guard on `isJ9()` here. Unless the J9 unwinder can gracefully ignore the missing library ...
Hm, will have to try it. I will merge this one to have a quick fix and see how it works with J9.

zhengyu123: Okay

zhengyu123 (approved):

ddprof-lib/src/main/cpp/profiler.cpp
jbachorik: Just FYI, J9, apparently, comes with libgcc_s hardlinking, so without that library JVM does not even start.

PR #723, opened 2026-08-07
Remove dead Lookup::_class_cache

rkennke:
## Summary
- `Lookup::_class_cache` (a `std::map<u32, const char*>`) and its populator `initClassCache()` are write-only: a full-repo search finds no reader anywhere.
- The call site's own comment claims it "seeds vtable-receiver class names for resolveMethod(BCI_ALLOC)", but that path actually resolves through the separate `_vtable_receiver_cache` / `resolveVTableReceiverCached()` — this map plays no role in it. Looks like leftover from an earlier version of the vtable-receiver feature that was superseded.
- Every JFR chunk dump was paying to snapshot the entire class table into this map just to discard it.

## Test plan
- [x] `./gradlew :ddprof-lib:compileRelease` — all 66 C++ sources compile clean
- [x] `./gradlew :ddprof-lib:gtestRelease_flightRecorder_result_ut :ddprof-lib:gtestRelease_lineNumberTableCopy_ut :ddprof-lib:gtestRelease_methodInfo_hash_ut :ddprof-lib:gtestRelease_methodMapId_ut` — all pass
- [x] repo-wide grep for `_class_cache`/`initClassCache` confirms no remaining references

### DIFF
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -672,22 +672,6 @@ MethodInfo *Lookup::resolveMethod(ASGCT_CallFrame &frame) {
   return mi;
 }
 
-void Lookup::initClassCache() {
-  // Snapshot _classes into _class_cache for use by resolveMethod(BCI_ALLOC).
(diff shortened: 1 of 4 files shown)


kaahos (approved):
looks good to me, thanks!

PR #733, opened 2026-08-11
Return JMETHODID_NOT_WALKABLE instead of nullptr for unpopulated jmethodID slots

jbachorik:
**What does this PR do?**:
Guards the raw `Method*` fallback in `fillJavaFrame` on `fjmethodid=false`, instead of patching `VMMethod::id()` return values to distinguish unprimed from post-invalidation cache states. `id()` now returns `nullptr` for both `ids == NULL` (unprimed klass) and `num >= len` (post-invalidation cache shrink); `fillJavaFrame` decides whether the raw `Method*` fallback is appropriate based on the configuration.

**Motivation**:
`VMMethod::id()` returned `nullptr` when a klass's jmethodID cache was not yet populated. `fillJavaFrame` treats `nullptr` as "store raw `Method*` pointer", deferring a dereference of that pointer to dump time (up to ~60s later per JFR chunk length). If HotSpot's redefinition cleanup reclaimed the underlying `Method*`/`Symbol*` metadata in that window, the unguarded memcpys in `HotspotSupport::resolve()` faulted on freed memory.

This fires on every `RedefineClasses`/`RetransformClasses` (routine, continuous — e.g. dd-trace-java agent retransformation): the cache is briefly empty between invalidation and repopulation, and a thread sampled in that window hits the `nullptr` path.

**Chain**:
1. Agent retransforms a class → jmethodID cache empty
2. Concurrently-sampled frame → `id()` returns `nullptr` → `fillFrameRaw` stores raw `Method*` + `isRawPointer` flag
3. Dump time → `resolveMethod` sees `isRawPointer(bci)` → calls `HotspotSupport::resolve()` → unguarded memcpys fault on reclaimed metadata

**Fix**: Move the safety boundary from `id()` (which cannot know the `fjmethodid` mode) to `fillJavaFrame` (which can):
- `fjmethodid=true`: all jmethodIDs should be preloaded, so `nullptr` from `id()` is a transient invalidation gap or missed preload. Use the sentinel (`JMETHODID_NOT_WALKABLE`) rather than deferring a raw `Method*` dereference to the dump thread.
- `fjmethodid=false`: the user explicitly opted into the raw `Method*` path. `nullptr` — whether from unprimed klass or post-invalidation — uses the raw `Method*` fallback as designed. The deferred-dereference race is inherent to this mode.
- `JMETHODID_NOT_WALKABLE` (bad pointer chain / faulted load): always sentinel. The `Method*` itself may be bogus, so raw fallback gains nothing.

This eliminates the need to classify every `nullptr`-producing return path in `id()` individually — the configuration check in `fillJavaFrame` handles all cases uniformly.

**Additional Notes**:
- `fillFrameRaw` and `isRawPointer` left in place as the `fjmethodid=false` path
- `HotspotSupport::resolve()` still returns `nullptr` as the dump-time "unknown" signal — that contract with `flightRecorder.cpp` is unchanged
- OpenJ9/Zing paths untouched — they don't use `fillJavaFrame`/`fillFrameRaw`/`isRawPointer`

**How to test the change?**:
- `HotspotMethodIdTest` — 5 gtests covering all `id()` cache-state branches (unprimed, shrunk, empty slot, populated, rejected) — all pass
- `JMethodIDInvalidationStressTest` — passes (integration, slow suite)
- All gtest suites pass; spotlessApply clean

**For Datadog employees**:
- [x] JIRA: [PROF-15568]

[PROF-15568]: https://datadoghq.atlassian.net/browse/PROF-15568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbachorik:
@codex review

zhengyu123:
Sorry, I am not convinced - for an invalid jmethod ID, jvmti->GetLinenumberTable() should fail, or caller owns the table (https://github.com/openjdk/jdk/blob/master/src/hotspot/share/prims/jvmtiEnv.cpp#L3287) if succeeded. In fact, I don't think `memcpy()` is necessary at all.

I think https://datadoghq.atlassian.net/browse/PROF-15645 is better protection for this crash, before we can find the root cause.

jbachorik:
@zhengyu123 The debugger pinpointed the failing instruction within the code that is extracting data from `NMethod*` - we MUST NOT DO THAT when `force-jmethodids` is not set to false. This is a correctness fix that avoids falling into code which should not be executing with the current setup. Adding a longjump around it would be a bandaid. We can do that if this fix proves to be ineffective, but I kind of doubt it since the instruction offset was clearly pointing to `fillFrameRaw` code, which does not make sense in the default config. 

This clearly pointed to a similar cause as we fixed previously - when stored `methodId` is `nullptr` we kind of treat it as, yeah, it's deliberate, pull the `NMethod*` instead. But it turns out, we can legitimately be not able to resolve `jmethodid` and capturing `NMethod*` instead can blow up when class retransformation is involved.

zhengyu123:
> @zhengyu123 The debugger pinpointed the failing instruction within the code that is extracting data from `NMethod*` - we MUST NOT DO THAT when `force-jmethodids` is not set to false. This is a correctness fix that avoids falling into code which should not be executing with the current setup. Adding a longjump around it would be a bandaid. We can do that if this fix proves to be ineffective, but I kind of doubt it since the instruction offset was clearly pointing to `fillFrameRaw` code, which does not make sense in the default config.
> 
> This clearly pointed to a similar cause as we fixed previously - when stored `methodId` is `nullptr` we kind of treat it as, yeah, it's deliberate, pull the `NMethod*` instead. But it turns out, we can legitimately be not able to resolve `jmethodid` and capturing `NMethod*` instead can blow up when class retransformation is involved.

@jbachorik I 

> @zhengyu123 The debugger pinpointed the failing instruction within the code that is extracting data from `NMethod*` - we MUST NOT DO THAT when `force-jmethodids` is not set to false. This is a correctness fix that avoids falling into code which should not be executing with the current setup. Adding a longjump around it would be a bandaid. We can do that if this fix proves to be ineffective, but I kind of doubt it since the instruction offset was clearly pointing to `fillFrameRaw` code, which does not make sense in the default config.
> 
> This clearly pointed to a similar cause as we fixed previously - when stored `methodId` is `nullptr` we kind of treat it as, yeah, it's deliberate, pull the `NMethod*` instead. But it turns out, we can legitimately be not able to resolve `jmethodid` and capturing `NMethod*` instead can blow up when class retransformation is involved.

@jbachorik Okay, I can see the scenario (https://github.com/openjdk/jdk/blob/master/src/hotspot/share/prims/jvmtiRedefineClasses.cpp#L4413), it is a nasty race - fortunately, the array only grows larger.

I fail to see the downside for recording the `Method*` for later resolution - but why it is NMethod* instead ?- this must be a bug we want to hunt down.

jbachorik:
> I fail to see the downside for recording the Method* for later resolution - but why it is NMethod* instead ?- this must be a bug we want to hunt down.

I don't think we should be recording `Method*` (`NMethod*` was not the right type ...) if the feature is not enabled. It makes debugging potential instabilities totally confusing.

jbachorik:
@zhengyu123 I have modified the patch - the `jmethodid` is kept `nullptr` if it is not found in the cache but then we guard the `fillFrameRaw` and enter that only when using `Method*` is allowed by config.

zhengyu123 (approved):
Thanks.

PR #734, opened 2026-08-11
Make ProfiledThread's UnwindFailures field debug only

zhengyu123:
**What does this PR do?**:

Removes the UnwindFailures/UnwindStats per-thread diagnostic counters from release builds, gating them entirely behind #ifdef DEBUG, and backfills a test.

- New macro — DEBUG_ONLY(s) expands to s when DEBUG is defined, else to nothing.

- threadLocalData.h `ProfiledThread::_unwind_failures` is now debug_only(...), and the unwindFailures() accessor is wrapped in #ifdef DEBUG. This is the actual memory-overhead fix — the field no longer exists per-thread on release builds.
-  unwindStats.h: entire UnwindStats class and its static members wrapped in #ifdef DEBUG — the whole subsystem is compiled out on release.

-flightRecorder.cpp: Recording::writeUnwindFailures() body gated by #ifdef DEBUG (becomes a no-op on release).

- hotspotSupport.cpp: the call sites that record into UnwindFailures are gated by #ifdef DEBUG.

- build.gradle.kts: the unwindFailuresBenchmark now compiles with -DDEBUG, since the code it benchmarks no longer exists without that define.

- Tests: 
-- existing UnwindFailures/UnwindStats gtest cases in ddprof_ut.cpp wrapped in #ifdef DEBUG; 
-- new UnwindFailuresTest.java  — a JFR smoke test under cstack=vm that skips unless the build config is debug/*san, asserts the profiler runs cleanly, and sanity-checks datadog.UnwindFailure event schema if any happen to fire (a real stub-unwind failure is described as too rare to force deterministically).

**Motivation**:
Reduce profiler memory usage in production

**Additional Notes**:

**How to test the change?**:
- CI tests - no regression
- New test to test `UnwindFailures` in debug build

**For Datadog employees**:
- [X] JIRA: [PROF-15696](https://datadoghq.atlassian.net/browse/PROF-15696)

[PROF-15696]: https://datadoghq.atlassian.net/browse/PROF-15696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddprof-lib/src/main/cpp/common.h
```diff
@@ -32,6 +32,12 @@
 #  endif
 #endif
 
+#ifdef DEBUG
+#define debug_only(s)   s
```
jbachorik: Can we keep it upper case like the other macros?

ddprof-lib/src/main/cpp/threadLocalData.h
```diff
@@ -266,12 +268,14 @@ class ProfiledThread : public ThreadLocalData {
   inline void setFiRng(u64 seed) { _fi_rng = seed ? seed : 1; }
 #endif
 
+#ifdef DEBUG
```
jbachorik, line 271: Hm, can this be also handled by the macro? 
It feels a bit off, having two ways of handling `unwindFailures`

zhengyu123: This is multi-line block, cramping into DEBUG_ONLY() is not readable. 

This mirrors Hotspot coding style.

jbachorik (approved):

PR #743, opened 2026-08-19
Protect HotspotSupport::resolve() with longjmp

zhengyu123:
**What does this PR do?**:

Hardens `HotspotSupport::resolve()` — the dump-time fallback that resolves a raw `Method*` to a `jmethodID` when `cstack=vm,fjmethodid=false` — against crashes from stale JVM metadata, and fixes three separate bugs that caused `<clinit>` frames captured under that configuration to always serialize as `"unknown"` instead of resolving correctly:

1. `JVMSupport::initExecution()` short-circuited on a stale `Fully_loaded` state left over from a previous session, so a later `fjmethodid=false` start never re-evaluated whether jmethodID preloading should actually be disabled.
2. The `<clinit>` JVMTI fallback called `GetClassMethods` directly instead of going through `loadMethodIDsIfNeededImpl()`, skipping the JDK-8062116 classloader-data patch that every other preload path applies.
3. `HotspotSupport::fillJavaFrame()` read `VM::arguments()._force_jmethodID` — a global `Arguments` object that's only updated by the command-line agent-attach path, never by the `JavaProfiler.execute()` JNI API. Every session started via the JNI API (which is how the test suite, and most embedding use, starts the profiler) therefore saw a permanently stale `_force_jmethodID`, so un-preloaded methods were always treated as unwalkable and serialized as `"unknown"` instead of taking the intended raw-`Method*` fallback path. Fixed by giving `Profiler` its own `_force_jmethodID` member (mirroring the existing `_cstack` field) set from the actual startup `Arguments`.

Also adds a malloc-backed fallback for method/class/signature names that don't fit the fixed inline buffers used during crash-protected resolution (previously such names silently resolved as `"unknown"` with no fallback), and closes out the correctness gaps that fallback introduced: a double-free on the crash-recovery path (`release()` is now idempotent), a setjmp/longjmp indeterminate-value issue (the malloc'd pointers are now `volatile`, since they're mutated between `sigsetjmp()` and a possible `siglongjmp()` and then read back by the recovery path), and a dead-code fallback (the malloc path was unreachable for class-name/signature since their short-buffer size equaled the hard rejection ceiling — lowered to 1024 so the fallback actually engages).

**Motivation**:

`ClinitResolutionTest` was failing: a class's `<clinit>` frame, sampled while spinning for ~2s under `cpu=1ms,cstack=vm,fjmethodid=false`, never resolved to its real class/method name in the resulting JFR — it always showed as the shared `"unknown"` frame. Root-causing this surfaced the three bugs above, none of which were specific to `<clinit>` — they affect any raw-`Method*` resolution under `fjmethodid=false`, `<clinit>` just happened to be the scenario the test exercised.

**Additional Notes**:

- Fixes (1) and (2) were ported from `paul.fournillon/jmethod_clinit_fix`, adapted to this branch's crash-protection refactor.
- Known, lower-priority items surfaced during review but intentionally not addressed here (happy to split into follow-ups if preferred):
  - Name/signature resolution is still bounded (512/1024/4096 bytes, not the full 65535 the JVM permits) — a deliberate fidelity/complexity tradeoff, not a full fix.
  - `METHOD_RESOLVE_FAULT_RECOVERED` overlaps `STACKWALK_LONGJMP_RECOVERED` for the same fault; nothing enforces the "subtract, never sum" comment, so a naive counter-summing dashboard would double-count.
  - The `<clinit>` fallback's `patchClassLoaderData()` call (JDK 8 only) is now reachable from the dump thread as well as from `ClassPrepare`; a narrow race could invoke it twice for the same class.
  - `resolve()`'s top-level doc comment ("only resolves system-classloader methods") is slightly stale given the `<clinit>` fallback's `load_all=true`; the constraint still holds in practice via the earlier `FindClass` call, so this is cosmetic.

**How to test the change?**:

- `./gradlew :ddprof-lib:gtestDebug_hotspotSupport_ut :ddprof-lib:gtestDebug_jvmSupport_ut` — new/updated unit tests for `JVMSupport::initExecution()`'s stale-state bug and the `<clinit>` fallback's `loadMethodIDsIfNeededImpl()` routing.
- `./gradlew testdebug --tests com.datadoghq.profiler.cpu.ClinitResolutionTest` — the regression test for this bug, passes reliably on the first attempt (previously failed even after all 5 retries).
- `./gradlew testdebug --tests "com.datadoghq.profiler.cpu.*"` — full `cpu` test package, including `VtableReceiverFrameTest`, all passing.

**For Datadog employees**:
- [X] JIRA: [PROF-15786](https://datadoghq.atlassian.net/browse/PROF-15786)

[PROF-15786]: https://datadoghq.atlassian.net/browse/PROF-15786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/counters.h
+++ b/ddprof-lib/src/main/cpp/counters.h
@@ -134,6 +134,17 @@
   X(SAFECOPY_FAILED, "safecopy_failed")                                       \
   X(SAFEFETCH_FAILED, "safefetch_failed")                                     \
   X(STACKWALK_LONGJMP_RECOVERED, "stackwalk_longjmp_recovered")               \
+  /* Dump-time raw-Method* resolution (HotspotSupport::resolve, reached only    \
+   * for cstack=vm + fjmethodid=false frames). NOT additive with               \
+   * STACKWALK_LONGJMP_RECOVERED: checkFault() bumps that one unconditionally  \
+   * before every siglongjmp, so each fault counted here is counted there too. \
+   * Subtract, never sum. Non-zero means stale HotSpot metadata (GC or class   \
+   * unloading) reached the dump thread; the frame serializes as "unknown". */ \
+  X(METHOD_RESOLVE_FAULT_RECOVERED, "method_resolve_fault_recovered")         \
+  /* Symbol length/body rejected during the same resolution: unreadable,       \
+   * empty, or longer than the fixed dump-time buffers in hotspotSupport.cpp.  \
+   * The frame serializes as "unknown". */                                     \
+  X(METHOD_RESOLVE_SYMBOL_UNREADABLE, "method_resolve_symbol_unreadable")     \
   /* Strict subset of SAMPLES_DROPPED_THREAD_LOCAL, not an independent count: \
    * ThreadLocalDataPool::claim() increments this on capacity exhaustion, and \
    * every acquireCurrent() caller that gets nullptr back -- for this or any  \
--- a/ddprof-lib/src/main/cpp/counters.h
+++ b/ddprof-lib/src/main/cpp/counters.h
@@ -141,9 +141,11 @@
    * Subtract, never sum. Non-zero means stale HotSpot metadata (GC or class   \
    * unloading) reached the dump thread; the frame serializes as "unknown". */ \
   X(METHOD_RESOLVE_FAULT_RECOVERED, "method_resolve_fault_recovered")         \
-  /* Symbol length/body rejected during the same resolution: unreadable,       \
-   * empty, or longer than the fixed dump-time buffers in hotspotSupport.cpp.  \
-   * The frame serializes as "unknown". */                                     \
+  /* Symbol length/body rejected during the same resolution: unreadable body,  \
+   * empty (recycled slot), or over MAX_SYMBOL_LEN. A name that merely exceeds \
+   * the fixed inline buffers in hotspotSupport.cpp's ResolvedNames still      \
+   * resolves via its malloc fallback and does not land here. The frame        \
+   * serializes as "unknown". */                                               \
   X(METHOD_RESOLVE_SYMBOL_UNREADABLE, "method_resolve_symbol_unreadable")     \
   /* Strict subset of SAMPLES_DROPPED_THREAD_LOCAL, not an independent count: \
    * ThreadLocalDataPool::claim() increments this on capacity exhaustion, and \
--- a/ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
+++ b/ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
@@ -1437,6 +1437,165 @@ bool HotspotSupport::loadMethodIDsIfNeededImpl(jvmtiEnv *jvmti, JNIEnv *jni, jcl
     return JVMSupport::loadMethodIDsImpl(jvmti, jni, klass);
 }
 
+// Fixed dump-time buffers, replacing the malloc/free trio this function used to
+// carry. HotSpot's Symbol length field is a u2, so the VM permits up to 65535
+// bytes; these caps trade that theoretical maximum for zero heap traffic on the
+// dump path and, more importantly, for having no cleanup obligation inside the
+// crash-protected region -- a siglongjmp out of resolve() cannot leak anything.
+// 4096 mirrors Lookup::resolveVTableReceiverCached (flightRecorder.cpp), which
+// reads the same Symbol bodies on the same thread. A symbol over its cap reports
+// METHOD_RESOLVE_SYMBOL_UNREADABLE and serializes as "unknown" -- the same
+// outcome as a class FindClass cannot see.
+static const size_t MAX_KLASS_NAME_LEN  = 4096;  // internal names; realistically < 256
+static const size_t MAX_METHOD_NAME_LEN =  512;  // realistically < 64
+// A descriptor can in principle exceed this (255 argument *slots*, each able to
+// carry an arbitrarily long L...; type name), so this cap is a fidelity choice,
+// not a proof: over-cap descriptors serialize as "unknown" rather than being
+// truncated. METHOD_RESOLVE_SYMBOL_UNREADABLE makes it visible if that bites.
+static const size_t MAX_SIGNATURE_LEN   = 4096;
+
+// Copies a Symbol's body into a fixed buffer, NUL-terminating it.
+// Must be called with crash protection installed: length() and body()
+// dereference the Symbol directly with no safefetch.
+static bool copySymbolBody(VMSymbol* sym, char* dst, size_t cap) {
+  unsigned len = sym->length();  // raw u2 deref; PC stays inside this library
+  // A method name, descriptor or class name is never empty; 0 means the Symbol
+  // slot has been recycled. `>=` leaves room for the NUL.
+  if (len == 0 || len >= cap) {
+    return false;
+  }
+  const char* body = sym->body();
+  if (!SafeAccess::safeCopy(dst, body, len)) {
+    return false;
+  }
+  dst[len] = '\0';
+  return true;
+}
+
+// The three names resolve() needs, owned by resolve()'s frame so that the
+// metadata walk has no cleanup obligation of its own.
+struct ResolvedNames {
+  char method_name[MAX_METHOD_NAME_LEN];
+  char method_signature[MAX_SIGNATURE_LEN];
+  char klass_name[MAX_KLASS_NAME_LEN];
+};
+
+// PHASE 1 -- the raw HotSpot metadata walk. MUST run with a jmp ctx installed:
+// every step is a raw *(void**)(this + offset) whose target may have been freed
+// by GC or class unloading since the sample was taken. Deliberately contains no
+// JNI, no JVMTI and no allocation, so the whole protected region stays inside
+// this library, where Profiler::checkFault() can actually recover.
+//
+// Returns false if the metadata is unusable. On success either *out_id holds an
+// already-valid jmethodID (and `names` is untouched), or *out_id is null and
+// `names` has been filled in for the JNI lookup the caller does afterwards.
+static bool readMethodNames(const void* method, VMMethod** out_vm_method,
+                            jmethodID* out_id, ResolvedNames* names) {
+  *out_vm_method = nullptr;
+  *out_id = nullptr;
+
+  VMMethod* vm_method = VMMethod::cast_or_null(method);
+  if (vm_method == nullptr) {
+    return false;
+  }
+  *out_vm_method = vm_method;
+
+  // May have been populated by following code or JMETHODID_NOT_WALKABLE
+  jmethodID method_id = vm_method->validatedId();
+  if (isValidJMethodID(method_id)) {
+    *out_id = method_id;
+    return true;
+  }
+
+  VMConstMethod* const_method = vm_method->constMethod_or_null();
+  if (const_method == nullptr) {
+    return false;
+  }
+
+  VMConstantPool* const_pool = const_method->constants_or_null();
+  if (const_pool == nullptr) {
+    return false;
+  }
+
+  VMSymbol* name_sym = const_method->name();
+  VMSymbol* sig_sym = const_method->signature();
+  VMKlass* klass = const_pool->holder_or_null();
+
+  if (name_sym == nullptr || sig_sym == nullptr || klass == nullptr) {
+    return false;
+  }
+
+  VMSymbol* klass_sym = klass->name();
+  if (klass_sym == nullptr) {
+    return false;
+  }
+
+  if (!copySymbolBody(name_sym, names->method_name, sizeof(names->method_name)) ||
+      !copySymbolBody(sig_sym, names->method_signature, sizeof(names->method_signature)) ||
+      !copySymbolBody(klass_sym, names->klass_name, sizeof(names->klass_name))) {
+    Counters::increment(METHOD_RESOLVE_SYMBOL_UNREADABLE);
+    return false;
+  }
+  return true;
+}
+
+// PHASE 2 -- the JNI/JVMTI lookup, reading only the buffers phase 1 filled.
+// MUST run with crash protection *off*. Two reasons, both about siglongjmp
+// unwinding frames it must not:
+//   - A fault inside libjvm.so is unrecoverable anyway (checkFault() only
+//     recovers PCs inside this library), so a landing pad buys nothing here.
+//   - FindClass() loads the class when it is not already loaded, which
+//     synchronously runs our own JVMTI ClassPrepare callback ->
+//     patchClassLoaderData(), which holds the JVM's ClassLoaderData mutex. That
+//     code *is* in this library, so with a pad installed a fault there would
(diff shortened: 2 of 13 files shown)


r1viollet:
Can readMethodNames() be rewritten as a dedicated SafeAccess-only traversal, rather than installing a broad siglongjmp landing pad?

r1viollet:
This effectively protects for crashes, but are we not worried about downstream effects of passing invalid data ? Using data that was freed would not crash but still produce invalid data.

r1viollet:
I like the deterministic tests, though have we considered fuzzing for this type of problem ? 
I can imagine we could:
 - randomize the metadata graph
 - fault every pointer edge
 - mutate every offset, index and length
 - exercise readable-but-recycled metadata
 - fuzz combinations of nested signals and cleanup states

zhengyu123:
> Can readMethodNames() be rewritten as a dedicated SafeAccess-only traversal, rather than installing a broad siglongjmp landing pad?

@r1viollet The goal is the opposite - `SafeAccess` is *very* expensive, because every single plain load becomes a method call, and the cost to recover is literally the same as siglongjmp - a trip to signal handler and unwinding states.

Once completing `siglongjmp` protections, I would like to remove all `SafeAccess` inside the protected blocks, that should improve performance.

zhengyu123:
> This effectively protects for crashes, but are we not worried about downstream effects of passing invalid data ? Using data that was freed would not crash but still produce invalid data.

@r1viollet Yes, this is my main concern - this approach does mask off some bad bugs inside the protected blocks. I am think about setting up a special build, that disables `siglongjmp` protection for testing ...

zhengyu123:
> I like the deterministic tests, though have we considered fuzzing for this type of problem ? I can imagine we could:
> 
> * randomize the metadata graph
> * fault every pointer edge
> * mutate every offset, index and length
> * exercise readable-but-recycled metadata
> * fuzz combinations of nested signals and cleanup states

@r1viollet Yes, we may need more fuzzing tests. 

I think fault-injection and fuzzing overlap in certain ways, and fault-injection has certain advantages, please refer to my reply to Thomas' comment [here](https://datadoghq.atlassian.net/wiki/spaces/~712020ea9694b4505d4d3e97e094ef9d253abe/pages/6942556488/Fault+injection+framework?focusedCommentId=6992953539)

Fault-injection allows fault to propagate through call graph, instead of selected points.

jbachorik (approved):
Nice cleanup! Looks sane to me, there is one minor issue Sphinx found, if you can address it before merge, it would be great, but it is not a high priority.

zhengyu123:
> Nice cleanup! Looks sane to me, there is one minor issue Sphinx found, if you can address it before merge, it would be great, but it is not a high priority.

@jbachorik I don't see any comments from Sphinx, could you share?

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -1437,103 +1437,331 @@ bool HotspotSupport::loadMethodIDsIfNeededImpl(jvmtiEnv *jvmti, JNIEnv *jni, jcl
...
+  // A method name, descriptor or class name is never empty; 0 means the Symbol
+  // slot has been recycled. `>=` leaves room for the NUL.
+  if (len == 0 || len >= MAX_SYMBOL_LEN) {
+    return false;
+  }
+
+  char* dest = short_name;
+  if (len >= short_limit) {
+    long_name = (char*)malloc(len + 1);
+    dest = long_name;
```
jbachorik, line 1557: **[HIGH] robustness**: `ResolvedNames::setImpl()` does not check the return value of `malloc()`. On allocation failure `long_name` stays nullptr, `dest` is set to nullptr, and the code proceeds to call `SafeAccess::safeCopy(nullptr, sym->body(), len)` and then `dest[len] = '\0'` — a write through a null pointer. `safeCopy`'s fault protection covers reads of the (possibly stale) source Symbol body; a store to address 0 either faults (turning an OOM into a siglongjmp that is then miscounted as `METHOD_RESOLVE_FAULT_RECOVERED`) or corrupts memory.

Suggested fix:
```cpp
if (len >= short_limit) {
  long_name = (char*)malloc(len + 1);
  if (long_name == nullptr) {
    return false;   // caller bumps METHOD_RESOLVE_SYMBOL_UNREADABLE
  }
  dest = long_name;
}
```
This keeps the failure on the deliberate "serialize as unknown" path instead of relying on the crash-recovery landing pad to absorb a null store.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -1437,103 +1437,331 @@ bool HotspotSupport::loadMethodIDsIfNeededImpl(jvmtiEnv *jvmti, JNIEnv *jni, jcl
...
+  // Hard ceiling for the malloc fallback in setImpl(), independent of which
+  // field is being resolved. Not a JVM/class-file limit (Symbol::length() is
+  // a u2, so up to 65535 is legal) -- a name that would already have been rejected
+  // as too long for those fields' fixed buffers doesn't get an unbounded
+  // allocation just because it went through the malloc path instead.
+  static constexpr size_t MAX_SYMBOL_LEN = 64 * 1024;
+
+  // Fixed-size fast-path buffers for the common case, sized so that a
+  // realistic name/signature/class name never needs to allocate. 4096 mirrors
+  // Lookup::resolveVTableReceiverCached (flightRecorder.cpp), which reads the
```
jbachorik: **[MEDIUM] consistency**: This comment describes a 4096-byte buffer that no longer exists in the shipped code ("4096 mirrors `Lookup::resolveVTableReceiverCached` (flightRecorder.cpp) ... precedent to stay consistent with"). The actual constants below are 1024/512/1024, and the PR description confirms 4096 was deliberately lowered to 1024 so the malloc fallback engages. Comment and code now contradict each other.

Suggestion: drop or rewrite the 4096/resolveVTableReceiverCached sentence to match the constants actually declared, and update the `counters.h` comment accordingly (see separate comment on that file).

ddprof-lib/src/test/cpp/hotspotMethodId_ut.cpp
```diff
@@ -237,3 +312,314 @@ TEST(HotspotMethodIdTest, IdReturnsValidIdForPopulatedSlot) {
...
+    f->link();
+    ResolveFakes::setSymbol(f->sig_sym, "()V");
+    ResolveFakes::setSymbol(f->klass_sym, "java/lang/Object");
+    f->name_sym.length = 0xFFFF;  // far above MAX_METHOD_NAME_LEN
+
+    long long before = Counters::getCounter(METHOD_RESOLVE_SYMBOL_UNREADABLE);
+
+    EXPECT_EQ(nullptr, HotspotSupport::resolve(&f->method));
+
+    EXPECT_EQ(before + 1, Counters::getCounter(METHOD_RESOLVE_SYMBOL_UNREADABLE));
```
jbachorik, line 727: **[MEDIUM] correctness**: `ResolveReturnsNullForOverlongSymbol` doesn't test what its name/comment claims ("pins the cap behaviour"). It sets `name_sym.length = 0xFFFF` (65535), which is *below* `MAX_SYMBOL_LEN` (65536), so the length check doesn't reject it — `setImpl` takes the malloc path, allocates 64KiB, and fails only because `SafeAccess::safeCopy` can't read 65535 bytes from a 64-byte fake body. The test passes for the wrong reason.

Suggestion: split into two tests — (a) a name exceeding `MAX_METHOD_NAME_LEN` but with a fully readable body, asserting the malloc fallback succeeds; (b) a length at/above `MAX_SYMBOL_LEN`, asserting the hard ceiling rejects it. Neither is currently covered.

ddprof-lib/src/test/cpp/jvmSupport_ut.cpp
```diff
@@ -193,3 +212,82 @@ TEST(JvmSupportErrorLatchTest, CheckStateStaysBlockedOnceInError) {
...
+
+    JVMSupport::initExecution(full_args, &mock_jvmti, reinterpret_cast<JNIEnv*>(&mock_jni));
+    EXPECT_EQ(JVMSupportTestAccessor::FullyLoaded(), JVMSupportTestAccessor::getLoadState());
+
+    Arguments partial_args;
+    partial_args._force_jmethodID = false;
+    partial_args._cstack = CSTACK_VM; // -> shouldPreloadJmethodIDs()==false -> Partial_loaded
+
+    JVMSupport::initExecution(partial_args, &mock_jvmti, reinterpret_cast<JNIEnv*>(&mock_jni));
+
```
jbachorik, line 291: **[MEDIUM] consistency**: This regression test's comment still says "Fails today: the stale Fully_loaded state short-circuits initExecution() ..." even though this same PR removes that short-circuit and makes the assertion pass. Future readers will read this as a known-failing test.

Suggestion: reword to describe the guarded behavior, e.g. "Regression: before the fix, a stale Fully_loaded state short-circuited initExecution()... so the state stayed Fully_loaded instead of downgrading."

ddprof-lib/src/test/cpp/jvmSupport_ut.cpp
```diff
@@ -193,3 +212,82 @@ TEST(JvmSupportErrorLatchTest, CheckStateStaysBlockedOnceInError) {
...
+        jni_tbl.FindClass = &mock_FindClass_returns_null;
+        jni_tbl.ExceptionClear = &mock_ExceptionClear_noop;
+        mock_jni.functions = &jni_tbl;
+    }
+
+    void TearDown() override {
+        VMTestAccessor::setHotspot(_orig_hotspot);
+    }
+};
+
```
jbachorik, line 278: **[LOW] robustness**: `JVMSupportRestartTest::SetUp()` resets the process-global `JVMSupport::jmethodID_load_state` to `No_loaded`, but `TearDown()` only restores `VM::_hotspot`. The test leaves global load state at `Partial_loaded` (or `Fully_loaded` if it fails midway), making other tests in the same gtest binary order-dependent.

Suggestion: snapshot the load state in `SetUp()` alongside `_orig_hotspot` and restore it in `TearDown()`; `JVMSupportTestAccessor` already exposes both getter and setter.

ddprof-lib/src/main/cpp/jvmSupport.cpp
```diff
@@ -71,10 +71,6 @@ void JVMSupport::setLoadState(JMethodIDLoadStats state) {
 
 void JVMSupport::initExecution(Arguments& args, jvmtiEnv* jvmti, JNIEnv* jni) {
     JMethodIDLoadStats current_state = getLoadState();
```
jbachorik, line 73: **[LOW] performance**: Removing `if (current_state == Fully_loaded) return;` fixes the stale-state bug but also removes the only guard against repeating the full jmethodID preload on every profiler start. A restart with the same `fjmethodid=true` args now re-runs `GetLoadedClasses`/`GetClassMethods` over the whole loaded-class set on every start — previously a no-op.

Suggestion: compute the desired state from `args` first, then skip the preload only when it already matches the current state, e.g. `if (current_state == Fully_loaded && shouldPreloadJmethodIDs(args)) return;`, so a same-mode restart stays free while a mode change is still re-evaluated.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -1437,103 +1437,331 @@ bool HotspotSupport::loadMethodIDsIfNeededImpl(jvmtiEnv *jvmti, JNIEnv *jni, jcl
...
+// vm_method->validatedId() below is safefetch-based, so it is safe unprotected.
+jmethodID lookupMethodIdViaJni(VMMethod* vm_method, const ResolvedNames& names) {
+  jmethodID method_id = nullptr;
+  const char* method_name = names.methodName();
+  const char* method_signature = names.methodSignature();
+  const char* klass_name = names.klassName();
+
+  JNIEnv *jni = VM::jni();
+  jclass clz = jni->FindClass(klass_name);
+  if (clz == nullptr) {
```
jbachorik, line 1678: **[LOW] consistency**: `lookupMethodIdViaJni()` is defined with external linkage in the global namespace (no `static`, no anonymous namespace) and is only declared via a friend declaration in the header. This exports an internal helper symbol from the shared library, unlike the adjacent `readMethodNames()` which is correctly `static`. Ownership/contract of the returned `jmethodID` is also undocumented here.

Suggestion: give it internal-linkage-equivalent hiding, or make it a private static member of `HotspotSupport` and drop the free-function friend declaration.

ddprof-lib/src/main/cpp/counters.h
```diff
@@ -134,6 +134,17 @@
...
   X(STACKWALK_LONGJMP_RECOVERED, "stackwalk_longjmp_recovered")               \
+  /* Dump-time raw-Method* resolution (HotspotSupport::resolve, reached only    \
+   * for cstack=vm + fjmethodid=false frames). NOT additive with               \
+   * STACKWALK_LONGJMP_RECOVERED: checkFault() bumps that one unconditionally  \
+   * before every siglongjmp, so each fault counted here is counted there too. \
+   * Subtract, never sum. Non-zero means stale HotSpot metadata (GC or class   \
+   * unloading) reached the dump thread; the frame serializes as "unknown". */ \
+  X(METHOD_RESOLVE_FAULT_RECOVERED, "method_resolve_fault_recovered")         \
+  /* Symbol length/body rejected during the same resolution: unreadable,       \
+   * empty, or longer than the fixed dump-time buffers in hotspotSupport.cpp.  \
```
jbachorik: **[LOW] consistency**: This doc comment for `METHOD_RESOLVE_SYMBOL_UNREADABLE` is already stale relative to code added in the same PR — it says rejection happens when the symbol is "longer than the fixed dump-time buffers," but the new malloc fallback means exceeding those buffers no longer causes rejection.

Suggestion: restate as unreadable body, zero length, or (if retained) over the hard ceiling — drop the reference to the fixed inline buffers, which are now only an allocation fast path.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.h
```diff
@@ -18,9 +18,16 @@
 
 class ProfiledThread;
 class VMMethod;
+class ResolvedNames;
```
jbachorik: **[LOW] consistency**: The public header now forward-declares `ResolvedNames` — a class defined entirely inside `hotspotSupport.cpp` with no header of its own — and befriends a non-static free function `lookupMethodIdViaJni` with external linkage. Both are `.cpp` implementation details leaking into the public interface every consumer compiles against.

Suggestion: make `lookupMethodIdViaJni` a `static` member of `HotspotSupport` (or a private static helper), removing both the free-function friend declaration and the need to name `ResolvedNames` in the header.

jbachorik (commented):
Automated review findings (sphinx multi-stream review, deduplicated 31 → 9 after adversarial verification).

ddprof-lib/src/main/cpp/jvmSupport.cpp
zhengyu123: Apparently, it has been fixed by commit 1f2e1163c ("fix: profiler looks at wrong fjmethodid flag")

PR #746, opened 2026-08-21
Fix CollectedHeapWrapper stale layout on JDK 25+

jbachorik:
**What does this PR do?**:
Version-gates `CollectedHeapWrapper`'s hand-maintained layout of HotSpot's internal `CollectedHeap` class. JDK-8356848 ("Separate Metaspace and GC printing"), shipped in JDK 25 GA, inserted a `GCMetaspaceLog*` field ahead of `_capacity_at_last_gc`/`_used_at_last_gc`, shifting both fields one pointer width later on 64-bit builds. The wrapper was reading through the old (pre-25) layout unconditionally, so on JDK 25+ `_used_at_last_gc` actually read `_capacity_at_last_gc`'s bytes.

Splits the wrapper into `CollectedHeapWrapperPre25` and `CollectedHeapWrapperV25Plus`, selected via a `hotspot_version() >= 25` gate in `collectedHeapUsedAtLastGc()`/`collectedHeapCapacityAtLastGc()`.

**Motivation**:
Reported as PROF-15803: on JDK 25+, `_used_at_last_gc` (used to detect a rising heap floor) was reporting the heap's capacity instead, breaking floor-based leak detection heuristics.

**Additional Notes**:
Verified via GitHub's compare API that the upstream layout-changing commit (`85af573cb6`) is an ancestor of the `jdk-25-ga` tag, i.e. the break aligns exactly with the JDK 25 major-version boundary rather than a later update release, so a `hotspot_version() >= 25` gate is correct and sufficient.

**How to test the change?**:
Added `ddprof-lib/src/test/cpp/jvmHeap_ut.cpp`: pins both wrapper layouts' relative size, verifies the version gate selects the correct layout on both sides of the boundary, reproduces the exact reported symptom (capacity misread as used when the gate is missing), and sanity-checks that used never exceeds capacity in either layout. All tests pass under `./gradlew :ddprof-lib:gtestDebug_jvmHeap_ut`.

- [x] JIRA: [PROF-15803]

[PROF-15803]: https://datadoghq.atlassian.net/browse/PROF-15803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/hotspot/vmStructs.cpp
+++ b/ddprof-lib/src/main/cpp/hotspot/vmStructs.cpp
@@ -1126,8 +1126,8 @@ HeapUsage HeapUsage::get(bool allow_jmx) {
         if (_heap_usage_func != NULL) {
             // this is the JDK 17+ path
             usage = _heap_usage_func(*(char**)_collected_heap_addr);
-            usage._used_at_last_gc =
-                ((CollectedHeapWrapper *)*(char**)_collected_heap_addr)->_used_at_last_gc;
+            usage._used_at_last_gc = collectedHeapUsedAtLastGc(
+                *(char**)_collected_heap_addr, VM::hotspot_version());
         } else if (_gc_heap_summary_func != NULL) {
             // this is the JDK 11 path
             // we need to collect GCHeapSummary information first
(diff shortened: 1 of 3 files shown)


zhengyu123 (approved):
LGTM

PR #752, opened 2026-08-24
Protect Lookup::resolveMethod() with siglongjmp

zhengyu123:
**What does this PR do?**:

***Crash protection for method symbolication***
Note: This mechanism cannot protect against faults originating from JNI or JVMTI calls, as the faulting PCs are outside the profiler’s address space. Since JNI and JVMTI are public APIs, such faults should generally not occur unless the APIs are misused.
 
Split `resolveMethod()` into a fast, deliberately-unprotected path (already-marked cache hit — provably non-faulting) and a new `fillMethod()` slow path, armed with its own `sigsetjmp/JmpCtxScope` landing pad before touching any VM metadata.

Added `ResolveMethodState`, an RAII holder for the intermediate JNI/JVMTI-owned pointers (`_demangled`, `_class_name`, `_method_name`, `_method_signature`, `_line_number_table`) acquired mid-resolution. `state.release()` frees whatever's still pending — on both the normal path and the `siglongjmp` recovery path — via `jvmti->Deallocate()`.

Added a shared `_unknown_method` row (`Lookup::unknownMethod()`) that lives outside `MethodMap`, so the disarmed recovery branch can return a valid, already-marked row without allocating (a `MethodMap::operator[]` insert can throw `std::bad_alloc`, which must not happen with no landing pad installed).
`fillJavaMethodInfo()` now returns bool so a row is only marked/keyed when it was actually filled, preventing a permanently-stuck empty row when `PushLocalFrame` fails or the JVM isn't in JVMTI_PHASE_START/LIVE.
Added `INJECT_CRASH_LIKELY()` inside the protected window (compiled only under -PenableFaultInjection) to exercise the recovery path under test.

***Line-number-table handling***

`SharedLineNumberTable` no longer makes a private malloc'd copy of the JVMTI buffer — it now holds the buffer JVMTI itself returned and frees it via `jvmti->Deallocate()` instead of `free()`. (The previous copy-then-free() pairing could crash under -XX:NativeMemoryTracking, since HotSpot's Allocate/Deallocate route through NMT-tracked os::malloc/os::free, which add a header free() doesn't know about.)
Validation before trusting the buffer is now: a sanity bound on the reported size (MAX_LINE_NUMBER_TABLE_ENTRIES = 65535, the JVM spec's code_length cap) and a `SafeAccess` probe of the pointer — `isReadableRange()` for a non-zero size, `isReadable()` for the legitimate zero-size case Hotspot returns (calling `isReadableRange(ptr, 0)` would trip its `assert(size > 0)`).
LINE_NUMBER_TABLE_UNREADABLE is now actually incremented when the table is rejected.

***Signal-depth bookkeeping***

`_signal_depth` widened from `uint8_t` to `int`, and `exitSignalScope()/getInSignalDepth()` now assert() on an unmatched exit instead of silently clamping at zero — a pairing bug becomes visible in debug/test builds rather than self-healing invisibly. `isInTrackedSignalContext()` checks `> 0` rather than `!= 0` so a (now theoretically possible) negative depth reads as "not in a signal context" rather than permanently pinning dlopen_hook to its synchronous refresh path.

***Misc***

faultInjection.h/.cpp: `crashNow()`'s declaration and definition are now consistently guarded behind `__FAULT_INJECTION__ || DEBUG`; the unused INJECT_CRASH_ALWAYS() macro (previously only defined in the disabled branch) was removed; Profiler's test-only forced-crash path now calls crashNow() instead of a local null-deref.
counters.h: added METHOD_RESOLUTION_DROPPED_TLS (dump thread couldn't allocate a ProfiledThread, expected to stay 0); trimmed the METHOD_RESOLVE_FAULT_RECOVERED comment (previously explained it as a non-additive subset of STACKWALK_LONGJMP_RECOVERED, tied to the old raw-Method*-only code path — worth re-checking whether that non-additive relationship still holds and re-documenting it, since the comment was dropped rather than updated).

**Motivation**:
`Lookup::resolveMethod()` resolves `jmethodID`s into MethodInfo rows at JFR dump time by calling JVMTI (`GetMethodDeclaringClass`, `GetClassSignature`, `GetMethodName`, `GetLineNumberTable`). If the declaring class was unloaded between sample capture and dump, these calls can return stale/garbage pointers even with `JVMTI_ERROR_NONE` — production crash telemetry confirmed this on stock HotSpot, not just OpenJ9. This PR wraps that resolution path in a `sigsetjmp/siglongjmp` recovery window (building on the pattern already used for `HotspotSupport::resolve()`, #743) and fixes a related correctness/lifetime bug in how the JVMTI line-number-table buffer is validated and owned.

**Additional Notes**:

**How to test the change?**:
- New lookup_resolveMethod_ut.cpp: exercises fillNativeMethodInfo()'s demangle-cut-cache-free lifecycle, and fillJavaMethodInfo()'s PushLocalFrame/PopLocalFrame balance (normal return and PushLocalFrame failure) via faked JNI/JVMTI, plus a line-number-table-survives-re-resolve-in-a-later-chunk case.

- lineNumberTableCopy_ut.cpp rewritten to match the current hold-not-copy design: rejects an unmapped source, accepts a valid table and confirms it's held (not copied), accepts a zero-size table without tripping the isReadableRange assert, and rejects negative/oversized reported sizes. Ran directly (gtestDebug_lineNumberTableCopy_ut) — all 6 cases pass.

- hotspotMethodId_ut.cpp updated: a rejected method id now resolves to the shared _unknown_method row instead of inserting into MethodMap.

- signalSafety_ut.cpp: removed the now-invalid "saturates at zero" test, since double-exit is now an assert failure by design, not a saturating no-op.
**For Datadog employees**:
- [X] JIRA: [PROF-15645](https://datadoghq.atlassian.net/browse/PROF-15645)

[PROF-15645]: https://datadoghq.atlassian.net/browse/PROF-15645?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -121,6 +121,8 @@ ResolveMethodState::~ResolveMethodState() {
   release();
 }
 
+// Don't expect following JNI and JVMTI calls to fail, as they are public APIs
+// of JVM
 void ResolveMethodState::release() {
   if (_framePushed) {
     JNIEnv* jni = VM::jni();
@@ -312,13 +314,23 @@ bool Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
         jvmtiError line_table_error = jvmti->GetLineNumberTable(method, &line_number_table_size,
                                   &line_number_table);
         
-        bool is_table_readable = line_number_table != nullptr &&
-                                 SafeAccess::isReadableRange(line_number_table, line_number_table_size * sizeof(jvmtiLineNumberEntry));
+        bool is_table_valid = (line_number_table_size >= 0 && line_number_table_size <= MAX_LINE_NUMBER_TABLE_ENTRIES);
+        bool is_table_readable =  (line_number_table != nullptr && is_table_valid &&
+                                   (line_number_table_size > 0 ?
+                                    SafeAccess::isReadableRange(line_number_table, line_number_table_size * sizeof(jvmtiLineNumberEntry)) :
+                                    SafeAccess::isReadable(line_number_table)));
         if (line_table_error != JVMTI_ERROR_NONE ||
-          // On Hotspot, it returns a malloc'd pointer even table size = 0, but there is no point to keep it
-          line_number_table_size == 0) {
-            if (is_table_readable && line_number_table != nullptr) {
+            // On Hotspot, it returns a malloc'd pointer even table size = 0, but there is no point to keep it
+            line_number_table_size == 0 ||
+            // a corrupted table 
+            !is_table_valid ||
+            !is_table_readable) {
+            if (is_table_readable) {
               jvmti->Deallocate((unsigned char*)line_number_table);
+            }
+
+            if (!is_table_valid || !is_table_readable) {
+              Counters::increment(LINE_NUMBER_TABLE_UNREADABLE);
             }
             line_number_table = nullptr;
             line_number_table_size = 0;
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -705,7 +705,7 @@ MethodInfo *Lookup::fillMethod(ASGCT_CallFrame &frame, jmethodID method_id,
   }
 
   assert(method_id != nullptr && "Already filtered by caller");
-    // Reinstates the thread's previous landing pad on every exit from this frame,
+  // Reinstates the thread's previous landing pad on every exit from this frame,
   // including a std::bad_alloc thrown by one of the map or dictionary inserts
   // underneath. Leaving ours installed past the end of this frame would leave
   // checkFault() jumping into a dead stack frame.
@@ -726,7 +726,7 @@ MethodInfo *Lookup::fillMethod(ASGCT_CallFrame &frame, jmethodID method_id,
     jmp_scope.restore();
     Counters::increment(METHOD_RESOLVE_FAULT_RECOVERED);
     // state.release() may fault. Unfortunately, it faults outside of profiler
-    // code where checkFault() can not absorbs.
+    // code where checkFault() can not absorb.
     state.release();
     // A member, already filled above -- no map lookup, no allocation, and no
     // reliance on a local surviving siglongjmp (the value of a non-volatile
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -329,7 +329,13 @@ bool Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
               jvmti->Deallocate((unsigned char*)line_number_table);
             }
 
-            if (!is_table_valid || !is_table_readable) {
+            // Only a genuinely corrupt/unreadable table output should count
+            // here: JVMTI_ERROR_ABSENT_INFORMATION (compiled without debug
+            // line info) and other non-success errors are ordinary, expected
+            // outcomes with no table to speak of, not corruption -- counting
+            // them would swamp this signal with normal methods and mask real
+            // stale/corrupted-jmethodID failures.
+            if (line_table_error == JVMTI_ERROR_NONE && (!is_table_valid || !is_table_readable)) {
               Counters::increment(LINE_NUMBER_TABLE_UNREADABLE);
             }
             line_number_table = nullptr;
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -167,7 +167,7 @@ void Lookup::cutArguments(char *func) {
 }
 
 void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
-                                bool first_time, bool& framePushed) {
+                                bool first_time, volatile bool& framePushed) {
   JNIEnv *jni = VM::jni();
   if (jni->PushLocalFrame(64) != 0) {
     return;
@@ -639,7 +639,7 @@ MethodInfo *Lookup::fillMethod(ASGCT_CallFrame &frame, jmethodID method_id,
   // underneath. Leaving ours installed past the end of this frame would leave
   // checkFault() jumping into a dead stack frame.
   JmpCtxScope jmp_scope(prof_thread);
-  bool framePushed = false;
+  volatile bool framePushed = false;
 
   sigjmp_buf crash_protection_ctx;
   // savemask must be 1: the siglongjmp originates inside segvHandler, where
@@ -1809,7 +1809,7 @@ int Recording::writeMethods(Buffer *buf, Lookup *lookup) {
   // declaration), so the walk above cannot see it. It still has to be emitted:
   // writeStackTraces() wrote its _key for every frame that resolved to it, and a
   // _key absent from this pool is a dangling reference in the chunk.
-  if (marked_count > 0 && lookup->_unknown_method._mark) {
+  if (lookup->_unknown_method._mark) {
     marked_count++;
   }
 
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -167,11 +167,13 @@ void Lookup::cutArguments(char *func) {
 }
 
 void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
-                                bool first_time) {
+                                bool first_time, bool& framePushed) {
   JNIEnv *jni = VM::jni();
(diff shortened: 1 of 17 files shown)


ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -363,78 +466,29 @@ void Lookup::fillJavaMethodInfo(MethodInfo *mi, jmethodID method,
...
-    if (method_name) {
-      jvmti->Deallocate((unsigned char *)method_name);
-    }
-    if (method_sig) {
-      jvmti->Deallocate((unsigned char *)method_sig);
-    }
-    if (class_name) {
-      jvmti->Deallocate((unsigned char *)class_name);
+      mi->_line_number_table = std::make_shared<SharedLineNumberTable>(
+                  line_number_table_size, line_number_table);
```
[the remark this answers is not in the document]
zhengyu123: Returned `_line_number_table` is owned by caller. I will clean up the tests mentioned above.

jbachorik:
A few nits and clarifying questions

jbachorik:
A few nits and clarifying questions

jbachorik:
A few nits and clarifying questions

jbachorik:
A few nits and clarifying questions

jbachorik:
A few nits and clarifying questions

jbachorik:
A few nits and clarifying questions

jbachorik:
A few nits and clarifying questions

jbachorik:
A few nits and clarifying questions

jbachorik:
A few nits and clarifying questions

ddprof-lib/src/main/cpp/flightRecorder.h
```diff
@@ -409,6 +447,28 @@ class Lookup {
   // increments VTABLE_RECEIVER_RESOLVE_FAILED.
   u32 resolveVTableReceiverCached(void *sym);
 
+  // The MethodMap row `frame` belongs to. Factored out so resolveMethod()'s
```
jbachorik: Test comment from magpie

jbachorik: Sorry, what kind of "row" we are talking about here? I want to be sure to understand.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -560,12 +631,98 @@ MethodInfo *Lookup::resolveMethod(ASGCT_CallFrame &frame) {
...
+  if (sigsetjmp(crash_protection_ctx, 1) != 0) {
+    // checkFault() absorbed a fault raised somewhere in fillMethod() and
+    // jumped back here, bypassing the SIGNAL_HANDLER_GUARD() destructor in
+    // segvHandler()/busHandler(); compensate for it, then disarm before
+    // touching anything else.
+    SIGNAL_HANDLER_UNWIND_AFTER_LONGJMP();
+    jmp_scope.restore();
+    Counters::increment(METHOD_RESOLVE_FAULT_RECOVERED);
+    // state.release() may fault. Unfortunately, it faults outside of profiler
+    // code where checkFault() can not absorbs.
```
jbachorik: Typo: 'absorbs' -> 'absorb'

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -560,12 +631,98 @@ MethodInfo *Lookup::resolveMethod(ASGCT_CallFrame &frame) {
...
+  // the kernel has SIGSEGV blocked, so without restoring the saved mask the
+  // signal would stay blocked and the next fault on this thread would be
+  // fatal.
+  if (sigsetjmp(crash_protection_ctx, 1) != 0) {
+    // checkFault() absorbed a fault raised somewhere in fillMethod() and
+    // jumped back here, bypassing the SIGNAL_HANDLER_GUARD() destructor in
+    // segvHandler()/busHandler(); compensate for it, then disarm before
+    // touching anything else.
+    SIGNAL_HANDLER_UNWIND_AFTER_LONGJMP();
+    jmp_scope.restore();
```
jbachorik, line 726: Also, looks like a stray leading space

jbachorik:
A few nits and clarifying questions

jbachorik:
A few nits and clarifying questions

jbachorik:
A few nits and clarifying questions

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -560,12 +631,98 @@ MethodInfo *Lookup::resolveMethod(ASGCT_CallFrame &frame) {
...
+    // Reinstates the thread's previous landing pad on every exit from this frame,
+  // including a std::bad_alloc thrown by one of the map or dictionary inserts
+  // underneath. Leaving ours installed past the end of this frame would leave
+  // checkFault() jumping into a dead stack frame.
+  JmpCtxScope jmp_scope(prof_thread);
+  ResolveMethodState state;
+
+  sigjmp_buf crash_protection_ctx;
+  // savemask must be 1: the siglongjmp originates inside segvHandler, where
+  // the kernel has SIGSEGV blocked, so without restoring the saved mask the
```
jbachorik, line 717: This looks like a weird indentation

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -100,20 +170,22 @@ void Lookup::fillNativeMethodInfo(MethodInfo *mi, const char *name,
...
-        std::string rust_demangled = RustDemangler::demangle(demangled);
+      if (RustDemangler::is_probably_rust_legacy(state._demangled)) {
+        std::string rust_demangled = RustDemangler::demangle(state._demangled);
         mi->_name = _symbols.lookup(rust_demangled.c_str());
       } else {
-        mi->_name = _symbols.lookup(demangled);
+        mi->_name = _symbols.lookup(state._demangled);
       }
-      free(demangled);
+      free(state._demangled);
```
jbachorik, line 189: Is there a chance that some other thread might observe freed `state._demangled` and still not set to `nullptr`?

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -560,12 +649,98 @@ MethodInfo *Lookup::resolveMethod(ASGCT_CallFrame &frame) {
...
+    // checkFault() absorbed a fault raised somewhere in fillMethod() and
+    // jumped back here, bypassing the SIGNAL_HANDLER_GUARD() destructor in
+    // segvHandler()/busHandler(); compensate for it, then disarm before
+    // touching anything else.
+    SIGNAL_HANDLER_UNWIND_AFTER_LONGJMP();
+    jmp_scope.restore();
+    Counters::increment(METHOD_RESOLVE_FAULT_RECOVERED);
+    // state.release() may fault. Unfortunately, it faults outside of profiler
+    // code where checkFault() can not absorbs.
+    state.release();
```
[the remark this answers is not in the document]
zhengyu123: As stated in `Crash protection for method symbolication` section, this PR does not expect JVMI calls to lead to crashes.

ddprof-lib/src/main/cpp/flightRecorder.h
zhengyu123: "row" -> "entry" 
Comment was generated by AI, I believe it mis-read `MethodMap` as an array.

ddprof-lib/src/main/cpp/flightRecorder.cpp
zhengyu123: Fixed

ddprof-lib/src/main/cpp/flightRecorder.cpp
zhengyu123: Hmmm ... looks find in my IntelliJ and VS code IDEs

ddprof-lib/src/main/cpp/flightRecorder.cpp
zhengyu123: No, `state` is on stack without any public alias

ddprof-lib/src/main/cpp/flightRecorder.cpp
zhengyu123: Line number does not seem to match. Fixed leading space on line 708.
```
-    // Reinstates the thread's previous landing pad on every exit from this frame,
+  // Reinstates the thread's previous landing pad on every exit from this frame,
```

jbachorik (approved):
AFAICS, the change does what it says and I don't see any red flags.

Are we running it somewhere with fault injection? Just to make sure the recovery is working properly?

zhengyu123:
> AFAICS, the change does what it says and I don't see any red flags.
> 
> Are we running it somewhere with fault injection? Just to make sure the recovery is working properly?

@jbachorik Yes, I injected a fault in`Lookup::fillMethod()` to test the recovery.

PR #753, opened 2026-08-25
fix(nativemem): refresh NM_NATIVE_SYMBOLS where the tables change

rkennke:
## Summary
- `NM_NATIVE_SYMBOLS` was written only by `Profiler::updateNativeLibMemStats()`, which runs solely from `Profiler::stop()`/`Profiler::dump()`. Neither fires during a recording, so the gauge read 0 for the entire life of a profiled process while the native-library symbol tables (`CodeCache`) really held real memory (~13 MB in our measurements), under-reporting the profiler's own native memory by that much at every steady-state sample point.
- Fix: refresh the gauge in `Libraries::updateSymbols()` instead — the only function that grows `_native_libs`, and a published `CodeCache` is immutable (`add()`/`expand()`/`setDwarfTable()` all assert they run before publication), so recomputing there is sufficient for the gauge to stay accurate at any later instant.

Found via a from-scratch audit of `NativeMem` counter accuracy (comparing counted bytes against a whole-process allocation ledger) done as part of a native-memory overhead investigation; coverage of the profiler's own malloc went from 37% to 96% (99% idle) after this fix, with no other counter defects found on this path.

## Test plan
- [x] `ddprof-lib:assembleRelease` builds clean
- [x] Verified against a live sweep: `NM_NATIVE_SYMBOLS` now tracks `CodeCache::memoryUsage()` correctly during an active recording instead of reading 0

### DIFF
--- a/ddprof-lib/src/main/cpp/libraries.cpp
+++ b/ddprof-lib/src/main/cpp/libraries.cpp
@@ -10,6 +10,7 @@
 #include "libraryPatcher.h"
 #include "log.h"
 #include "mallocTracer.h"
+#include "nativeMem.h"
 #include "os.h"
 #include "profiler.h"
 #include "symbols.h"
@@ -60,6 +61,19 @@ end:
 void Libraries::updateSymbols(bool kernel_symbols) {
   Symbols::parseLibraries(&_native_libs, kernel_symbols);
   LibraryPatcher::patch_libraries();
+  // Refresh the native-symbol memory gauge here, where the tables it measures
+  // actually change. This is the only function that grows _native_libs, and a
+  // published CodeCache is immutable -- add(), expand() and setDwarfTable() all
+  // assert they run before publication -- so recomputing at this point is
+  // sufficient for the gauge to be accurate at any later instant.
+  //
+  // Previously NM_NATIVE_SYMBOLS was written only by
+  // Profiler::updateNativeLibMemStats(), which runs solely from Profiler::stop()
+  // and Profiler::dump(). Neither fires during a recording, so the gauge read 0
+  // for the entire life of a profiled process while these tables really held
+  // ~10 MB, making the profiler under-report its own native memory by that much
+  // at every steady-state sample point.
+  NativeMem::setLive(NM_NATIVE_SYMBOLS, (long long)_native_libs.memoryUsage());
 }
 
 void Libraries::refresh() {
--- a/ddprof-lib/src/main/cpp/libraries.cpp
+++ b/ddprof-lib/src/main/cpp/libraries.cpp
@@ -10,7 +10,6 @@
 #include "libraryPatcher.h"
 #include "log.h"
 #include "mallocTracer.h"
-#include "nativeMem.h"
 #include "os.h"
 #include "profiler.h"
 #include "symbols.h"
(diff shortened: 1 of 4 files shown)


ddprof-lib/src/main/cpp/libraries.cpp
```diff
@@ -60,6 +61,19 @@ void Libraries::mangle(const char *name, char *buf, size_t size) {
...
+  // assert they run before publication -- so recomputing at this point is
+  // sufficient for the gauge to be accurate at any later instant.
+  //
+  // Previously NM_NATIVE_SYMBOLS was written only by
+  // Profiler::updateNativeLibMemStats(), which runs solely from Profiler::stop()
+  // and Profiler::dump(). Neither fires during a recording, so the gauge read 0
+  // for the entire life of a profiled process while these tables really held
+  // ~10 MB, making the profiler under-report its own native memory by that much
+  // at every steady-state sample point.
+  NativeMem::setLive(NM_NATIVE_SYMBOLS, (long long)_native_libs.memoryUsage());
```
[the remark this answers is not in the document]
rkennke: Confirmed via the call graph, and it's real: `updateSymbols()` now runs `_native_libs.memoryUsage()` (O(total symbols across every loaded library)) on every synchronous dlopen refresh (`profiler.cpp:933`), not just from `stop()`/`dump()` as before. That's a genuine new cost at a higher-frequency call site.

Scoped a proper fix — event-driven accounting via `NativeMem::record()` deltas in `CodeCache::add()`/`expand()`/`setDwarfTable()`, matching how every other `NM_*` category already works, with `memoryUsage()` reduced to an O(1) read of a running total. It's more involved than it first looks though: `symbols_linux.cpp` creates a `CodeCache`, parses it, and can `delete` it before publication on failure/duplicate detection, so an incremental increment needs a matching decrement in `~CodeCache()` to avoid leaking phantom accounting on the discard path — a real lifecycle change, not a one-line patch, and in code that just went through a large merge from `main`.

Given `NM_NATIVE_SYMBOLS` is a best-effort diagnostic gauge (not correctness-critical to profiling itself) and the cost is bounded by actual dlopen frequency rather than a polling interval, I'm treating this as a legitimate follow-up rather than blocking this fix on it. Happy to open a tracking issue for the incremental-accounting version if that's useful.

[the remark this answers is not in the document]
rkennke: Confirmed real, not speculative — traced every caller of `refresh()`/`updateSymbols()`. `Profiler::stop()` explicitly calls `_libs->stopRefresher()` (joining the refresher thread) before its own `refresh()`, so that path is safe. But `Profiler::dump()` (`profiler.cpp:1919`) calls `Libraries::instance()->refresh()` under `_state_lock` only — the background refresher thread's own periodic `refresh()` call never takes that lock, so the two can genuinely race: whichever `memoryUsage()`-read-then-`setLive()`-write pair finishes last wins, even if it started first with a smaller snapshot.

Same resolution as the sibling comment: the clean fix is converting this to incremental `NativeMem::record()` deltas at the mutation points instead of a periodic recompute-and-overwrite, which eliminates the read-then-write window entirely (there'd be nothing to race — each mutator atomically records its own delta). That's the same lifecycle change discussed there (needs a matching decrement in `~CodeCache()` for the pre-publication discard path in `symbols_linux.cpp`), so treating both as one follow-up rather than a quick patch here.

Practical impact in the meantime: `NM_NATIVE_SYMBOLS` is a best-effort diagnostic gauge, not something profiling correctness depends on, and it self-corrects on the next refresh (dlopen or timer tick) regardless of which write won. Not blocking this fix, but good catch — thanks.

rkennke:
Pushed a follow-up commit (\`de0c23a59\`) implementing the incremental-accounting approach discussed in both review threads above.

Summary: \`CodeCache\` now tracks a running \`_memory_usage\` total, updated by the same pre-publication mutators (constructor, \`add()\`, \`expand()\`, \`setDwarfTable()\`) that already assert they never run after publication. \`NativeMem::record()\` -- a relaxed atomic add -- fires exactly once per library, at the moment \`CodeCacheArray::add()\` publishes it. A cache discarded before publication (\`Symbols::parseLibraries\` deletes one directly on parse failure/duplicate detection) is simply never recorded, so no destructor decrement is needed. \`memoryUsage()\` itself drops from an O(symbols) loop to an O(1) field read, and the periodic \`setLive()\` overwrites in both \`Libraries::updateSymbols()\` and \`Profiler::updateNativeLibMemStats()\` are removed -- there's nothing left to race, since each publish is an independent atomic add.

Verified:
- \`NM_NATIVE_SYMBOLS\` reads the identical 13.03 MiB on the same workload as the rescan-based version (no behavior change, just how it's computed)
- Counter-audit coverage unchanged (95.5%, +0.89 MiB unaccounted, same as before this follow-up)
- \`codeCache_ut\` (release+debug), \`libraries_ut\`, \`nativeMem_ut\`, \`stackWalker_ut\` all pass

Both review threads addressed at the code level rather than just discussed -- resolving them.

ddprof-lib/src/main/cpp/libraries.cpp
```diff
@@ -60,6 +60,21 @@ void Libraries::mangle(const char *name, char *buf, size_t size) {
...
+  // stop()/dump() as before it existed at all -- see git history for why the
+  // gauge previously read 0 for a recording's entire lifetime). Two review
+  // findings on that approach: it turned every dlopen into an O(total symbols
+  // across every loaded library) rescan instead of O(new symbols only), and a
+  // dump()-triggered refresh could race the background refresher thread's own
+  // refresh() and overwrite a newer total with a stale smaller one (neither
+  // path takes a common lock). The publish-time NativeMem::record() approach
+  // has neither problem: each publish is an O(1) atomic add of that one
+  // library's already-computed total, so there is nothing to rescan and
+  // nothing to race.
```
jbachorik: Is the 'previous' version actually in git history or it was just a transient artifact?

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -1889,8 +1889,18 @@ Error Profiler::check(Arguments &args) {
 
 void Profiler::updateNativeLibMemStats() {
   // CodeCache here is the profiler's native-symbol tables (not the JVM code
-  // cache). memoryUsage() is a recomputed gauge; read it once and publish the
-  // counters plus the NativeMem gauge (NATIVE_SYMBOLS) as absolutes.
+  // cache). memoryUsage() is now an O(1) read of each library's running
```
jbachorik, line 1892: Not necessary to 'pin' the current implementation vs. some previous one ...

```suggestion
  // cache). memoryUsage() is an O(1) read of each library's running
```

jbachorik (approved):
I have two things related to the generated comments to take a look at and perhaps adjust before merge.

ddprof-lib/src/main/cpp/libraries.cpp
rkennke: It's in the git history of this PR, but not mainline. I'll reword it so that it does not refer to history that does not actually exist.

PR #754, opened 2026-08-25
feat(nativemem): emit post-serialization counters so flush cost is visible

rkennke:
## Summary
- A JFR chunk's `NM_*` counters are sampled *before* `writeCpool()` runs in that same chunk, so the in-chunk values can never reflect what serialization itself costs (e.g. dictionary/method-map growth during constant-pool writing). There was previously no way to observe this cost from a live JFR recording at all.
- Adds `Recording::capturePostFlushNativeMem()`, called immediately after `writeCpool()`'s patching completes (so it captures state *after* serialization, before `chunk_end`). Snapshots per-category live/max bytes into `_post_flush_live[]`/`_post_flush_max[]`, which the *following* chunk emits as `native_mem_post_flush_live_bytes.<category>` / `native_mem_post_flush_max_bytes.<category>` counter events, alongside a refresh of the `Counters::` JNI-visible mirrors.
- Deliberately does not call `NativeMem::sample()` a second time per chunk — that would advance a 64-tick moving-average window, silently redefining `avg()` as a 32-chunk mean.
- Absent on the first chunk (no prior flush to report); every subsequent chunk carries the previous chunk's post-serialization snapshot.

Motivated by a native-memory overhead investigation that found a chunk-flush burst (dictionary/method-map growth during constant-pool serialization) invisible to any existing counter — the spike shows up in the per-category `max` counters of the chunk *after* it happens, once this lands.

## Test plan
- [x] `ddprof-lib:assembleRelease` builds clean
- [x] Verified via a live multi-chunk JFR recording: `native_mem_post_flush_*` counters appear from the second chunk onward and reflect the prior chunk's post-`writeCpool()` state

### DIFF
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -691,7 +691,10 @@ char *Recording::_jvm_flags = NULL;
 char *Recording::_java_command = NULL;
 
 Recording::Recording(int fd, Arguments &args)
-    : _fd(fd), _method_map() {
+    : _fd(fd), _method_map(), _has_post_flush(false) {
+
+  memset(_post_flush_live, 0, sizeof(_post_flush_live));
+  memset(_post_flush_max, 0, sizeof(_post_flush_max));
 
   args.save(_args);
   _chunk_start = lseek(_fd, 0, SEEK_END);
@@ -833,6 +836,12 @@ off_t Recording::finishChunk(bool end_recording, bool do_cleanup) {
   _buf->put8(0, (char)pool_count);
   result = pwrite(_fd, _buf->data(), 1, cpool_offset + count_offset_in_cpool);
   (void)result;
+
+  // Serialization is complete: the method map is built and the dictionary has
+  // grown. Capture that state now for the next chunk to emit, and refresh the
+  // JNI-visible counter mirrors so a live process reading getDebugCounters0()
+  // after a dump() sees post-serialization values rather than pre-.
+  capturePostFlushNativeMem();
 
   off_t chunk_end = lseek(_fd, 0, SEEK_CUR);
 
@@ -1809,6 +1818,20 @@ void Recording::writeLogLevels(Buffer *buf) {
   }
 }
 
+void Recording::capturePostFlushNativeMem() {
+  for (int c = 0; c < NM_NUM_CATEGORIES; c++) {
+    NativeMemCategory cat = (NativeMemCategory)c;
+    _post_flush_live[c] = NativeMem::live(cat);
+    _post_flush_max[c] = NativeMem::max(cat);
+  }
+  _has_post_flush = true;
+  // Deliberately NOT NativeMem::sample(): that advances a 64-tick moving
+  // average window, so calling it a second time per chunk would silently
+  // redefine avg() as a 32-chunk mean.
+  Counters::set(NATIVE_MEM_LIVE_BYTES, NativeMem::liveTotal());
+  Counters::set(NATIVE_MEM_MAX_BYTES, NativeMem::maxTotal());
+}
+
 void Recording::updateNativeMemStats() {
   // Refresh the moving-window averages and the observed total peak. Per-category
   // peaks are maintained precisely at allocation time, so they are not sampled
@@ -1863,6 +1886,21 @@ void Recording::writeNativeMem(Buffer *buf) {
       char label[64];
       snprintf(label, sizeof(label), "%s%s", m.prefix, name);
       emit(label, m.value);
+    }
+  }
+
+  // State immediately after the PREVIOUS chunk's writeCpool(), which is the
+  // only way to see what serialization itself costs -- the in-chunk values
+  // above are necessarily sampled before it runs. Absent on the first chunk,
+  // since no flush has happened yet.
+  if (_has_post_flush) {
+    for (int c = 0; c < NM_NUM_CATEGORIES; c++) {
+      const char *name = NativeMem::categoryName((NativeMemCategory)c);
+      char label[64];
+      snprintf(label, sizeof(label), "native_mem_post_flush_live_bytes.%s", name);
+      emit(label, _post_flush_live[c]);
+      snprintf(label, sizeof(label), "native_mem_post_flush_max_bytes.%s", name);
+      emit(label, _post_flush_max[c]);
     }
   }
 
--- a/ddprof-lib/src/main/cpp/flightRecorder.cpp
+++ b/ddprof-lib/src/main/cpp/flightRecorder.cpp
@@ -1024,12 +1024,6 @@ off_t Recording::finishChunk(bool end_recording, bool do_cleanup) {
   result = pwrite(_fd, _buf->data(), 1, cpool_offset + count_offset_in_cpool);
   (void)result;
 
-  // Serialization is complete: the method map is built and the dictionary has
-  // grown. Capture that state now for the next chunk to emit, and refresh the
-  // JNI-visible counter mirrors so a live process reading getDebugCounters0()
-  // after a dump() sees post-serialization values rather than pre-.
-  capturePostFlushNativeMem();
-
   off_t chunk_end = lseek(_fd, 0, SEEK_CUR);
 
   // // Workaround for JDK-8191415: compute actual TSC frequency, in case JFR is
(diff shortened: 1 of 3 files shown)


ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -2070,6 +2093,21 @@ void Recording::writeNativeMem(Buffer *buf) {
     }
   }
 
+  // State immediately after the PREVIOUS chunk's writeCpool(), which is the
+  // only way to see what serialization itself costs -- the in-chunk values
+  // above are necessarily sampled before it runs. Absent on the first chunk,
+  // since no flush has happened yet.
+  if (_has_post_flush) {
```
rkennke, line 2108: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** No test covers the new post-flush snapshot: neither the first-chunk-absent case gated by _has_post_flush, nor the later-chunk-present case, nor the deliberate 'do not call NativeMem::sample() twice' invariant (which is documented only in a comment and would be silently violated by a future refactor, redefining avg() as a 32-chunk mean). The new counter labels native_mem_post_flush_live_bytes.* / native_mem_post_flush_max_bytes.* have zero references anywhere else in the repo, and the stated verification is a manual live JFR recording.

**Suggestion:** Extend NativeMemAccountingTest (which already provides the getRecordedCounterValue harness over datadog.ProfilerCounter) with a case that performs an explicit dump() to force a chunk switch, asserts native_mem_post_flush_live_bytes.<category> is absent in the first chunk and present from the second onward — pinning both the absent-on-first-chunk invariant and the label spelling — and add a gtest asserting the moving-average window advances exactly once per finishChunk().

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -2013,6 +2022,20 @@ void Recording::writeLogLevels(Buffer *buf) {
...
+  for (int c = 0; c < NM_NUM_CATEGORIES; c++) {
+    NativeMemCategory cat = (NativeMemCategory)c;
+    _post_flush_live[c] = NativeMem::live(cat);
+    _post_flush_max[c] = NativeMem::max(cat);
+  }
+  _has_post_flush = true;
+  // Deliberately NOT NativeMem::sample(): that advances a 64-tick moving
+  // average window, so calling it a second time per chunk would silently
+  // redefine avg() as a 32-chunk mean.
+  Counters::set(NATIVE_MEM_LIVE_BYTES, NativeMem::liveTotal());
```
rkennke, line 2042: 🗿 🤖 🔴

**[Sphinx Review — LOW]** capturePostFlushNativeMem() refreshes only the NATIVE_MEM_LIVE_BYTES and NATIVE_MEM_MAX_BYTES JNI mirrors, leaving NATIVE_MEM_AVG_BYTES at its pre-serialization value from updateNativeMemStats(). The call-site comment at line 1029 claims the refresh lets 'a live process reading getDebugCounters0() after a dump() see post-serialization values rather than pre-', which holds for only two of the three related counters, so the JNI-visible triple mixes post-flush live/max with a pre-flush avg and the max >= avg relation asserted elsewhere is no longer guaranteed for that mirror snapshot.

**Suggestion:** Either also set NATIVE_MEM_AVG_BYTES (to NativeMem::avgTotal(), unchanged since sample() is deliberately not re-run) so the mirror stays a coherent snapshot, or note in the comment — and at the call site at line 1029 — that avg intentionally stays at its sampled-tick value because avg() cannot be recomputed without advancing the 64-tick window, so the three mirrors are not from the same instant.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -1021,6 +1024,12 @@ off_t Recording::finishChunk(bool end_recording, bool do_cleanup) {
   result = pwrite(_fd, _buf->data(), 1, cpool_offset + count_offset_in_cpool);
   (void)result;
 
+  // Serialization is complete: the method map is built and the dictionary has
+  // grown. Capture that state now for the next chunk to emit, and refresh the
+  // JNI-visible counter mirrors so a live process reading getDebugCounters0()
+  // after a dump() sees post-serialization values rather than pre-.
+  capturePostFlushNativeMem();
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** capturePostFlushNativeMem() is invoked before cleanupUnreferencedMethods() (line 1065), which runs later in the same finishChunk() and frees method-map and line-table allocations. On the do_cleanup=true path (switchChunk/dump) the captured per-category live values therefore include memory reclaimed microseconds later, so native_mem_post_flush_live_bytes.method_map / .line_tables overstate real post-flush residency. The other caller, finishChunk() at line 935, passes do_cleanup=false, so the same counter name means 'retained after serialization' there and 'peak including memory about to be freed' on the dump path — a consumer diffing the series across chunks cannot separate steady-state growth from transient serialization cost.

**Suggestion:** Either move the capture to after the cleanupUnreferencedMethods() block (still before the return, so it still reflects post-serialization state) so the snapshot consistently means 'retained after serialization and cleanup', or state explicitly in both the call-site comment and the header comment that the snapshot is deliberately pre-cleanup and represents the serialization peak rather than steady-state residency.

ddprof-lib/src/main/cpp/flightRecorder.h
```diff
@@ -344,6 +345,25 @@ class Recording {
...
+  // chunk header records cpool_offset as the boundary between the event section
+  // and the constant pool, so no event may be appended once writeCpool() has
+  // run. But writeCpool() is where the method map is built and the dictionary
+  // grows -- by two orders of magnitude in a large recording -- so a consumer
+  // reading only the in-chunk values never sees the cost of serialization.
+  //
+  // native_mem_max_bytes does eventually reflect it, because record() raises
+  // the peak at allocation time and nothing ever resets it, but only one chunk
+  // late and only as a lifetime maximum: after several flushes it can no longer
+  // say which flush was responsible. These snapshots give the per-flush figure.
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** The rationale comment states these snapshots 'give the per-flush figure', but _post_flush_max is NativeMem::max(), a lifetime monotonic high-water mark that nothing resets in production (NativeMem::reset() has callers only in nativeMem_ut.cpp and linearAllocator_nativemem_ut.cpp). The post-flush max series therefore has exactly the attribution defect the comment ascribes to native_mem_max_bytes; a per-flush peak is only obtainable by differencing against the previous chunk's in-chunk max. Only the live series is genuinely new information.

**Suggestion:** Reword the comment to state that the post-flush live values are the genuinely new signal, and that a per-flush peak delta must be computed as post_flush_max[N] minus the in-chunk native_mem_max_bytes emitted in chunk N.

ddprof-lib/src/main/cpp/flightRecorder.h
```diff
@@ -344,6 +345,25 @@ class Recording {
...
+  bool _has_post_flush;
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — LOW]** New instance state (_has_post_flush, _post_flush_live, _post_flush_max) is declared in Recording's public section, whereas every other data member of the class lives in the private block at lines 208-252. capturePostFlushNativeMem() is likewise only called from within Recording, and Profiler and Lookup are already friends, so neither the state nor the method needs public visibility.

**Suggestion:** Move the three data members, and capturePostFlushNativeMem() which has no external callers, into the private section alongside the other Recording state.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -1866,6 +1889,21 @@ void Recording::writeNativeMem(Buffer *buf) {
     }
   }
 
+  // State immediately after the PREVIOUS chunk's writeCpool(), which is the
+  // only way to see what serialization itself costs -- the in-chunk values
+  // above are necessarily sampled before it runs. Absent on the first chunk,
+  // since no flush has happened yet.
+  if (_has_post_flush) {
```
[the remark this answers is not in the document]
rkennke: Fixed in 73b3bd65b.

ddprof-lib/src/main/cpp/flightRecorder.cpp
rkennke: Fixed in 73b3bd65b.

ddprof-lib/src/main/cpp/flightRecorder.cpp
rkennke: Fixed in 73b3bd65b.

ddprof-lib/src/main/cpp/flightRecorder.cpp
rkennke: Fixed in 73b3bd65b.

ddprof-lib/src/main/cpp/flightRecorder.h
rkennke: Fixed in 73b3bd65b.

ddprof-lib/src/main/cpp/flightRecorder.h
rkennke: Fixed in 73b3bd65b.

PR #755, opened 2026-08-25
fix(nativemem): skip building DWARF unwind tables when cstack mode doesn't need them

rkennke:
## Summary
- `ElfParser::parseDwarfInfo()` builds SFrame/DWARF unwind tables for every parsed native library unconditionally, even though they're read only by `StackWalker::walkDwarf()` under `CSTACK_DWARF`. Measured at 4.52 MiB / 22 allocations at 150,000 loaded classes, paid even when the resolved mode is `CSTACK_VM` (the common default whenever HotSpot VMStructs are available).
- Adds a gate: skip building once we know for certain the mode isn't `CSTACK_DWARF`. Getting "certain" right requires a dedicated `Profiler::_cstack_resolved` flag, not just comparing `cstackMode()` — `_cstack`'s constructor default is `CSTACK_NO`, a real user-selectable mode, not a distinguishable "not yet decided" sentinel. Since libraries are parsed exactly once (`Symbols::parseLibraries` tracks `_parsed_inodes`), including the handful parsed at JVMTI `Agent_OnLoad` before `Profiler::start()` ever resolves cstack mode, a naive check would permanently and silently skip building the table for those libraries (`libjvm.so` among them) even in an actual `CSTACK_DWARF` session — degrading native-frame unwind quality with no visible error. `_cstack_resolved` is set once, right after `start()`'s resolution logic, before any sampling begins.

## Honest note on measured impact — why this is closed rather than pursued further right now

The corrected fix is safe (verified `cstack=dwarf` still builds the full table, no regression). But re-measuring after fixing the above turned up something worth flagging: on the benchmark workload used to find this (`classes` mode, 150,000 loaded classes), the *real* savings are only ~0.07 MiB, not the ~4.52 MiB originally measured with the buggy gate. Almost the entire native-library set for that workload loads before `Profiler::start()` runs, where the gate must conservatively always build (the eventual mode isn't knowable yet). Only a handful of JDK-internal libraries that load lazily after start (`libmanagement.so`, `libnet.so`, `libnio.so`, etc.) actually benefit.

This fix should have more value for workloads that `dlopen` native libraries *while actively profiling* in a non-DWARF mode (JNI-heavy applications, dynamically-attached native agents) — but that scenario isn't covered by any workload we've measured, so the benefit there is currently unquantified, not just small. Given the added complexity (a new `Profiler` field threaded through a lifecycle-ordering invariant) isn't clearly justified by a measured benefit at this point, closing this without merging. Leaving it open here as reference in case someone has a workload where this matters and wants to pick it up — the fix itself is complete and tested.

## Test plan
- [x] `stackWalker_ut`, `libraries_ut`, `codeCache_ut` (release+debug) all pass
- [x] Verified `cstack=dwarf` builds the full table (`NM_NATIVE_SYMBOLS` = 13.03 MiB, matching the ungated value exactly — no regression)
- [x] Verified `cstack=fp` correctly skips the ~0.07 MiB that's actually skippable for that workload

### DIFF
--- a/ddprof-lib/src/main/cpp/profiler.cpp
+++ b/ddprof-lib/src/main/cpp/profiler.cpp
@@ -1609,6 +1609,12 @@ Error Profiler::start(Arguments &args, bool reset) {
   }
 
   args._cstack = _cstack;
+  // From here on, cstackMode() reflects this session's final, resolved mode.
+  // Anything parsed before this point (e.g. ElfParser::parseDwarfInfo() for
+  // the handful of libraries loaded at JVMTI Agent_OnLoad, before this
+  // start() call) must not trust cstackMode() yet -- see cstackResolved()'s
+  // declaration for why.
+  _cstack_resolved.store(true, std::memory_order_release);
   // Prepare JVMSupport for execution
   JVMSupport::initExecution(args, VM::jvmti(), VM::jni());
 
--- a/ddprof-lib/src/main/cpp/profiler.h
+++ b/ddprof-lib/src/main/cpp/profiler.h
@@ -140,6 +140,15 @@ private:
   StackWalkFeatures _features;
   int _safe_mode;
   CStack _cstack;
+  // Set once, immediately after start() resolves _cstack (CSTACK_DEFAULT ->
+  // a concrete mode) and before any sampling begins. _cstack's constructor
+  // default is CSTACK_NO -- a real, user-selectable mode, not a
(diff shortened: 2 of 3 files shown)


rkennke:
Closing without merging per the honest cost/benefit note above: the corrected fix is real and tested, but the measured benefit on the only workload we tested it against is ~0.07 MiB, not worth the added complexity right now. Left open as a reference/starting point for whoever has a workload where dynamically-loaded native libraries during active non-DWARF profiling actually matter.

PR #756, opened 2026-08-25
refactor(profiling): Signal handler boilerplate

yaronguro-datadog:
**What does this PR do?**:

Extracts the boilerplate duplicated across the CPU/wall profiling signal
handlers (`CTimer::signalHandler`, `CTimerJvmti::signalHandler`,
`ITimerJvmti::signalHandler`, `WallClockASGCT::signalHandler`,
`WallClockJvmti::signalHandler`) into two shared helpers, fixes several
latent errno-restore gaps found along the way, and extends the same
treatment to two handlers outside PROF-14748's original five
(`ITimer::signalHandler`, `PerfEvents::signalHandler`):

- Adds `SighandlerTidScope` (`guards.h`), a narrowly-scoped RAII guard around
  `Shims::instance().setSighandlerTid(tid)` / `setSighandlerTid(-1)`,
  replacing the manual set/reset pairs in all seven handlers.
- Adds `tickInitWindowIfNeeded()` (`threadLocalData.inline.h`), a shared
  helper for the 4-line init-window guard that was copied verbatim into the
  original five handlers. It lives there because it bridges `JVMThread::current()`
  and `ProfiledThread::inInitWindow()`/`tickInitWindow()`, and that file already
  owns `ProfiledThread`'s inline members; the new `#include "jvmThread.h"` adds
  no external headers, since `threadLocalData.h` already pulled `<jvmti.h>`
  (which includes `jni.h`) and `threadLocal.h`. Split into
  `tickInitWindowIfNeededImpl(bool has_jvm_thread, ProfiledThread*)` plus a
  thin wrapper that derives the bool, so the branch logic is unit testable
  without a JVM (`JVMThread::current()` asserts on an invalid pthread key). Deliberately *not* applied to `ITimer::signalHandler` or
  `PerfEvents::signalHandler` (see Additional Notes).
- Fixes a pre-existing errno-restore gap in `CTimer::signalHandler`'s
  `!_enabled` early return, which returned without restoring `errno` (the
  three Jvmti handlers already did this correctly on the equivalent path).
- Adds missing `saved_errno` save/restore to `WallClockASGCT::signalHandler`,
  which previously didn't save/restore errno at all, unlike its sibling
  `WallClockJvmti::signalHandler`.
- Fixes a second, narrower errno-restore gap in `WallClockJvmti::signalHandler`:
  `saved_errno` was captured *after* the `CriticalSection` entry check, so the
  `!cs.entered()` bail-out path left `errno` unrestored. The save now happens
  first, matching `WallClockASGCT`.
- Adds `saved_errno` save/restore to `ITimer::signalHandler` and
  `PerfEvents::signalHandler`, neither of which had it before, and wraps their
  `recordSample` span in `SighandlerTidScope` (replacing the manual
  `setSighandlerTid`/`-1` pair). `assert(current != nullptr)` ahead of the
  `tid` computation is added to `ITimer::signalHandler` only, mirroring the
  original five handlers -- `PerfEvents::signalHandler` already had it.
  `ITimer::signalHandler` also gains a direct `#include <cassert>`: the new
  assert would otherwise have compiled only via the `jvmThread.h` ->
  `threadLocal.h` chain this PR introduced, so trimming an include would
  have broken the build. `ITimerJvmti::signalHandler` was already fully
  migrated and needed no changes here.
- Fixes a latent errno-clobbering bug in `PerfFdRearmGuard::~PerfFdRearmGuard()`
  (`perfEvents_linux.cpp`): its `ioctl()`/`resetBuffer()` calls could clobber
  `errno` on every exit path, including the drop paths, and nothing restored
  it afterwards. Fixed by declaring `ErrnoPreserver` in
  `PerfEvents::signalHandler` ahead of `PerfFdRearmGuard`, so it destructs
  *last* — after `PerfFdRearmGuard` — and restores whatever `PerfFdRearmGuard`'s
  side effects clobbered. The guard itself is deliberately not self-contained:
  an earlier review pass asked for the nested `ErrnoPreserver` inside
  `~PerfFdRearmGuard()` to be removed as redundant, and the handler-level one
  (declared just below the `si_code <= 0` external-signal fast bail, so that
  path stays untouched) covers its only construction site.

**Motivation**:

[PROF-14748](https://datadoghq.atlassian.net/browse/PROF-14748?actionerId=5d4b47deba36300d23e5053d&sourceType=assign) : these five handlers were identified during review of
the jvmtistacks addition as sharing identical boilerplate with no shared
abstraction.

**Additional Notes**:

- `ITimer::signalHandler` and `PerfEvents::signalHandler` were not named in PROF-14748's
  original five. Both had the identical manual
  set/reset-`setSighandlerTid` pattern as their siblings and are now given
  the same `SighandlerTidScope` treatment, plus errno save/restore and a
  `current != nullptr` assert.
- `tickInitWindowIfNeeded()` was deliberately *not* wired into ITimer nor
  PerfEvents handlers, unlike the original five. The reason is simply that
  neither engine ever had this check: the init-window guard closes the race
  between `Profiler::registerThread()` and `thread_native_entry` publishing
  the JVM thread TLS, which is a property of the sampled thread and unrelated
  to signal-origin validation (an earlier revision of this description claimed
  otherwise — that rationale was wrong). Adding it here would be a behaviour
  change to two engines in a PR that is otherwise a boilerplate extraction, so
  it is left for separate review. A comment at each site now records that.
- `PerfEvents::signalHandler`'s existing control flow (the `_enabled` check
  gates only the `SighandlerTidScope`/`recordSample` block, after
  `noteCPUSample` already ran unconditionally) was left as-is; only errno
  handling and the `SighandlerTidScope`/assert changes were added.
- `resolveThreadId` (a candidate shared helper for the
  `current ? current->tid() : OS::threadId()` duplication) was not extracted.
  The underlying duplication was eliminated directly instead: the original
  five handlers now assert `current != nullptr` before computing `tid`,
  making the ternary dead code, so it was removed rather than factored out.
  (`ITimer::signalHandler` and `PerfEvents::signalHandler` already computed
  `tid` from `current->tid()` directly, with no ternary to remove.) The
  pattern still exists in `javaApi.cpp`, a non-signal-handler context outside
  this ticket's scope.
- No change to signal-origin validation, foreign-signal forwarding, or
  critical-section semantics in any handler.

**How to test the change?**:

- `./gradlew :ddprof-lib:compileDebug` compiles cleanly on macOS and linux.
- The changes are mechanical (RAII scope-exit timing, errno save/restore
  ordering) and don't alter sampling logic or control flow beyond exit-path
  cleanup. `tickInitWindowIfNeeded()` is not wired into `ITimer`/`PerfEvents`
  (see Additional Notes), so no runtime sampling behavior changes for those
  two engines.
- Adds `ddprof-lib/src/test/cpp/guards_ut.cpp`: three unit tests over
  `ErrnoPreserver` -- restore on normal scope exit, restore on an early
  return, and the ordering invariant the handlers rest on (declared first, it
  destructs last and wins over a later guard whose destructor clobbers
  `errno`). `SighandlerTidScope` is not covered: reading the sighandler tid
  back needs the `Shims` symbol a JVM provides, and this binary has none.
- Adds `ddprof-lib/src/test/cpp/tickInitWindow_ut.cpp`: four unit tests over
  `tickInitWindowIfNeededImpl()` covering all four
  `has_jvm_thread` x `inInitWindow()` combinations, including the case that
  would break if `&&` were mutated to `||`, plus the one-shot property (a
  second call must not tick again). The one-line derivation in
  `tickInitWindowIfNeeded()` itself (`JVMThread::current() != nullptr`) is
  deliberately *not* unit tested: `JVMThread::current()` reads a pthread key
  established by JVM startup and this gtest binary has no JVM attached, so
  covering it means mutating process-global state and leaning on
  `ThreadLocal<JVMThread*>::initialize()`'s scan over every live pthread key
  -- more cost and fragility than a single argument expression warrants. No
  integration test exercises the init-window path either, so that one line
  rests on compilation and inspection.
- Beyond the new unit tests, existing CPU/wall sampler
  correctness and signal-handler integration tests should be run to confirm
  no regression, particularly around `CTimer` and `WallClockJvmti` given the
  errno-ordering change in the latter, and around `ITimer`/`PerfEvents`
  given the new `SighandlerTidScope`/assert/errno-restore paths.

[PROF-14748]: https://datadoghq.atlassian.net/browse/PROF-14748?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jbachorik:
One small typo to fix, otherwise looks good!

ddprof-lib/src/main/cpp/jvmThread.h
```diff
@@ -53,4 +54,20 @@ class JVMThread {
     static void* currentThreadSlow();
 };
 
+// Shared init-window guard used by the CPU/wall profiling signal handlers.
+// Guards against the race window between Profiler::registerThread() and
+// thread_native_entry setting JVM TLS (PROF-13072): a pure native thread
+// (where JVMThread::current() is always null) is allowed through once its
+// one-shot init window has ticked down. Returns true iff the caller should
```
jbachorik:
```suggestion
// one-shot init window has ticked down. Returns true if the caller should
```

zhengyu123:
Good cleanup.

For preserving `errno`, probably can use a RAII class, e.g.
```
class ErrnoPreserver {
private:
  int _errno;
public:
   ErrnoPreserver(const ErrnoPreserver&) = delete;
   ErrnoPreserver& operator=(const ErrnoPreserver&) = delete;

  ErrnoPreserver() : _errno(errno) { }
  ~ErrnoPreserve() { errno = _errno};
};
 
```

zhengyu123 (approved):
LGTM

ddprof-lib/src/main/cpp/wallClock.cpp
```diff
@@ -232,18 +232,14 @@ void WallClockASGCT::sharedSignalHandler(int signo, siginfo_t *siginfo,
 
 void WallClockASGCT::signalHandler(int signo, siginfo_t *siginfo, void *ucontext,
                               u64 last_sample, ProfiledThread* current) {
+  ErrnoPreserver errno_preserver;
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** For the wall-clock engines the ErrnoPreserver is declared in the inner WallClockASGCT/WallClockJvmti::signalHandler helper rather than at the top of sharedSignalHandler, which is the actual installed sa_sigaction. Everything between signal entry and this line - Counters::increment, SIGNAL_HANDLER_GUARD_OR_DROP() -> SignalHandlerScope -> ProfiledThread::acquireCurrent() -> SignalBlocker -> pthread_sigmask, and BaseWallClock::eventsEnabled() - runs before errno is captured, so a clobbered errno is snapshotted and restored, leaving the interrupted user code with the wrong value. Symmetrically, sharedSignalHandler's SignalHandlerScope/InflightGuard destructors run after the inner restore. This is the same gap guards.h documents ('declare the ErrnoPreserver as the *first* local in the guarded function') and that the other five handlers do avoid, since there ErrnoPreserver is the first statement of the sa_sigaction itself.

**Suggestion:** Move the ErrnoPreserver declaration to be the first local of WallClockASGCT::sharedSignalHandler (wallClock.cpp:206) and WallClockJvmti::sharedSignalHandler (wallClock.cpp:412), and drop it from the inner signalHandler helpers. That makes it the first local of the real signal-entry frame, matching the CTimer/ITimer/PerfEvents handlers and the contract stated in guards.h.

ddprof-lib/src/main/cpp/jvmThread.h
```diff
@@ -53,4 +54,20 @@ class JVMThread {
     static void* currentThreadSlow();
 };
 
+// Shared init-window guard used by the CPU/wall profiling signal handlers.
+// Guards against the race window between Profiler::registerThread() and
+// thread_native_entry setting JVM TLS (PROF-13072): a pure native thread
+// (where JVMThread::current() is always null) is allowed through once its
+// one-shot init window has ticked down. Returns true if the caller should
+// tick-and-return, in which case the tick has already happened; the caller
+// remains responsible for restoring errno at its own return, since not all
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** The new doc comment states that callers of tickInitWindowIfNeeded() must restore errno at their own return 'since not all call sites save errno the same way'. That is untrue of every call site created in this diff: all five (ctimer_linux.cpp:231/281, itimer.cpp:113, wallClock.cpp:242/449) now save errno identically via ErrnoPreserver and none performs a manual restore. The comment documents the very pattern this PR removed.

**Suggestion:** Replace the errno sentence with a statement of the actual convention, e.g. that all call sites are signal handlers holding an ErrnoPreserver and therefore need no manual errno handling on this return path.

ddprof-lib/src/main/cpp/jvmThread.h
```diff
@@ -53,4 +54,20 @@ class JVMThread {
...
+// call sites save errno the same way.
+static inline bool tickInitWindowIfNeeded(ProfiledThread* current) {
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** No test in this PR would detect changing && to || in the init-window condition of tickInitWindowIfNeeded(), which would make the function return true if either condition (JVMThread::current() == nullptr OR current->inInitWindow()) holds, instead of requiring both.

**Suggestion:** Add a test case that calls tickInitWindowIfNeeded() with different combinations of conditions and verifies the return value and whether tickInitWindow() is called in each case (null/in-window, null/not-in-window, not-null/in-window, not-null/not-in-window).

ddprof-lib/src/main/cpp/perfEvents_linux.cpp
```diff
@@ -764,6 +764,8 @@ class PerfFdRearmGuard {
 public:
   PerfFdRearmGuard(int fd, int tid) : _fd(fd), _tid(tid) {}
   ~PerfFdRearmGuard() {
+    // These calls must not leak an errno change to the caller.
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — LOW]** The ErrnoPreserver inside ~PerfFdRearmGuard is redundant. The PR description justifies it by claiming PerfFdRearmGuard 'is the first local constructed in PerfEvents::signalHandler, so it destructs last - after any errno = saved_errno restore in the handler body'. The same diff makes ErrnoPreserver the first local (line 782), so PerfFdRearmGuard now destructs strictly before the errno restore and cannot clobber it. PerfFdRearmGuard has a single instantiation, so there is no other caller the extra guard protects.

**Suggestion:** Drop the nested ErrnoPreserver (and its comment) from ~PerfFdRearmGuard, relying on the handler-level ErrnoPreserver; or, if the belt-and-braces guard is intentional, correct the PR description since its stated rationale no longer holds.

ddprof-lib/src/main/cpp/ctimer_linux.cpp
```diff
@@ -206,56 +206,50 @@ Error CTimerJvmti::start(Arguments &args) {
...
   CriticalSection cs(current);
   if (!cs.entered()) {
-    errno = saved_errno;
     return;
   }
   if (!__atomic_load_n(&_enabled, __ATOMIC_ACQUIRE)) {
-    errno = saved_errno;
     return;
   }
   int tid = 0;
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — LOW]** `int tid = 0;` is declared before the init-window early return and then unconditionally reassigned at line 236 without ever being read. The refactor removed the reason for the split declaration; CTimer::signalHandler just below (line 285) already uses `int tid = current->tid();`.

**Suggestion:** Delete line 229 and change line 236 to `int tid = current->tid();`, matching CTimer::signalHandler.

ddprof-lib/src/main/cpp/itimer.cpp
```diff
@@ -26,13 +26,15 @@
 #include "threadLocalData.inline.h"
 #include "threadState.inline.h"
 #include "guards.h"
+#include <errno.h>
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — LOW]** `#include <errno.h>` was added to itimer.cpp, but the diff simultaneously removed every direct use of `errno` from that file (only ErrnoPreserver remains, whose definition in guards.h already includes <cerrno>). The include is unnecessary.

**Suggestion:** Remove the `#include <errno.h>` line; guards.h already provides <cerrno> for ErrnoPreserver.

ddprof-lib/src/main/cpp/os_linux.cpp
```diff
@@ -474,16 +475,14 @@ void OS::forwardForeignSignal(int signo, siginfo_t* siginfo, void* ucontext) {
     // chained handler) may set errno. Callers that save errno AFTER
     // forwardForeignSignal (e.g. CTimer::signalHandler) would see a clobbered
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — LOW]** The rationale comment above the new ErrnoPreserver names CTimer::signalHandler as a caller that 'saves errno AFTER forwardForeignSignal'. It does not - ctimer_linux.cpp:252 declares ErrnoPreserver as the handler's first statement, before the forwardForeignSignal call, as does CTimerJvmti at line 209. The real justification for the guard is that the interrupted code must not observe an errno clobbered by rt_sigprocmask or the chained handler.

**Suggestion:** Restate the rationale in terms of not leaking errno changes from the chained handler / rt_sigprocmask to the interrupted code, and drop the now-incorrect CTimer::signalHandler example.

ddprof-lib/src/main/cpp/jvmThread.h
```diff
@@ -10,6 +10,7 @@
 #include <jvmti.h>
 
 #include "threadLocal.h"
+#include "threadLocalData.h"
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — LOW]** Adding #include "threadLocalData.h" to jvmThread.h drags a heavy header (which transitively pulls context.h, otel_context.h, os.h, unwindStats.h, <atomic>, <vector>, jvmti.h) into all 17 TUs that include jvmThread.h, and transitively into everything including hotspot/vmStructs.h, solely to support a 5-line helper used by 5 call sites that all already include threadLocalData.h themselves.

**Suggestion:** Consider putting tickInitWindowIfNeeded in a lighter, purpose-specific header (or in threadLocalData.inline.h, which already has ProfiledThread available), keeping jvmThread.h's forward-declaration-only footprint.

ddprof-lib/src/main/cpp/ctimer_linux.cpp
```diff
@@ -206,56 +205,49 @@ Error CTimerJvmti::start(Arguments &args) {
...
-  event._execution_mode = getThreadExecutionMode();
-  // Opted into JVMTI delegation; drop the sample if the JVM rejects the
-  // request (WRONG_PHASE if JFR is not recording, NOT_AVAILABLE if
-  // jdk.StackTraceRequest is disabled). recordSampleDelegated() bumps the
-  // failure counters; there is no fallback to ASGCT in this engine.
-  Profiler::instance()->recordSampleDelegated(ucontext, _interval, tid,
-                                               BCI_CPU, &event);
-  Shims::instance().setSighandlerTid(-1);
-  errno = saved_errno;
+  int tid = current->tid();
```
rkennke: very minor nit: this can also go into the scope below.

ddprof-lib/src/main/cpp/ctimer_linux.cpp
```diff
@@ -284,26 +276,19 @@ void CTimer::signalHandler(int signo, siginfo_t *siginfo, void *ucontext) {
...
-  // per thread. Pure native threads (where JVMThread::current() is always null)
-  // are allowed through once the one-shot window expires.
-  if (JVMThread::current() == nullptr && current->inInitWindow()) {
-    current->tickInitWindow();
-    errno = saved_errno;
+  if (tickInitWindowIfNeeded(current)) {
     return;
   }
   current->noteCPUSample(Profiler::instance()->recordingEpoch());
   int tid = current->tid();
```
rkennke: Same here -> move statement into scope.

ddprof-lib/src/main/cpp/itimer.cpp
```diff
@@ -53,13 +56,14 @@ void ITimer::signalHandler(int signo, siginfo_t *siginfo, void *ucontext) {
   }
   current->noteCPUSample(Profiler::instance()->recordingEpoch());
   int tid = current->tid();
-  Shims::instance().setSighandlerTid(tid);
```
rkennke, line 56: Same here, you can move tid inside the new scope.

ddprof-lib/src/main/cpp/itimer.cpp
```diff
@@ -102,40 +106,35 @@ bool ITimerJvmti::_enabled = false;
...
-  ExecutionEvent event;
-  event._execution_mode = getThreadExecutionMode();
-  // setitimer(ITIMER_PROF) delivers SIGPROF to an arbitrary thread chosen by
-  // the OS, so ucontext may be from a JVM-internal thread.  Pass nullptr to
-  // force the JVM into safepoint-based stack walking instead.
-  Profiler::instance()->recordSampleDelegated(nullptr, _interval, tid,
-                                               BCI_CPU, &event);
-  Shims::instance().setSighandlerTid(-1);
-  errno = saved_errno;
+
```
rkennke, line 129: And here too.

rkennke (approved):
I did a manual review pass, and only found very minor problems. I also ran an AI assisted/Sphinx review pass which found more .. but see the comments there.

ddprof-lib/src/main/cpp/ctimer_linux.cpp
```diff
@@ -206,56 +205,49 @@ Error CTimerJvmti::start(Arguments &args) {
...
     OS::forwardForeignSignal(signo, siginfo, ucontext);
     return;
   }
   Counters::increment(CTIMER_SIGNAL_OWN);
 
-  SIGNAL_HANDLER_GUARD_OR_DROP_WITH_ERRNO(saved_errno);
+  SIGNAL_HANDLER_GUARD_OR_DROP();
   InflightGuard inflight;
   ProfiledThread *current = SIGNAL_HANDLER_CURRENT_THREAD();
   assert(!current->isDeepCrashHandler());
```
rkennke, line 220: 🗿 🤖 🔴

**[Sphinx Review — LOW]** CTimerJvmti::signalHandler is now the only one of the seven refactored handlers that dereferences the SIGNAL_HANDLER_CURRENT_THREAD() result without the `assert(current != nullptr)` this PR standardised on (newly added to ITimer at itimer.cpp:49, already present in CTimer/PerfEvents/ITimerJvmti).

**Suggestion:** Add `assert(current != nullptr);` immediately after the SIGNAL_HANDLER_CURRENT_THREAD() assignment so all seven handlers document the same invariant.

ddprof-lib/src/main/cpp/guards.h
```diff
@@ -308,4 +310,59 @@ class SignalBlocker {
   SignalBlocker& operator=(const SignalBlocker&) = delete;
 };
 
+/**
+ * RAII guard around the span of a signal handler during which the current
+ * thread is the one being sampled. Sets Shims::instance().setSighandlerTid(tid)
+ * on construction and resets it to -1 on destruction, guaranteeing the reset
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — LOW]** SighandlerTidScope's and ErrnoPreserver's doc comments state the cleanup is guaranteed on every return path. The codebase itself documents a counterexample at profiler.cpp:975-987: a chained J9 SIGSEGV handler siglongjmps past our frames and skips RAII destructors — which is precisely why SIGNAL_HANDLER_GUARD_RELEASE() exists. Neither new guard has an equivalent escape hatch, so on that path the sighandler tid is left pinned at a stale value and errno is not restored. Behaviour is unchanged from the pre-refactor manual statements, but the absolute wording will mislead the next reader into assuming the siglongjmp case is now covered.

**Suggestion:** Qualify both comments, e.g. '... on every normal return path out of the guarded scope. Note: a siglongjmp that unwinds past this frame (see Profiler::segvHandler's chained-handler note) bypasses the destructor, matching the pre-existing manual set/reset behaviour.'

ddprof-lib/src/main/cpp/guards.h
```diff
@@ -308,4 +310,59 @@ class SignalBlocker {
...
+ * Must be scoped narrowly around exactly the existing set/reset span (right
+ * before recordSample and right after) rather than wrapped around the whole
+ * handler: widening the scope would change the window during which the
+ * sighandler tid is observably set for other consumers of Shims (e.g.
+ * crash-handler / re-entrant stack-walking code).
+ */
+class SighandlerTidScope {
+public:
+  explicit SighandlerTidScope(int tid) {
+    Shims::instance().setSighandlerTid(tid);
```
rkennke, line 334: 🗿 🤖 🔴

**[Sphinx Review — LOW]** SighandlerTidScope calls Shims::instance() twice per sample, and Shims::instance() returns the singleton *by value* (debugSupport.h:14: `inline static Shims instance() { return _instance; }`), copying the object (including a volatile member load) on each call rather than returning a reference. The new guard cements this pattern into shared infrastructure that now runs on every CPU and wall sample.

**Suggestion:** Change Shims::instance() in debugSupport.h to return `Shims&` (`inline static Shims& instance() { return _instance; }`); all call sites use it as a temporary receiver only, so no caller needs updating.

ddprof-lib/src/main/cpp/guards.h
```diff
@@ -308,4 +310,59 @@ class SignalBlocker {
...
+  }
+  ~SighandlerTidScope() {
```
rkennke, line 336: 🗿 🤖 🔴

**[Sphinx Review — LOW]** SighandlerTidScope restores a hardcoded -1 instead of the prior sighandler tid, so it is not nesting-safe. It is now a shared primitive in guards.h, but its doc comment does not state that non-nesting is a precondition (that property currently comes only from the CriticalSection upstream of every call site).

**Suggestion:** Either document explicitly in the class comment that SighandlerTidScope must not be nested (and why CriticalSection guarantees that at all current sites), or make it save/restore: have the constructor capture the previous tid and the destructor restore it.

ddprof-lib/src/main/cpp/guards.h
```diff
@@ -308,4 +310,59 @@ class SignalBlocker {
...
+  // Non-copyable
+  SighandlerTidScope(const SighandlerTidScope&) = delete;
+  SighandlerTidScope& operator=(const SighandlerTidScope&) = delete;
+};
+
+/**
+ * RAII guard that saves errno on construction and restores it on
+ * destruction, regardless of which return path is taken in between.
+ *
+ * Replaces the previous pattern of `int saved_errno = errno;` at handler
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — LOW]** The ErrnoPreserver doc comment (and SighandlerTidScope's 'exactly the existing set/reset span' at :319, plus PerfFdRearmGuard's new destructor comment) explain the change relative to the pre-PR code rather than the current behavior and contract. Project convention is that comments describe current behavior, with migration rationale going in the PR description.

**Suggestion:** Trim the 'replaces the previous pattern' paragraph and keep only the forward-looking contract: errno is captured at construction and restored at destruction, and the guard must be the first-declared local so it destructs after every other guard that may touch errno.

ddprof-lib/src/main/cpp/guards.h
```diff
@@ -308,4 +310,59 @@ class SignalBlocker {
...
+ * done consistently, still leaves a gap: any destructor that runs after
+ * the manual restore (e.g. a signal-scope guard making a syscall) can
+ * clobber errno again before the handler actually exits.
+ *
+ * To close that gap, declare the ErrnoPreserver as the *first* local in
+ * the guarded function: C++ destroys locals in reverse declaration order,
+ * so it destructs last, after every other guard's destructor has already
+ * run.
+ */
+class ErrnoPreserver {
```
rkennke, line 356: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** ErrnoPreserver and SighandlerTidScope -- the two new RAII guards this PR is built around -- have no unit tests, while the comparatively trivial tickInitWindowIfNeededImpl got a dedicated new gtest file. Their contracts (errno restored on every return path; restored last, after every other guard's destructor; sighandler tid reset on every exit from the scope) are cheap to assert in the existing gtest binary with no JVM and no real signal.

**Suggestion:** Add a small guards_ut.cpp covering: (1) ErrnoPreserver restores errno after a scope that sets it; (2) ErrnoPreserver declared before a second guard whose destructor clobbers errno still wins (the exact ordering invariant the comment documents); (3) SighandlerTidScope sets tid on entry and -1 on exit, including on an early return out of the guarded scope.

rkennke, line 356: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** ErrnoPreserver's correctness depends entirely on it being the first-declared local in the handler (so it destructs last, after SignalHandlerScope / PerfFdRearmGuard destructors that make syscalls). That contract is documented only in comments (guards.h:104-114, 350-353) with zero compile-time or debug-time enforcement. All current call sites are correct, but a future handler — or a refactor that moves the declaration below SIGNAL_HANDLER_GUARD_OR_DROP() — silently reintroduces exactly the errno-clobbering bug this PR fixes, with no test or diagnostic that would catch it.

**Suggestion:** Add a cheap debug-only self-check so the contract is machine-verified rather than comment-only. E.g. give ErrnoPreserver a DEBUG_ONLY member capturing the SignalHandlerScope depth at construction and assert in ~ErrnoPreserver() that no SignalHandlerScope is still live on this thread (i.e. getInSignalDepth() has already returned to its pre-construction value). Alternatively, fold the ErrnoPreserver declaration into SIGNAL_HANDLER_GUARD_OR_DROP() itself — declaring `ErrnoPreserver _errno_preserver;` as the macro's first statement is not sufficient (it must precede the macro), so instead add a SIGNAL_HANDLER_ENTER() macro that declares ErrnoPreserver and is required as the handler's first line, removing the ordering decision from each call site.

ddprof-lib/src/main/cpp/itimer.cpp
```diff
@@ -53,13 +56,14 @@ void ITimer::signalHandler(int signo, siginfo_t *siginfo, void *ucontext) {
   }
```
rkennke, line 56: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** tickInitWindowIfNeeded() is deliberately not wired into ITimer::signalHandler or PerfEvents::signalHandler, and the PR justifies that with "different signal-origin validation characteristics". Signal-origin validation is unrelated to what the init-window guard does: it closes the race between Profiler::registerThread() and thread_native_entry publishing the JVM thread TLS, which is a property of the sampled thread, not of where the signal came from. Both engines call Profiler::recordSample() with a live ucontext on arbitrary threads and are therefore exposed to exactly the crash the guard was added for in ctimer_linux.cpp / wallClock.cpp. Now that the check is a one-line shared helper, the cost of applying it is a single call, and leaving it out with an unrelated rationale makes the omission look reviewed and intentional.

**Suggestion:** Either add `if (tickInitWindowIfNeeded(current)) { return; }` immediately after the CriticalSection entry check in ITimer::signalHandler and PerfEvents::signalHandler, or replace the PR-description rationale with the actual reason these engines are exempt (e.g. they are never enabled on a JVM configuration where the registerThread/thread_native_entry window exists) and record that reason in a code comment at each site.

ddprof-lib/src/main/cpp/itimer.cpp
```diff
@@ -53,13 +56,14 @@ void ITimer::signalHandler(int signo, siginfo_t *siginfo, void *ucontext) {
...
   int tid = current->tid();
-  Shims::instance().setSighandlerTid(tid);
 
-  ExecutionEvent event;
-  event._execution_mode = getThreadExecutionMode();
-  Profiler::instance()->recordSample(ucontext, _interval, tid, BCI_CPU, 0,
-                                     &event);
-  Shims::instance().setSighandlerTid(-1);
+  {
+    SighandlerTidScope sighandlerTid(tid);
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — LOW]** The two new guards introduced by this PR use conflicting local-variable naming in the same functions: `errno_preserver` (snake_case) versus `sighandlerTid` (camelCase). Surrounding locals in these files (saved_errno, call_trace_id, recorded_call_trace_id, observed_state_valid) are snake_case.

**Suggestion:** Rename the SighandlerTidScope locals to snake_case (e.g. sighandler_tid) across all five handlers for consistency with errno_preserver and the surrounding code.

```suggestion
    SighandlerTidScope sighandler_tid(tid);
```

ddprof-lib/src/main/cpp/perfEvents_linux.cpp
```diff
@@ -764,6 +764,9 @@ class PerfFdRearmGuard {
 public:
   PerfFdRearmGuard(int fd, int tid) : _fd(fd), _tid(tid) {}
   ~PerfFdRearmGuard() {
+    // Errno changes made here are caught by the handler-level
```
rkennke, line 767: 🗿 🤖 🔴

**[Sphinx Review — LOW]** PerfFdRearmGuard's destructor performs three errno-setting syscalls and delegates errno restoration to an ErrnoPreserver declared in a different function, enforced only by a comment. The guard cannot verify that precondition, so any future construction site without a preceding ErrnoPreserver silently leaks a clobbered errno into interrupted code.

**Suggestion:** Make the guard self-contained: give PerfFdRearmGuard a private ErrnoPreserver member (member destructors run after the destructor body, so errno is restored after resetBuffer/ioctl) and drop the cross-function comment.

ddprof-lib/src/main/cpp/perfEvents_linux.cpp
```diff
@@ -777,6 +780,7 @@ class PerfFdRearmGuard {
 };
 
 void PerfEvents::signalHandler(int signo, siginfo_t *siginfo, void *ucontext) {
+  ErrnoPreserver errno_preserver;
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — LOW]** ErrnoPreserver is constructed before the external-signal (si_code <= 0) fast-bail, so every non-profiling signal reaching this handler now pays a TLS errno read plus a TLS errno write on a path that previously never touched errno. No profiler-owned work happens before that return, so nothing on that path can clobber errno.

**Suggestion:** Move the ErrnoPreserver declaration below the si_code <= 0 early return, so it still precedes PerfFdRearmGuard (preserving the documented destruction order) but leaves the external-signal fast path untouched.

ddprof-lib/src/main/cpp/threadLocalData.inline.h
```diff
@@ -51,5 +52,27 @@ inline bool ProfiledThread::claimAcquire(int tid) {
     return false;
 }
 
+// Core logic of tickInitWindowIfNeeded(), split out so unit tests can drive
+// has_jvm_thread directly with a plain bool instead of needing a live
+// JVMThread::current(). See tickInitWindow_ut.cpp.
+static inline bool tickInitWindowIfNeededImpl(bool has_jvm_thread, ProfiledThread* current) {
+    if (!has_jvm_thread && current->inInitWindow()) {
```
rkennke, line 62: 🗿 🤖 🔴

**[Sphinx Review — LOW]** tickInitWindowIfNeededImpl() unconditionally dereferences `current` but neither its own doc comment nor tickInitWindowIfNeeded()'s states the non-null precondition, and nothing asserts it. The extraction turned a per-handler inline block (where the SIGNAL_HANDLER_GUARD_OR_DROP non-null invariant was visible three lines above) into a shared header helper reusable by any future signal handler, including ones that source ProfiledThread via ProfiledThread::current() (which can return nullptr).

**Suggestion:** Add an assert(current != nullptr) at the top of tickInitWindowIfNeededImpl() and document the precondition in tickInitWindowIfNeeded()'s comment, so debug/ASan builds catch a future caller that passes a null ProfiledThread instead of faulting inside a signal handler in release.

ddprof-lib/src/main/cpp/threadLocalData.inline.h
```diff
@@ -51,5 +52,27 @@ inline bool ProfiledThread::claimAcquire(int tid) {
...
+        current->tickInitWindow();
+        return true;
+    }
+    return false;
+}
+
+// Shared init-window guard used by the CPU/wall profiling signal handlers.
+// Guards against the race window between Profiler::registerThread() and
+// thread_native_entry setting JVM TLS (PROF-13072): a pure native thread
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — LOW]** Newly added comment carries a Jira ticket ID. Ticket references belong in the PR description, not in code comments.

**Suggestion:** Drop the ticket ID and describe the race directly, e.g. '// thread_native_entry setting JVM TLS: a pure native thread'.

```suggestion
// thread_native_entry setting JVM TLS: a pure native thread
```

ddprof-lib/src/main/cpp/threadLocalData.inline.h
```diff
@@ -51,5 +52,27 @@ inline bool ProfiledThread::claimAcquire(int tid) {
...
+// (where JVMThread::current() is always null) is allowed through once its
+// one-shot init window has ticked down. Returns true if the caller should
+// tick-and-return, in which case the tick has already happened. All call
+// sites are signal handlers holding an ErrnoPreserver, so no manual errno
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** tickInitWindowIfNeeded()'s contract ('All call sites are signal handlers holding an ErrnoPreserver') is unenforceable and already only transitively true: the two wallclock call sites hold no ErrnoPreserver in their own frame -- it is declared in sharedSignalHandler, one frame up. A future call site added without an enclosing ErrnoPreserver loses errno preservation on the tick-and-return path with no compiler diagnostic.

**Suggestion:** Reword the comment to state the precondition as a requirement on callers rather than an assertion of current fact (e.g. 'Callers must be signal handlers with an ErrnoPreserver live in the current handler frame or an enclosing one'), and note that for the wallclock engines the guard is in sharedSignalHandler.

ddprof-lib/src/main/cpp/threadLocalData.inline.h
```diff
@@ -51,5 +52,27 @@ inline bool ProfiledThread::claimAcquire(int tid) {
...
+// handling is needed on this return path.
+static inline bool tickInitWindowIfNeeded(ProfiledThread* current) {
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** tickInitWindowIfNeeded() evaluates JVMThread::current() (a pthread_getspecific) unconditionally on every CPU and wall sample, even though the one-shot init window is closed for the entire life of the thread after the first signal. Because has_jvm_thread is passed by value into tickInitWindowIfNeededImpl(), the cheap inInitWindow() byte load cannot short-circuit the TLS lookup. Gating on inInitWindow() first removes a TLS syscall-free-but-non-trivial lookup from the signal hot path with no semantic change (the composite condition is a pure conjunction).

**Suggestion:** Short-circuit on the cheap per-thread byte before doing the pthread TLS lookup. The condition is a conjunction, so evaluating inInitWindow() first is semantically identical.

ddprof-lib/src/main/cpp/threadLocalData.inline.h
```diff
@@ -51,5 +52,27 @@ inline bool ProfiledThread::claimAcquire(int tid) {
...
+    return tickInitWindowIfNeededImpl(JVMThread::current() != nullptr, current);
```
rkennke, line 89: 🗿 🤖 🔴

**[Sphinx Review — LOW]** tickInitWindowIfNeeded() and tickInitWindowIfNeededImpl() are declared `static inline` at namespace scope in a header. Every other definition in threadLocalData.inline.h uses plain `inline` (external linkage, one merged definition). `static inline` gives each of the ~28 including translation units its own internal-linkage copy of both functions, defeating linker folding and giving the function a different address per TU -- a gratuitous deviation from the file's convention for a helper that has exactly five call sites.

**Suggestion:** Drop the `static` from both tickInitWindowIfNeeded() and tickInitWindowIfNeededImpl(), matching the plain `inline` used by the rest of threadLocalData.inline.h.

```suggestion
inline bool tickInitWindowIfNeeded(ProfiledThread* current) {
```

ddprof-lib/src/test/cpp/tickInitWindow_ut.cpp
```diff
@@ -0,0 +1,75 @@
...
+// which costs far more than a single `!= nullptr` argument expression that
+// compilation and inspection already cover.
+
+#include <gtest/gtest.h>
+#include "threadLocalData.inline.h"
+
+class TickInitWindowImplTest : public ::testing::Test {
+protected:
+    void SetUp() override {
+        ProfiledThread::initCurrentThread();
```
rkennke, line 31: 🗿 🤖 🔴

**[Sphinx Review — LOW]** The new gtest fixture drives real ProfiledThread allocation/TLS publish/teardown but omits the project-wide installGtestCrashHandler<TEST_NAME>() / restoreDefaultSignalHandlers() SetUp/TearDown rig used by ~27 other files in ddprof-lib/src/test/cpp (including threadLocal_ut.cpp, whose fixture covers the same TLS machinery). A crash in the init/release path therefore produces no symbolized diagnostic in CI.

**Suggestion:** Follow the existing convention: call installGtestCrashHandler<TICKINITWINDOW_TEST_NAME>() at the top of SetUp() and restoreDefaultSignalHandlers() at the end of TearDown(), matching threadLocal_ut.cpp.

ddprof-lib/src/main/cpp/ctimer_linux.cpp
yaronguro-datadog: Done — moved into the scope.

ddprof-lib/src/main/cpp/ctimer_linux.cpp
yaronguro-datadog: Done.

ddprof-lib/src/main/cpp/itimer.cpp
yaronguro-datadog: Done.

ddprof-lib/src/main/cpp/itimer.cpp
yaronguro-datadog: Done.

ddprof-lib/src/main/cpp/ctimer_linux.cpp
yaronguro-datadog: Added `assert(current != nullptr);` right after the `SIGNAL_HANDLER_CURRENT_THREAD()` assignment, so all seven handlers now state the same invariant.

ddprof-lib/src/main/cpp/guards.h
yaronguro-datadog: Qualified both comments: the reset/restore is guaranteed on every *normal* return path, with an explicit note that a siglongjmp past the frame (the chained-handler case in `Profiler::segvHandler`) bypasses the destructor, matching the pre-refactor manual statements.

ddprof-lib/src/main/cpp/guards.h
yaronguro-datadog: Documented rather than changed: the class comment now states that the guard is not nesting-safe (it restores a hardcoded -1) and that every current call site sits inside a `CriticalSection`, which rules out a second live guard on the same thread. Making it save/restore would add state for a case that cannot currently occur.

ddprof-lib/src/main/cpp/guards.h
yaronguro-datadog: Trimmed — the comment now states only the current contract (captured at construction, restored at destruction, declare it first so it destructs last).

ddprof-lib/src/main/cpp/threadLocalData.inline.h
yaronguro-datadog: Added `assert(current != nullptr)` to `tickInitWindowIfNeededImpl()` (plus `<cassert>`) and documented the precondition on `tickInitWindowIfNeeded()`.

ddprof-lib/src/main/cpp/threadLocalData.inline.h
yaronguro-datadog: Done — ticket ID dropped, the race is described directly.

ddprof-lib/src/main/cpp/threadLocalData.inline.h
yaronguro-datadog: Reworded as a caller requirement, and it now names `sharedSignalHandler` as where the guard lives for the wallclock engines.

ddprof-lib/src/main/cpp/threadLocalData.inline.h
yaronguro-datadog: Done — the wrapper now bails on `!current->inInitWindow()` before evaluating `JVMThread::current()`, keeping the TLS lookup off the hot path. `tickInitWindowIfNeededImpl()` is unchanged, so the existing unit tests still pin the branch logic.

ddprof-lib/src/main/cpp/threadLocalData.inline.h
yaronguro-datadog: Done — both are plain `inline` now.

ddprof-lib/src/test/cpp/tickInitWindow_ut.cpp
yaronguro-datadog: Done — `installGtestCrashHandler<TICKINITWINDOW_TEST_NAME>()` in SetUp and `restoreDefaultSignalHandlers()` at the end of TearDown, mirroring threadLocal_ut.cpp.

ddprof-lib/src/main/cpp/itimer.cpp
yaronguro-datadog: Done — renamed to `sighandler_tid` at all seven sites.

ddprof-lib/src/main/cpp/perfEvents_linux.cpp
yaronguro-datadog: Done — `ErrnoPreserver` now sits just below the `si_code <= 0` return and still ahead of `PerfFdRearmGuard`, so the external-signal fast path no longer touches errno.

ddprof-lib/src/main/cpp/guards.h
yaronguro-datadog: Added `ddprof-lib/src/test/cpp/guards_ut.cpp` covering `ErrnoPreserver`: restore on normal scope exit, restore on an early return, and the ordering invariant (declared first, it destructs last and wins over a later guard whose destructor clobbers errno). `SighandlerTidScope` is left uncovered — reading the tid back goes through `Shims`, whose symbol is resolved from the JVM, and this gtest binary has none; building a harness for that is more machinery than the two-line guard warrants.

yaronguro-datadog: Not doing this one — both suggested forms add machinery (a DEBUG_ONLY depth capture, or a new mandatory `SIGNAL_HANDLER_ENTER()` macro layered on top of `SIGNAL_HANDLER_GUARD_OR_DROP()`) beyond what a boilerplate-extraction PR should carry. The contract is now stated in the guard's own comment and in the macro comment, and the new guards_ut.cpp ordering test pins the invariant itself. Happy to revisit the macro consolidation separately if you'd like it enforced structurally.

ddprof-lib/src/main/cpp/itimer.cpp
yaronguro-datadog: You're right that the stated rationale was wrong — signal-origin validation has nothing to do with the registerThread/thread_native_entry race. I took the second option: the PR description is corrected (and says so explicitly), and each of `ITimer::signalHandler` and `PerfEvents::signalHandler` now carries a comment recording the real reason — neither engine ever had this check, and adding it is a behaviour change to two engines that a boilerplate-extraction PR shouldn't smuggle in. I deliberately did not claim these engines are exempt because of their JVM configuration; I can't verify that.

ddprof-lib/src/main/cpp/perfEvents_linux.cpp
yaronguro-datadog: Leaving as-is, because the two review passes conflict here: round 1 (3915102779) asked for the nested `ErrnoPreserver` inside `~PerfFdRearmGuard()` to be *removed* as redundant, which is what the tree reflects; this comment asks for it back. Rather than ratchet between passes I've kept the handler-level guard — now declared just below the `si_code <= 0` bail and still ahead of `PerfFdRearmGuard` — and documented the coupling in both the PR description and the guard's comment. The guard has exactly one construction site.

ddprof-lib/src/main/cpp/guards.h
yaronguro-datadog: Deferring this one: `Shims::instance()` lives in debugSupport.h, outside this PR's diff, and switching it from by-value to by-reference changes the semantics for every existing caller for a micro-optimisation. Worth doing, but as its own change rather than folded into a boilerplate extraction.

ddprof-lib/src/main/cpp/wallClock.cpp
yaronguro-datadog: Addressed in an earlier push: `ErrnoPreserver` is now the first local of both `WallClockASGCT::sharedSignalHandler` (wallClock.cpp:207) and `WallClockJvmti::sharedSignalHandler` (wallClock.cpp:413), and is gone from the inner `signalHandler` helpers.

ddprof-lib/src/main/cpp/jvmThread.h
yaronguro-datadog: Addressed. The helper moved to threadLocalData.inline.h and the comment now states the convention as a caller requirement — a signal handler with an `ErrnoPreserver` live in its own frame or an enclosing one (for the wallclock engines that's `sharedSignalHandler`) — with no manual errno handling on this path.

ddprof-lib/src/main/cpp/jvmThread.h
yaronguro-datadog: Addressed by `ddprof-lib/src/test/cpp/tickInitWindow_ut.cpp`, which covers all four `has_jvm_thread` x `inInitWindow()` combinations against `tickInitWindowIfNeededImpl()`, including the not-null/in-window case that an `&&` -> `||` mutation would flip.

ddprof-lib/src/main/cpp/perfEvents_linux.cpp
yaronguro-datadog: Addressed — the nested `ErrnoPreserver` is gone from `~PerfFdRearmGuard()`; only the handler-level one remains. (Round 2's 3933609737 asks for the opposite; I've kept round 1's outcome and noted the conflict there.)

ddprof-lib/src/main/cpp/ctimer_linux.cpp
yaronguro-datadog: Addressed — the split `int tid = 0;` declaration is gone; `tid` is now declared where it's used, inside the `SighandlerTidScope` block.

ddprof-lib/src/main/cpp/itimer.cpp
yaronguro-datadog: Addressed — `#include <errno.h>` is no longer in itimer.cpp; the file's only errno contact is `ErrnoPreserver`, and guards.h brings `<cerrno>`.

ddprof-lib/src/main/cpp/os_linux.cpp
yaronguro-datadog: Addressed — the comment in `OS::forwardForeignSignal` now states the actual rationale (not leaking an errno clobbered by rt_sigprocmask or the chained handler into the interrupted code) and no longer cites CTimer::signalHandler.

ddprof-lib/src/main/cpp/jvmThread.h
yaronguro-datadog: Addressed — the helper moved to threadLocalData.inline.h, which already had `ProfiledThread`; jvmThread.h is back to its forward-declaration-only footprint.

ddprof-lib/src/test/cpp/jvmThread_ut.cpp
```diff
@@ -0,0 +1,132 @@
...
+    return key;
+}
+
+// A marker value guaranteed unique across the whole test binary run, so the
+// initializeKey() scan below can never land on a stale value from an
+// earlier test that happens to still be sitting in sharedTestKey().
+void* nextUniqueMarker() {
+    static std::atomic<uintptr_t> counter{1};
+    return reinterpret_cast<void*>(counter.fetch_add(1));
+}
```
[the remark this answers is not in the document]
yaronguro-datadog: Moot — jvmThread_ut.cpp was removed in 629af9be; the init-window test no longer fabricates a pthread key or synthesises markers, it drives the branch logic directly through `tickInitWindowIfNeededImpl()`.

ddprof-lib/src/main/cpp/guards.h
rkennke: Ok!

ddprof-lib/src/main/cpp/perfEvents_linux.cpp
rkennke: Ok!

rkennke (approved):
Looks good now, thank you! This is a great change!

PR #757, opened 2026-08-27
fix(nativemem): count NM_CALLTRACE residency, not reserved chunk capacity

rkennke:
## What

`NM_CALLTRACE` reported the **virtual capacity** of the call-trace arena rather than the memory actually touched, unlike every other `NM_*` category. This makes it report residency instead.

`LinearAllocator::allocateChunk()` recorded a whole 8 MiB `CALL_TRACE_CHUNK` the moment it was `mmap`'d — before any byte in it was touched. Chunks are then filled incrementally by `alloc()`'s bump pointer, and `reserveChunk()` eagerly reserves the *next* chunk once the current crosses 50% fill, so at any instant at least one fully-counted chunk was mostly or entirely untouched.

The fix records in `alloc()`, for exactly the bytes handed out. Because the bump pointer advances monotonically and every returned pointer is written into immediately by its caller, cumulative bump-allocated bytes track touched — and therefore resident — memory directly, with no correction factor.

## Why it matters

This counter was feeding a memory-overhead reconciliation that had been closed with a `×0.829` residency correction factor borrowed from a different workload. That factor was inflating the call-trace term by ~16 MiB and masking a real, unexplained gap. With the counter measuring directly, the reconciliation is honestly 20–33% under-explained rather than apparently closed — see `doc/performance/dd-trace-doe-reconciliation-2026-08-27.md` on `memory-usage-sweep-handoff`.

It also reorders the optimisation targets: call-trace was believed to be the dominant category and is actually third, behind `native_symbols` and `dictionary`.

## Measurements

dd-trace-doe `enterprise` workload, 12 reps: **calltrace 24.23 → 3.53 MiB**, a 6.9× over-report. Peak touched bytes never exceed 6.42 MiB, so this is not a sampling-phase artifact.

Isolated synthetic sweep, same build, only the accounting differing:

| Accounting | `NM_CALLTRACE` | All categories |
| --- | --- | --- |
| reserved capacity (before) | 24.53 MiB | 42.23 MiB |
| touched bytes (after) | 5.38 MiB | 23.07 MiB |

Every other category was byte-identical across that pair, confirming the change is isolated to this counter.

## Cost

Isolated allocator microbenchmark, 3 runs each, sd < 1 ns:

| Scenario | Before | After | After, peak established |
| --- | --- | --- | --- |
| single thread | 12.62 ns | 25.66 ns | 18.15 ns |
| 8 threads, contended | 64.87 ns | 116.3 ns | ~78.3 ns |

Relative cost roughly doubles, **but `alloc()` is reached only from the `key_value == 0` branch of `CallTraceHashTable::put()`** — only for a call trace not already in the table. A repeat sample of a known stack takes `findCallTrace()` and never allocates. So this is a per-*distinct*-trace cost, not per-sample, which puts it in the low milliseconds per 90-second run.

About 7.4 ns of the single-thread increase is `record()`'s peak high-water CAS. Its comment notes it fires rarely "since `_max` is monotonic" — true for the old once-per-chunk call site, but at a per-alloc call site during pure growth every allocation sets a new high-water. Normal rotation leaves live below an established peak, giving the third column. Coarsening that CAS is a follow-up worth measuring; deliberately not bundled here.

## Also in this change

- **Latent null dereference fixed.** `allocateChunk()` returns `NULL` on `mmap` failure and `detachChunks()` stores that into `_tail`, so the destructor's `freeChunk(_tail)` could be reached with nothing to free. Previously harmless; this change dereferences the chunk to read its extent, so it now guards.
- **Stale header comment corrected.** `nativeMem.h` described `NM_CALLTRACE` as an "arena reserved" figure with `CALLTRACE_STORAGE_BYTES` as its used slice. Both now count touched bytes, and their small difference is the call-trace hash tables — so computing arena waste as the difference would yield ~0.

## Known limitation

Bump-allocated bytes are a **lower bound** on residency. `clear()` resets `offs` and un-records, but does not `munmap` the retained `_tail`, so pages it already touched stay resident while the counter forgets them. The dominant rotation path (`detachChunks` → `freeChunks`) does genuinely `munmap`, bounding this at roughly one chunk. Tracking a per-chunk touched high-water — kept across `clear()`, dropped only on real `munmap` — would close it. Documented in the code and in `memory-sweep-results-linux.md`.

## Testing

`linearAllocator_nativemem_ut.cpp` (new) pins the behaviour: byte-accuracy on a fresh chunk, across a chunk boundary with one pre-reserved, and clean return to zero via both `clear()` and `freeChunks()`. **Verified non-vacuous** — against the old code it fails, reporting a whole 1 MiB chunk where 12,800 B had been handed out.

Release, debug, **ASan and TSan** all green for `linearAllocator_nativemem_ut`, `nativeMem_ut`, `test_callTraceStorage`, `stress_callTraceStorage`. The sanitizer runs matter here because the change adds a read of `current->offs` on the free path; TSan confirms no race and ASan no use-after-free.

### DIFF
--- a/ddprof-lib/src/main/cpp/linearAllocator.cpp
+++ b/ddprof-lib/src/main/cpp/linearAllocator.cpp
@@ -102,6 +102,10 @@ void LinearAllocator::clear() {
     _reserve = current;
     _tail = current;
   }
+  // _tail is kept (not freed) but its contents are discarded here, so its
+  // consumed bytes must be un-recorded explicitly -- freeChunk() won't run
+  // for this one.
+  NativeMem::record(NM_CALLTRACE, -(long long)(_tail->offs - sizeof(Chunk)));
   _tail->offs = sizeof(Chunk);
 
   // DON'T UNPOISON HERE - let alloc() do it on-demand!
@@ -162,15 +166,17 @@ void LinearAllocator::freeChunks(ChunkList& chunks) {
     __tsan_acquire(current);
     #endif
     Chunk* prev = current->prev;
+    // Capture before safeFree unmaps the chunk -- reading current->offs after
+    // that would touch freed memory. This is exactly the byte count alloc()
+    // recorded into NM_CALLTRACE for this chunk, so the decrement is exact.
+    long long used = (long long)(current->offs - sizeof(Chunk));
     #ifdef TSAN_ENABLED
     __tsan_release(current);
     #endif
     OS::safeFree(current, chunks.chunk_size);
     Counters::decrement(LINEAR_ALLOCATOR_BYTES, chunks.chunk_size);
     Counters::decrement(LINEAR_ALLOCATOR_CHUNKS);
-    // The LinearAllocator's only user is call-trace storage, so all of its
-    // chunk memory is attributed to the CALLTRACE category.
-    NativeMem::record(NM_CALLTRACE, -(long long)chunks.chunk_size);
+    NativeMem::record(NM_CALLTRACE, -used);
     current = prev;
   }
 
@@ -194,6 +200,11 @@ void *LinearAllocator::alloc(size_t size) {
          offs = __atomic_load_n(&chunk->offs, __ATOMIC_ACQUIRE)) {
       if (__sync_bool_compare_and_swap(&chunk->offs, offs, offs + size)) {
         void* allocated_ptr = (char *)chunk + offs;
+
+        // The LinearAllocator's only user is call-trace storage, so all of
+        // its bump-allocated bytes are attributed to the CALLTRACE category.
+        // A relaxed atomic add, safe to call from the sampling signal handler.
+        NativeMem::record(NM_CALLTRACE, (long long)size);
 
         // ASAN UNPOISONING: Unpoison ONLY the allocated region on-demand
         // This allows ASan to detect use-after-free of memory that was cleared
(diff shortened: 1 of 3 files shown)


jbachorik:
LGTM!

ddprof-lib/src/main/cpp/linearAllocator.cpp
```diff
@@ -102,6 +102,10 @@ void LinearAllocator::clear() {
     _reserve = current;
     _tail = current;
   }
+  // _tail is kept (not freed) but its contents are discarded here, so its
+  // consumed bytes must be un-recorded explicitly -- freeChunk() won't run
+  // for this one.
+  NativeMem::record(NM_CALLTRACE, -(long long)(_tail->offs - sizeof(Chunk)));
```
jbachorik, line 116: Would it make sense to add `_tail != nullptr` guard around here? Or is it too paranoid?

jbachorik (approved):
LGTM! One small nit only.

ddprof-lib/src/main/cpp/linearAllocator.cpp
rkennke: Good catch, and it turned out to be a real hole — just not at that line.

A guard there alone would not have achieved anything: the very next statement (pre-existing) also dereferences `_tail`, and `clear()` already dereferences `_reserve` unguarded at the top of the function. So the exposure is the whole function, not this line.

It is reachable. `detachChunks()` sets **both** `_tail` and `_reserve` to `NULL` when it cannot allocate a replacement chunk, deliberately leaving the allocator unusable rather than risking a double free. A `clear()` after that would have faulted on `_reserve->prev` before ever reaching the `_tail` read — so the crash was already possible; my change merely added a second dereference to an already-unguarded path.

Fixed in 50c92bb by early-returning on either pointer being `NULL`, which covers both dereferences. That complements the same guard added to `freeChunk()` for the destructor path (`~LinearAllocator` calls `clear()` then `freeChunk(_tail)`).

Not paranoid at all — thanks.

ddprof-lib/src/main/cpp/linearAllocator.cpp
```diff
@@ -195,6 +201,11 @@ void *LinearAllocator::alloc(size_t size) {
       if (__sync_bool_compare_and_swap(&chunk->offs, offs, offs + size)) {
         void* allocated_ptr = (char *)chunk + offs;
 
+        // The LinearAllocator's only user is call-trace storage, so all of
+        // its bump-allocated bytes are attributed to the CALLTRACE category.
+        // A relaxed atomic add, safe to call from the sampling signal handler.
+        NativeMem::record(NM_CALLTRACE, (long long)size);
```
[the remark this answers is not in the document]
rkennke: The inconsistency is real and worth recording, but the resolution goes the other way: converge everything on residency rather than reverting CALLTRACE to backing capacity.

Backing-capacity accounting is badly wrong as an RSS proxy, and the `StringArena` example makes the point rather than undermining it. Measured directly — 20 × `calloc(1, 512 KiB)`, i.e. 10,240 KiB of logical allocation:

```
RSS before        = 1336 KB
RSS after calloc  = 1480 KB   (+144 KB)
RSS after memset  = 11720 KB  (+10384 KB)
```

Large `calloc` gets zeroed pages lazily from `mmap`, so ~1.4 % of the recorded bytes were resident at record time. Whatever basis we pick, "full backing chunk at allocation time" is not measuring memory consumption.

On magnitude, the two cases are not comparable:

- **CALLTRACE** reserved eagerly — `reserveChunk()` grabs the next 8 MiB chunk at 50 % fill — so multiple whole chunks sat counted-but-untouched. Measured over-report on the real workload: **24.23 → 3.53 MiB**, with peak touched bytes never exceeding 6.42 MiB.
- **DICTIONARY** bump-fills sequentially, so every chunk except the active one is fully written and genuinely resident. Its over-report is bounded by the unfilled tail of a single 512 KiB chunk, i.e. **≤ 0.5 MiB**.

So this change moves the largest category from a ~21 MiB error to a bounded one, and leaves a ≤ 0.5 MiB inconsistency behind. Reverting to capacity accounting to restore a consistent basis would reinstate the larger error to remove the smaller.

On the "distinct metric" suggestion — that is roughly what #759 does: it adds `native_mem_chunk_overhead_bytes.<category>`, keeping logical bytes in the existing gauge (comparable to `sizeof()` arithmetic) while reporting the measured allocator cost separately, per category. Measured there: `native_symbols` pays 8.24 %, `dictionary` 0.77 %, a 96-byte tree node 16.67 % — which is also why a single blanket multiplier could not serve them.

Remaining basis gaps are tracked rather than dismissed: the counter still under-reports CALLTRACE residency slightly, because `clear()` un-records a retained chunk while its already-touched pages stay resident until unmapped (bounded by the 3.24 → 6.62 MiB live-to-peak spread, documented in the code and in `memory-sweep-results-linux.md`). Dictionary residency is the natural next one.

PR #758, opened 2026-08-27
Adopt JmpCtxScope for setting stack walk siglongjmp protection

zhengyu123:
**What does this PR do?**:
Replaces the hand-rolled `sigjmp_buf* prev = prof_thread->getJmpCtx(); ... prof_thread->setJmpCtx(prev);` siglongjmp-chaining pattern with the RAII `JmpCtxScope` guard (introduced in #743 for `HotspotSupport::resolve()`) across the remaining stack-walking entry points:

- `HotspotSupport::walkVM()`
- `HotspotSupport::walkJavaStack()`
- `StackWalker::walkFP()`
- `StackWalker::walkDwarf()`

Per site:
- `sigjmp_buf* prev_jmp_buf = getJmpCtx();` → `JmpCtxScope jmp_scope(prof_thread);`
- `setJmpCtx(&crash_protection_ctx);` → `jmp_scope.install(&crash_protection_ctx);`
- In the sigsetjmp fault-recovery branch: `setJmpCtx(prev_jmp_buf);` → `jmp_scope.restore();` (still an explicit disarm before touching anything else that could fault)
- On the normal exit path, the trailing `setJmpCtx(prev_jmp_buf);` is dropped in `walkFP`/`walkDwarf`/`walkJavaStack` — nothing after the walk loop can fault, so the guard's destructor restoring at function exit is sufficient. In `walkVM`, the explicit `jmp_scope.restore()` at the `done:` label is kept, since post-walk work in that frame (leaf-drop, counters) doesn't need the landing pad held live.

`guards.h` is now included by `stackWalker.cpp` (`hotspotSupport.cpp` already had it from #743).

**Motivation**:
Consolidate the siglongjmp landing-pad save/restore/chain logic behind one RAII abstraction instead of four duplicated get/set call sites, so every exit path (normal return, fault recovery, `goto`) is guaranteed to reinstate the previous landing pad without relying on the author remembering to add the restore call.

**Additional Notes**:
No functional/behavioral change intended — this is a pure refactor onto the existing `JmpCtxScope` abstraction from #743. `JmpCtxScope`'s constructor snapshots the previous landing pad before `install()` overwrites it, and `restore()`/the destructor always write that fixed snapshot, so it is safe to call `restore()` explicitly and then let the destructor run again at scope exit.

**How to test the change?**:
`./gradlew buildDebug` compiles clean. No behavior change, so existing stack-walking/crash-protection test coverage applies unchanged.

**For Datadog employees**:
- [X] JIRA: [PROF-15845](https://datadoghq.atlassian.net/browse/PROF-15845)

[PROF-15845]: https://datadoghq.atlassian.net/browse/PROF-15845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
+++ b/ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
@@ -279,7 +279,7 @@ __attribute__((no_sanitize("address"))) int HotspotSupport::walkVM(void* ucontex
     // then we end up with multiple HotspotSupport::walkVM() calls on stack,
     // each one sets up sigjmp_buf, they need to be chained to jump back to
     // correct location.
-    sigjmp_buf* prev_jmp_buf = prof_thread->getJmpCtx();
+    JmpCtxScope jmp_scope(prof_thread);
     // Should be preserved across sigsetjmp/siglongjmp
     volatile int depth = 0;
     int actual_max_depth = truncated ? max_depth + 1 : max_depth;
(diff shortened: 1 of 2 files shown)


jbachorik (approved):
This looks good. Nice cleanup!

PR #774, opened 2026-09-01
Possible out-of-bounds Context read

zhengyu123:
**What does this PR do?**:
Fixes an out-of-bounds read in `Context::get_tag()` (now bounds-checked `getTag`/`setTag`) that could be triggered by configuring more than `DD_TAGS_CAPACITY` (10) custom context attributes via the `attributes=` profiler argument.

- `Context::tags` is a fixed `Tag[DD_TAGS_CAPACITY]` array, but `attributes=` accepted an unbounded number of names and `_num_context_attributes` was set to that unbounded count with no cap.
- `Recording::writeContextSnapshot()` looped up to `numContextAttributes()` calling the unchecked `get_tag(i)`, so configuring 11+ attributes read past the end of `Context` into adjacent memory on every `datadog.HeapLiveObject` event.
- Fix: `Profiler::start()` now caps `args._context_attributes` at `DD_TAGS_CAPACITY` (logging a warning when truncating) before it drives both the JFR metadata schema and `_num_context_attributes`, so the schema and the per-event write count always agree. `Context::get_tag()`/`setTag()` are replaced with encapsulated, bounds-checked `getTag`/`setTag` accessors (private `tags[]`, shared `isValidIndex()` helper) as defense in depth for any future caller.

**Motivation**:
Security review flagged that `attributes=` had no upper bound on the number of configured names, while the native `Context` struct backing per-event tag storage is fixed-size. An out-of-bounds read on every liveness-tracking event can disclose adjacent native heap memory into the JFR recording, or crash the process (e.g. under ASan).

**Additional Notes**:
Iterated based on review feedback:
- `Tighten API`: replaced direct `tags[]` access with private storage + `getTag`/`setTag`.
- `Fix`: gave `setTag` a real runtime bounds check (not just a debug-only `assert`), matching `getTag`'s existing clamp-to-default behavior.
- `Dedup code`: factored the repeated bounds condition out into a shared `isValidIndex()` helper used by both accessors.

**How to test the change?**:
Added `ddprof-test/.../com/datadoghq/profiler/TooManyContextAttributesTest.java`, which configures `attributes=` with 13 names (3 over capacity) together with liveness tracking (`memory=...:L`, the code path that exercises `writeContextSnapshot`/`getTag`), generates enough live allocation volume to produce `datadog.HeapLiveObject` samples, and asserts:
- the recording parses cleanly (would previously corrupt the JFR event stream or crash under ASan on the out-of-bounds read), and
- `jdk.ActiveSetting` reports exactly the first 10 attribute names, confirming the list is capped rather than the requested 13.

**For Datadog employees**:
- [X] JIRA: [PROF-15847](https://datadoghq.atlassian.net/browse/PROF-15847)

[PROF-15847]: https://datadoghq.atlassian.net/browse/PROF-15847?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/context.h
+++ b/ddprof-lib/src/main/cpp/context.h
@@ -32,15 +32,19 @@ public:
   u64 rootSpanId;
 private:
   Tag tags[DD_TAGS_CAPACITY];
+
+  static bool isValidIndex(int i) {
+    return i >= 0 && (u32)i < DD_TAGS_CAPACITY;
+  }
 public:
   u32 getTag(int i) {
-    assert(i >= 0 && (u32)i < DD_TAGS_CAPACITY);
-    return i >= 0 && (u32)i < DD_TAGS_CAPACITY ? tags[i].value : 0;
+    assert(isValidIndex(i));
+    return isValidIndex(i) ? tags[i].value : 0;
   }
   
   void setTag(int i, u32 value) {
-    assert(i >= 0 && (u32)i < DD_TAGS_CAPACITY);
-    if (i >= 0 && (u32)i < DD_TAGS_CAPACITY) {
+    assert(isValidIndex(i));
+    if (isValidIndex(i)) {
       tags[i].value = value;
     }
   }
--- a/ddprof-lib/src/main/cpp/context.h
+++ b/ddprof-lib/src/main/cpp/context.h
@@ -40,7 +40,9 @@ public:
   
   void setTag(int i, u32 value) {
     assert(i >= 0 && (u32)i < DD_TAGS_CAPACITY);
-    tags[i].value = value;
+    if (i >= 0 && (u32)i < DD_TAGS_CAPACITY) {
+      tags[i].value = value;
+    }
   }
 };
 
--- a/ddprof-lib/src/main/cpp/context.h
+++ b/ddprof-lib/src/main/cpp/context.h
@@ -18,6 +18,7 @@
 #define _CONTEXT_H
 
 #include "arch.h"
+#include <cassert>
 
 static const u32 DD_TAGS_CAPACITY = 10;
 
@@ -29,9 +30,18 @@ class alignas(DEFAULT_CACHE_LINE_SIZE) Context {
 public:
   u64 spanId;
   u64 rootSpanId;
+private:
   Tag tags[DD_TAGS_CAPACITY];
-
-  Tag get_tag(int i) { return i >= 0 && (u32)i < DD_TAGS_CAPACITY ? tags[i] : Tag{0}; }
+public:
+  u32 getTag(int i) {
+    assert(i >= 0 && (u32)i < DD_TAGS_CAPACITY);
+    return i >= 0 && (u32)i < DD_TAGS_CAPACITY ? tags[i].value : 0;
+  }
+  
+  void setTag(int i, u32 value) {
+    assert(i >= 0 && (u32)i < DD_TAGS_CAPACITY);
+    tags[i].value = value;
+  }
 };
 
 #endif /* _CONTEXT_H */
--- a/ddprof-lib/src/main/cpp/context.h
+++ b/ddprof-lib/src/main/cpp/context.h
@@ -31,7 +31,7 @@ public:
   u64 rootSpanId;
   Tag tags[DD_TAGS_CAPACITY];
 
-  Tag get_tag(int i) { return tags[i]; }
+  Tag get_tag(int i) { return i >= 0 && (u32)i < DD_TAGS_CAPACITY ? tags[i] : Tag{0}; }
 };
 
 #endif /* _CONTEXT_H */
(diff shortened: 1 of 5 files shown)


jbachorik (approved):
LGTM!

ddprof-test/src/test/java/com/datadoghq/profiler/TooManyContextAttributesTest.java
```diff
@@ -0,0 +1,112 @@
...
+        Thread.sleep(300);
+
+        stopProfiler();
+        assertFalse(liveObjects.isEmpty()); // keep allocations reachable through the GC/dump above
+
+        // If the pre-fix out-of-bounds read had fired, an ASan build would already have
+        // aborted the JVM above. On any build, a mismatched schema/field count would make
+        // this parse fail or throw - reaching here with samples already proves the fix.
+        JfrEvents liveObjectEvents = verifyEvents("datadog.HeapLiveObject", false);
+        assertTrue(liveObjectEvents.hasItems(), "expected datadog.HeapLiveObject samples");
```
rkennke: **MEDIUM (robustness):** The deterministic capping assertions run after a GC/timing-dependent assertion (`liveObjectEvents.hasItems()`). If no `datadog.HeapLiveObject` sample is produced on a given run, the test fails before ever checking that `attributes=` was capped at `MAX_CONTEXT_SLOTS`, so a flake and a real regression of the fix look identical.

Suggestion: reorder the method body so the deterministic `jdk.ActiveSetting` cap assertions run first, then assert `liveObjectEvents.hasItems()` last.

ddprof-lib/src/main/cpp/profiler.cpp
```diff
@@ -1632,6 +1632,11 @@ Error Profiler::start(Arguments &args, bool reset) {
   // Always enable library trap to catch wasmtime loading and patch its broken sigaction
   switchLibraryTrap(true);
 
+  if (args._context_attributes.size() > DD_TAGS_CAPACITY) {
```
rkennke, line 1635: **LOW (completeness):** The `DD_TAGS_CAPACITY` invariant is enforced at a single call site deep inside `Profiler::start()` (after `switchLibraryTrap`/`startRefresher`), not at the point where the list is built in `Arguments::parse` (arguments.cpp:322/325), which still accepts an unbounded number of names. Any future reader of `args._context_attributes` added earlier in `start()` — or any other entry point feeding `JfrMetadata::initialize` — would see the uncapped list again.

Suggestion: apply (or additionally apply) the cap in `Arguments::parse` where `_context_attributes` is populated.

ddprof-lib/src/main/cpp/context.h
```diff
@@ -29,9 +30,24 @@ class alignas(DEFAULT_CACHE_LINE_SIZE) Context {
...
   u64 rootSpanId;
+private:
   Tag tags[DD_TAGS_CAPACITY];
 
-  Tag get_tag(int i) { return tags[i]; }
+  static bool isValidIndex(int i) {
+    return i >= 0 && (u32)i < DD_TAGS_CAPACITY;
+  }
+public:
+  u32 getTag(int i) {
```
rkennke, line 40: **LOW (consistency):** `getTag`/`setTag` take a signed `int` index while both call sites (flightRecorder.cpp:2117 and threadLocalData.cpp:142) iterate with `size_t`, producing an implicit narrowing conversion on every call and requiring the extra `i >= 0` half of `isValidIndex`. The neighboring, equivalent accessor `ProfiledThread::getOtelTagEncoding` (threadLocalData.h:404) already uses an unsigned index with a single-sided check.

Suggestion: take the index as `u32` (or `size_t`) in both `getTag`/`setTag` and `isValidIndex`; `isValidIndex` then reduces to a single `i < DD_TAGS_CAPACITY` comparison.

ddprof-lib/src/main/cpp/context.h
```diff
@@ -29,9 +30,24 @@ class alignas(DEFAULT_CACHE_LINE_SIZE) Context {
 public:
   u64 spanId;
   u64 rootSpanId;
+private:
   Tag tags[DD_TAGS_CAPACITY];
 
-  Tag get_tag(int i) { return tags[i]; }
+  static bool isValidIndex(int i) {
+    return i >= 0 && (u32)i < DD_TAGS_CAPACITY;
```
rkennke, line 37: **LOW (test-adequacy, mutation-testing):** No test in this PR would detect removing the `i >= 0 &&` part of the bounds check in `isValidIndex`.

Suggestion: add a test verifying `isValidIndex` rejects negative indices.

rkennke (commented):
Automated sphinx review — spot-check findings below, none HIGH/CRITICAL.

ddprof-lib/src/main/cpp/profiler.cpp
zhengyu123: Don't see why.

ddprof-lib/src/main/cpp/context.h
zhengyu123: Defer to later cleanup

PR #777, opened 2026-09-02
ci: quarantine flaky tests against a ticket instead of retrying past them

rkennke:
**What does this PR do?**:

Makes flaky tests a tracked, expiring debt instead of something CI quietly retries past.

A retry now buys a *label*, not a pass:

| | |
|---|---|
| **flaky** | failed one attempt, passed another |
| **broken** | failed every attempt |
| **unclassified** | failed, and the suite was not retried — so which of the two was never measured |
| **gating** | not on the quarantine list — red, whichever of the above it is |

A flake fails the build until it is quarantined against a PROF ticket in `ddprof-test/quarantine.txt`. Entries carry a `review_by` date and expire: once it passes, the entry stops excusing anything and CI goes red again. Quarantined tests still **run and report** — only the gating is suspended, so the pass rate still says whether a test is recovering or has quietly become permanently broken. The PR comment prints a paste-ready entry for each flake, with `cells` narrowed to the axes that actually failed.

Only the final attempt's own named failures are weighed against the list, and `flaky` requires observing a pass rather than mere absence from a later attempt. The list cannot excuse:

- a failure in any task other than the test task
- a final attempt that exited non-zero **and shows evidence of being cut short** — Gradle reporting a dead test JVM, or the attempt having reached fewer tests than another attempt managed. A quarantined test that fails makes Gradle exit non-zero by itself, so the exit code alone cannot separate the two
- results that could not be read — a snapshot missing root-owned files, or one merged with an earlier attempt's, is indistinguishable from an attempt whose missing tests all passed
- its own machinery failing: if the classifier cannot run, the job goes red rather than inheriting a pass nothing examined

Conversely, suspect evidence with **no** failure recorded holds no opinion: it is a reason to distrust an excuse, not a test failure, so a suite that passed stays green.

**Motivation**:

CI could not tell a flaky test from a broken one, and its single retry erased the evidence either way:

- the retry existed only for ASan, and a run that recovered on attempt 2 left a green job and a `::warning::` naming nothing
- reports upload only `if: failure()`, so the recovered run — the one worth studying — produced no artifact
- `generate-test-summary.sh` grepped `TEST-*.xml` out of artifacts that only ever held `build/reports/tests` (HTML), never `build/test-results` (XML), so every failed job rendered "No detailed failure information available"
- musl produced no JUnit XML at all, so nothing there could be classified, retried or gated

Answering "is this flaky?" therefore meant hand-comparing PRs by eye. Retrying until green would only have made the tolerance official.

**Additional Notes**:

*musl coverage.* `ProfilerTestRunner` runs musl's tests through the JUnit launcher directly and registered no XML listener. It now attaches a `LegacyXmlReportGeneratingListener` when `-Dtest.reportsDir` is set, and `ProfilerTestPlugin` points that at `build/test-results/<taskName>` — the layout Gradle's own `Test` task uses — so the `TEST-*.xml` scan finds musl's results the same way it finds every other cell's. It fails loudly if reports were requested and cannot be written, since producing none silently is indistinguishable from a cell with nothing to report.

*Patterns match the ids JUnit writes.* Gradle records a `@Test` method as `name="method()"`, so ids arrive as `Class.method()` while entries are written `Class.method`; `covers()` normalises that. `@ParameterizedTest`/`@RetryingTest` invocations appear as `[1]`, `[2]` with no method name at all, so they can only be quarantined class-wide. A trailing `.*` now covers one class's methods, as the format has always documented, rather than being an unbounded prefix — so one validate-clean line can no longer suspend the merge gate repo-wide. `validate` rejects all three unmatchable shapes, and a proposal widened to the class says so with a `WIDENED` line.

*Expiry is enforced where matching happens*, not only in the `validate` subcommand: `validate-quarantine` runs from `ci.yml` alone, while `find_entry` is reached from every workflow reusing `run_tests_with_retry.sh`. An unreadable `review_by` counts as expired rather than as never expiring.

*Proposals narrow on every axis.* `cells_glob` reads the shared axes positionally out of the `<libc>-<jdk>-<config>-<arch>[-slow]` grammar rather than matching a token list, so the JDK and the regular/slow axes cannot be silently dropped — a flake seen on one cell no longer proposes muting it across 13 JDKs and the slow suite.

*Behaviour changes.* Retry now covers all four platform jobs, not just ASan, so a flake-shaped failure costs that cell a second run; slow suites are capped at one attempt to stay clear of the 180-minute timeout; ASan keeps retry-on-abort via `RETRY_ON_NO_TEST_FAILURES`, naming no test when it dies at init.

*Status check names change.* The four matrix jobs gain an explicit `name:`, so their checks go from `test-linux-glibc-amd64 (17, debug)` to `test-linux-glibc-amd64 (17, debug, regular)`. The suffix is required: this workflow is called twice in one run (nightly, release-validated) with overlapping configs, and without it two different jobs collapse onto the same cell. No ruleset on this repo requires status checks by name, so nothing should block on the old strings — but anything outside GitHub that matches on them would need the new form.

**How to test the change?**:

Three hermetic suites, all run by the new `validate-quarantine` job: `test_quarantine.sh` (56 assertions), `test_generate_test_summary.sh` (4), `test_prepare_reports.sh` (2). Between them they cover every validation rejection, glob matching in both directions (a proposal matches the cells it came from and excludes those differing only in JDK or in the slow suffix), each thing the list cannot excuse, the classification edges (absent-from-retry, single-attempt, unnamed testcase), an empty command list, an artifact still shipping XML when a snapshot came out empty, and the rendered output. The summary suite also checks the job-name regex against `test_workflow.yml`'s own `name:` lines, so a rename fails here rather than rendering a summary that reports zero test jobs.

Every guard was mutation-checked: reverting it individually turns its own assertion red.

The retry and gating paths only run once something has failed, so nothing in CI has exercised them end-to-end — they are covered by fixtures instead. Worth knowing when reading the risk of this PR.

**Known gaps, deliberately not in this PR**:

- **No test runs the real producers.** Every XML fixture is hand-written, and nothing runs Gradle or `ProfilerTestRunner` to confirm the shape matches what `flake_report.py` parses. A mismatch there makes the whole list a no-op while every assertion stays green — which is exactly what the `Class.method()` bug above was. The highest-value follow-up.
- **The retry policy is duplicated** across the four platform jobs in two divergent encodings, so which value applies to which (config, suite) pair has no single home.
- **The outcome JSON has a second reader**: the shell re-parses it through an inline `python3 -c` and a two-line stdout protocol, rather than `flake_report.py` communicating its verdict directly.
- **The cell-name grammar has no owner** — six places build or parse it, one of them in a different shape.
- **Auto-filing PROF tickets** (needs dedupe and an Atlassian credential), and the **GitLab dd-trace integration matrix**, which reads validation logs rather than JUnit XML and so needs this policy against a different signal.

**For Datadog employees**:
- [x] JIRA: [PROF-15902](https://datadoghq.atlassian.net/browse/PROF-15902)

[PROF-15902]: https://datadoghq.atlassian.net/browse/PROF-15902?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/scripts/flake_report.py
+++ b/.github/scripts/flake_report.py
@@ -61,7 +61,13 @@ def attempt_results(root_dir):
             # named test either, so it is left to the exit code to report.
             continue
         for case in tree.iter("testcase"):
-            test_id = "{}.{}".format(case.get("classname") or "", case.get("name") or "")
+            name = case.get("name")
+            if not name:
+                # A testcase element with no name cannot be attributed to any
+                # real test; "Class." is not a test id worth counting, tabling
+                # or proposing for quarantine.
+                continue
+            test_id = "{}.{}".format(case.get("classname") or "", name)
             if case.find("skipped") is None:
                 observed.add(test_id)
             problem = case.find("failure")
@@ -116,7 +122,6 @@ def cmd_report(args):
         results.append({
             "test": test_id,
             "failed_attempts": failed_in,
-            "passed_attempts": passed_in,
             "message": next(f[test_id] for _, _, f in attempts if test_id in f),
             "flaky": bool(passed_in),
             "quarantined": hit is not None,
@@ -125,15 +130,33 @@ def cmd_report(args):
 
     gating = [r for r in results if not r["quarantined"]]
 
+    # The caller's exit-code decision must never be made from failures
+    # aggregated across every attempt: those can all be quarantined while the
+    # final attempt itself failed for a reason that named no test at all (a
+    # docker or Gradle configuration failure, an OOM-killed daemon, an ASan
+    # init abort). Report the final attempt's own standing separately so the
+    # caller can require it to have actually produced test results, with every
+    # one of its own named failures quarantined, before trusting the list.
+    final = attempts[-1] if attempts else None
+    final_attempt_ran = final is not None and final[0] == args.final_attempt
+    final_attempt_gating_count = None
+    if final_attempt_ran:
+        _, _, final_failures = final
+        final_attempt_gating_count = sum(
+            1 for test_id in final_failures
+            if quarantine.find_entry(entries, test_id, args.cell) is None
+        )
+
     report = {
         "cell": args.cell,
         "attempts": ran,
-        "status": args.final_status,
         "flaky": [r for r in results if r["flaky"] and not r["quarantined"]],
         "persistent": [r for r in results if not r["flaky"] and not r["quarantined"]],
         "quarantined": [r for r in results if r["quarantined"]],
         "gating_count": len(gating),
         "failure_count": len(results),
+        "final_attempt_ran": final_attempt_ran,
+        "final_attempt_gating_count": final_attempt_gating_count,
     }
 
     os.makedirs(os.path.dirname(args.out) or ".", exist_ok=True)
@@ -169,7 +192,8 @@ def main():
     report = sub.add_parser("report", help="classify failures and decide gating")
     report.add_argument("--cell", required=True)
     report.add_argument("--evidence-dir", required=True)
-    report.add_argument("--final-status", required=True, choices=["pass", "fail"])
+    report.add_argument("--final-attempt", required=True, type=int,
+                         help="the attempt number the caller actually ran last")
     report.add_argument("--out", required=True)
     report.set_defaults(func=cmd_report)
 
--- a/.github/scripts/flake_report.py
+++ b/.github/scripts/flake_report.py
@@ -1,0 +1,145 @@
+#!/usr/bin/env python3
+"""Classify a cell's test failures and decide whether they should gate.
+
+Two questions, answered separately:
+
+  Is it flaky?    Failed on one attempt and passed on another. This is what the
+                  retry exists to establish -- it does NOT excuse the failure.
+  Does it gate?   Only the quarantine list answers that. A failure not on the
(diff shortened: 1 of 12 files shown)


rkennke:
Addressed both P1 findings on the musl "cut short" false positive, and the P2 finding on `cells_glob` mixed-width fallback:

- `flake_report.py`: `cut_short_marker()` now takes `musl=<bool>` (derived from the cell's `<libc>` token) and drops the ambiguous "finished with non-zero exit value" line for musl cells — only the JVM's own crash banner (`hs_err_pid`, "A fatal error has been detected...") gates there now. Two new `test_quarantine.sh` cases cover the false positive and its guard (a real crash must still gate).
- `flake_summary.py`: `render_proposals()` now falls back to the literal observed cells (instead of `[]`, which `applies_to()` reads as "matches everywhere") when `cells_glob()` can't compute a safe glob across mixed-width cells (e.g. `17` + `17-j9`). New test case covers it.

All 59 `test_quarantine.sh` assertions (plus `test_generate_test_summary.sh` and `test_prepare_reports.sh`) pass. Pushed at acd142781.

jbachorik (approved):
Looks feasible, let's give it a try!

PR #779, opened 2026-09-02
[fault-injection] Unprotected access to JVM class loader data mutex and method list

zhengyu123:
**What does this PR do?**:
Hotspot JDK8 specific.

`patchClassLoaderData()` prepends `MethodList` blocks onto a class's `ClassLoaderData` to work around a JDK 8 jmethodID layout bug. This PR hardens that path:
- Guards the `cld->lock()`/`methodList()` mutation and its matching `cld->unlock()` with the profiler's signal/crash-protection machinery (`ProfiledThread`, `JmpCtxScope`, `sigsetjmp`/`SIGNAL_HANDLER_UNWIND_AFTER_LONGJMP`), via the new `LockState` RAII helper, so a crash while the classloader-data mutex is held during a signal-unsafe window unlocks it instead of deadlocking the JVM.
- Tags each patched `jclass` (via JVMTI `Get/SetTag`) with the `method_count` already prepended, so repeated calls for the same class (profiler restart replay, `RedefineClasses`/`RetransformClasses`) only prepend the delta instead of leaking another full set of `MethodList` blocks on every re-run.
- Releases the classloader-data lock before calling the JVMTI tag API, since `cld->lock()` suppresses safepoint checks while `Get/SetTag` are safepoint-polling JVMTI entry points.

**Motivation**:
Fault injection testing found that unprotected access to the JVM class loader data mutex and method list in `patchClassLoaderData()` could crash or deadlock the JVM.

**Additional Notes**:
The `SetTag()` call is not serialized against a concurrent `patchClassLoaderData()` call for the same class, so two racing callers can both observe a stale tag and both patch. This only wastes one extra round of preallocated blocks in that rare case; unlike the original unguarded code, it can no longer grow unboundedly.

**How to test the change?**:
Added unit tests in `ddprof-lib/src/test/cpp/hotspotSupport_ut.cpp` covering the locking/crash-protection and idempotent-tagging behavior of `patchClassLoaderData()`.

**For Datadog employees**:
- [X] JIRA: [PROF-15462](https://datadoghq.atlassian.net/browse/PROF-15462)

[PROF-15462]: https://datadoghq.atlassian.net/browse/PROF-15462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
+++ b/ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
@@ -1335,6 +1335,31 @@ int HotspotSupport::walkJavaStack(StackWalkRequest& request) {
   return java_frames;
 }
 
+class LockState {
+private:
+    VMClassLoaderData* volatile _cld;
+public:
+    LockState() : _cld(nullptr) {}
+    ~LockState() { reset(); }
+    void lock(VMClassLoaderData* cld);
+    void reset();
+};
+
+void LockState::lock(VMClassLoaderData* cld) {
+    cld->lock();
+    _cld = cld;
+}
+
+void LockState::reset() {
+    // Assume: if _cld->lock() did not fail, _cld->unlock() should not
+    // fail as well.
+    // The siglongjmp cannot protect _cld->lock()/unlock() calls
+    if (_cld != nullptr) {
+        _cld->unlock();
+        _cld = nullptr;
+    }
+}
+
 static void patchClassLoaderData(JNIEnv* jni, jclass klass) {
   bool needs_patch = VM::hotspot_version() == 8;
   if (needs_patch) {
@@ -1342,15 +1367,58 @@ static void patchClassLoaderData(JNIEnv* jni, jclass klass) {
     // Preallocate space for jmethodIDs at the beginning of the list (rather than at the end)
     // This is relevant only for JDK 8 - later versions do not have this bug
     if (VMStructs::hasClassLoaderData()) {
+      ProfiledThread* prof_thread = ProfiledThread::initCurrentThreadSignalSafe();
+      assert(prof_thread != nullptr);
+      JmpCtxScope jmp_scope(prof_thread);
+      sigjmp_buf crash_protection_ctx;
+      LockState state;
+      if (sigsetjmp(crash_protection_ctx, 1) != 0) {
+        jmp_scope.restore();
+        state.reset();        
+        return;
+      }
+      jmp_scope.install(&crash_protection_ctx);
       VMKlass *vmklass = VMKlass::fromJavaClass(jni, klass);
       int method_count = vmklass->methodCount();
       if (method_count > 0) {
-        VMClassLoaderData *cld = vmklass->classLoaderData();
-        cld->lock();
-        for (int i = 0; i < method_count; i += MethodList::SIZE) {
-          *cld->methodList() = new MethodList(*cld->methodList());
+        // patchClassLoaderData() re-runs for the same class on every ClassPrepare
+        // replay (profiler restart via loadAllMethodIDsIfNeeded()), RedefineClasses
+        // and RetransformClasses -- none of which change method_count in practice.
+        // Without this tag, each re-run would prepend another full set of
+        // MethodList blocks onto the classloader-wide list that nothing ever frees.
+        // The tag lives on the jclass itself, so it disappears with the class --
+        // no separate bookkeeping to leak or to clean up on unload.
+        jvmtiEnv* jvmti = VM::jvmti();
+        jlong already_patched = 0;
+        if (jvmti == nullptr || jvmti->GetTag(klass, &already_patched) != JVMTI_ERROR_NONE) {
+          already_patched = 0;
         }
-        cld->unlock();
+        if (method_count > already_patched) {
+          VMClassLoaderData *cld = vmklass->classLoaderData();
+          if (cld == nullptr) {
+            return;
+          }
+          state.lock(cld);
+          // Re-check under cld's lock: the GetTag() above is not serialized
+          // against a concurrent patchClassLoaderData() call for the same
+          // class (e.g. RetransformClasses on one thread racing the <clinit>
+          // fallback on the JFR dump thread), so another caller may have
+          // already patched (and updated the tag) while this thread was
+          // waiting for the lock. Only the mutation below is exclusive, so
+          // the decision to mutate must be re-validated inside it.
+          if (jvmti == nullptr || jvmti->GetTag(klass, &already_patched) != JVMTI_ERROR_NONE) {
+            already_patched = 0;
+          }
+          if (method_count > already_patched) {
+            int i;
+            for (i = (int) already_patched; i < method_count; i += MethodList::SIZE) {
+              *cld->methodList() = new MethodList(*cld->methodList());
+            }
+            if (jvmti != nullptr) {
+              jvmti->SetTag(klass, i);
+            }
+          }
+        }
       }
     }
   }
--- a/ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
+++ b/ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
@@ -1399,24 +1399,27 @@ static void patchClassLoaderData(JNIEnv* jni, jclass klass) {
             return;
           }
           state.lock(cld);
-          // Re-check under cld's lock: the GetTag() above is not serialized
-          // against a concurrent patchClassLoaderData() call for the same
-          // class (e.g. RetransformClasses on one thread racing the <clinit>
-          // fallback on the JFR dump thread), so another caller may have
-          // already patched (and updated the tag) while this thread was
-          // waiting for the lock. Only the mutation below is exclusive, so
-          // the decision to mutate must be re-validated inside it.
-          if (jvmti == nullptr || jvmti->GetTag(klass, &already_patched) != JVMTI_ERROR_NONE) {
-            already_patched = 0;
+          int i;
+          for (i = (int) already_patched; i < method_count; i += MethodList::SIZE) {
+            *cld->methodList() = new MethodList(*cld->methodList());
           }
-          if (method_count > already_patched) {
-            int i;
-            for (i = (int) already_patched; i < method_count; i += MethodList::SIZE) {
-              *cld->methodList() = new MethodList(*cld->methodList());
-            }
-            if (jvmti != nullptr) {
-              jvmti->SetTag(klass, i);
-            }
+          // Release cld's lock before touching the JVMTI tag: cld->lock()
+          // resolves to HotSpot's Monitor::lock_without_safepoint_check(), and
+          // GetTag()/SetTag() are full JVMTI entry points that do participate
+          // in safepoint polling -- calling them while holding a
+          // safepoint-check-suppressing lock is a JVM-deadlock hazard on its
+          // own, independent of any fault/siglongjmp path.
+          state.reset();
+          // This SetTag() is not serialized against a concurrent
+          // patchClassLoaderData() call for the same class (e.g.
+          // RetransformClasses on one thread racing the <clinit> fallback on
+          // the JFR dump thread): both can read the same stale tag and both
+          // patch. That only wastes one extra round of preallocated blocks in
+          // the rare concurrent case -- unlike the unguarded original, it
+          // cannot grow unboundedly, since the racing calls all still need a
+          // fresh method_count increase to trigger another round.
+          if (jvmti != nullptr) {
+            jvmti->SetTag(klass, i);
           }
         }
       }


ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -1335,25 +1335,104 @@ int HotspotSupport::walkJavaStack(StackWalkRequest& request) {
...
+  // MethodList blocks onto the classloader-wide list that nothing ever frees.
+  // The tag lives on the jclass itself, so it disappears with the class --
+  // no separate bookkeeping to leak or to clean up on unload.
+  jvmtiEnv* jvmti = VM::jvmti();
+  jlong already_patched = 0;
+  if (jvmti == nullptr || jvmti->GetTag(klass, &already_patched) != JVMTI_ERROR_NONE) {
+    already_patched = 0;
+  }
+  if (method_count <= already_patched) {
+    return -1;
```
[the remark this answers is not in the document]
zhengyu123: Fixed

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -1340,25 +1340,123 @@ int HotspotSupport::walkJavaStack(StackWalkRequest& request) {
...
+  if (prof_thread == nullptr) {
+    return -1;
+  }
+  JmpCtxScope jmp_scope(prof_thread);
+  sigjmp_buf crash_protection_ctx;
+  LockState state;
+  if (sigsetjmp(crash_protection_ctx, 1) != 0) {
+    SIGNAL_HANDLER_UNWIND_AFTER_LONGJMP();
+    jmp_scope.restore();
+    state.reset();
```
jbachorik: **[MEDIUM] Unwind path unlocks after crash protection is torn down**

On the siglongjmp unwind path, `state.reset()` (→ `_cld->unlock()`, HotSpot Monitor code) runs *after* `jmp_scope.restore()` has torn down the crash-protection context. If `unlock()` faults there, no signal protection is active and the fault becomes a raw fatal signal on the profiler-instrumented thread — the same hazard the surrounding code acknowledges for `cld->lock()`.

Suggestion: swap the order — `state.reset()` first (while the `JmpCtxScope` is still installed), then `jmp_scope.restore()`. The `_cld != nullptr` guard already ensures `reset()` only unlocks when the lock was actually acquired.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -1340,25 +1340,123 @@ int HotspotSupport::walkJavaStack(StackWalkRequest& request) {
...
+  // Preallocate space for jmethodIDs at the beginning of the list (rather than at the end)
+  // This is relevant only for JDK 8 - later versions do not have this bug
+  ProfiledThread* prof_thread = ProfiledThread::initCurrentThreadSignalSafe();
+  if (prof_thread == nullptr) {
+    return -1;
+  }
+  JmpCtxScope jmp_scope(prof_thread);
+  sigjmp_buf crash_protection_ctx;
+  LockState state;
+  if (sigsetjmp(crash_protection_ctx, 1) != 0) {
```
jbachorik: **[MEDIUM] Siglongjmp recovery block has zero test coverage**

None of the new tests in `hotspotSupport_ut.cpp` drive `INJECT_CRASH_LIKELY` / `SIGNAL_HANDLER_UNWIND_AFTER_LONGJMP` / `LockState::reset-on-unwind`, although the fault-injection hardening of the lock/unlock pairing is the PR's stated motivation. Every mutation of this block (dropping `state.reset()`, swapping restore/reset order, changing the return value) survives the suite.

Suggestion: add a fault-injection test that faults between `state.lock(cld)` and `state.reset()` and asserts `g_unlock_calls == 1` after the recovery path returns -1, and that a second call still patches from scratch. `gtest_crash_handler.h` is already included in the test file.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -1437,8 +1535,26 @@ bool HotspotSupport::loadMethodIDsIfNeededImpl(jvmtiEnv *jvmti, JNIEnv *jni, jcl
...
+    // consume the slots just prepended, a later successful retry would find
+    // the shared free list empty with no way to top it up again, since
+    // patchClassLoaderData() would see method_count <= already_patched.
+    // Use VM::jvmti(), not the jvmti argument above -- patchClassLoaderData()
+    // reads VM::jvmti() for GetTag, so SetTag must come from the same env for
+    // the two to agree on the tag's meaning.
+    if (loaded && new_tag >= 0) {
+        jvmtiEnv* patch_jvmti = VM::jvmti();
+        if (patch_jvmti != nullptr) {
+            patch_jvmti->SetTag(klass, new_tag);
```
jbachorik: **[LOW] SetTag return value silently discarded**

The tag-based dedup scheme relies on `SetTag()` succeeding: if it fails (e.g. `JVMTI_ERROR_INVALID_CLASS` during unload, or an env/capability mismatch), the tag is silently lost and the next ClassPrepare replay re-prepends the full MethodList block set — exactly the unbounded growth this PR is meant to prevent, with no observable signal.

Suggestion: check the return code and log (TEST_LOG/warn) on failure so the degradation from "tagged dedup" to "re-patch on every call" is diagnosable.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -1340,25 +1340,123 @@ int HotspotSupport::walkJavaStack(StackWalkRequest& request) {
...
+  sigjmp_buf crash_protection_ctx;
+  LockState state;
+  if (sigsetjmp(crash_protection_ctx, 1) != 0) {
+    SIGNAL_HANDLER_UNWIND_AFTER_LONGJMP();
+    jmp_scope.restore();
+    state.reset();
+    return -1;
+  }
+  jmp_scope.install(&crash_protection_ctx);
+  VMKlass *vmklass = VMKlass::fromJavaClass(jni, klass);
```
jbachorik: **[LOW] Missing null check on vmklass**

`vmklass` from `VMKlass::fromJavaClass()` is dereferenced (`vmklass->methodCount()` on the next line) without a null check, while `cld` is null-checked just below. If `fromJavaClass()` can return nullptr for an invalid/unloaded klass, this is a null dereference on a JVMTI-callback-reachable path.

Suggestion: add a symmetric `vmklass == nullptr -> return -1` guard before touching `methodCount()`.

ddprof-lib/src/test/cpp/hotspotSupport_ut.cpp
```diff
@@ -65,11 +71,392 @@ TEST_F(HotspotSupportLoadMethodIDsTest, LoadAllSucceedsAndCallsGetClassMethodsOn
...
+        s.class_loader_data_offset = (int)offsetof(FakePatchKlass, cld);
+        s.vmklass_size = sizeof(FakePatchKlass);
+        s.vmcld_size = sizeof(FakePatchCLD);
+        s.lock_func = &noopLock;
+        s.unlock_func = &noopUnlock;
+        VMStructsTestAccessor::apply(s);
+    }
+
+    void TearDown() override {
+        VMStructsTestAccessor::apply(saved_structs);
```
jbachorik: **[LOW] Global VM state restored only in TearDown**

This fixture mutates process-global VM state (`VM::_hotspot`, `VM::_hotspot_version`, `VM::_jvmti`, VMStructs offsets/lock funcs) in `SetUp()` and restores it only here in `TearDown()`. If a fault injected via the production `INJECT_CRASH_LIKELY()` path longjmps out of the test body and bypasses `TearDown()`, the process-global state stays corrupted for subsequent tests.

Suggestion: a process-wide RAII guard (e.g. installed by `HotspotSupportGlobalSetup`) or a gtest `Environment::TearDown` that unconditionally re-applies the saved state.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -1340,25 +1340,123 @@ int HotspotSupport::walkJavaStack(StackWalkRequest& request) {
...
+  // to leak or to clean up on unload.
+  // RedefineClasses/RetransformClasses callers pass force_patch=true instead
+  // of relying on this tag -- see the function comment above.
+  jlong already_patched = 0;
+  if (!force_patch) {
+    jvmtiEnv* jvmti = VM::jvmti();
+    if (jvmti == nullptr || jvmti->GetTag(klass, &already_patched) != JVMTI_ERROR_NONE) {
+      already_patched = 0;
+    }
+    if (method_count <= already_patched) {
```
jbachorik: **[LOW] Mutation-testing survivors — guard branches not pinned by tests**

- `method_count <= 0` (line 1416): no test exercises a zero-method class; a `<=` → `<` mutant survives. Test: `setMethodCount(0)`, assert no lock calls and no tag persisted.
- `!needs_patch || !VMStructs::hasClassLoaderData()` (line 1394): every test runs with hotspot=true, version=8, has_class_loader_data=true, so neither operand of the OR is ever false. Tests for `setHotspotVersion(9)` and for `has_class_loader_data=false` would pin both halves.
- `prof_thread == nullptr` (line 1401): the early-return branch is never exercised; if `ProfiledThread` init failure can't be forced at unit level, worth documenting why.
- `GetTag` failure (line 1433) is tested with a single error code (`JVMTI_ERROR_INVALID_OBJECT`); parameterizing the mock over `WRONG_PHASE`/`INTERNAL` kills code-specific comparison mutants.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -1437,8 +1535,26 @@ bool HotspotSupport::loadMethodIDsIfNeededImpl(jvmtiEnv *jvmti, JNIEnv *jni, jcl
             jni->DeleteLocalRef(cl);
         }
     }
-    patchClassLoaderData(jni, klass);
-    return JVMSupport::loadMethodIDsImpl(jvmti, jni, klass);
+    jlong new_tag = patchClassLoaderData(jni, klass, force_patch);
```
jbachorik: **[INFO] Tag semantics vs comments**

The persisted tag is the loop's exit value `i` — the next MethodList block boundary past `method_count` (e.g. 24 for method_count=20 with SIZE=8) — not the method count itself. Correct for the `method_count <= already_patched` skip logic (and covered by `SecondCallPatchesOnlyNewMethodRange`), but comments elsewhere describe the tag as "how many of its methods have already been preallocated".

Suggestion: align the comments with the block-boundary semantics (or store the exact `method_count`).

jbachorik (commented):
Sphinx review — 7 approved findings (deduplicated from 15 raw; 0 CRITICAL/HIGH, 2 MEDIUM, 4 LOW, 1 INFO).

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
zhengyu123: If `_cld->lock/unlock()` crashes, it crashes inside JVM, this crash-protection context cannot protect it.

ddprof-lib/src/test/cpp/hotspotSupport_ut.cpp
zhengyu123: We don't inject faults in gtests, as many of them definitely crash.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
zhengyu123: Fixed

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
zhengyu123: Comment updated.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
zhengyu123: Added new tests

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
zhengyu123: The call comes from jvmti->ClassPrepare() callback, therefore, there is not possible for the class to be unloaded inside callback handler. 
But, it does not hurt to add warning message.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
zhengyu123: Added tests

PR #781, opened 2026-09-03
Restore ucontext on faulted paths

zhengyu123:
**What does this PR do?**:
Fixes a corrupted-ucontext bug on the async Java stack-walk path, extends the same fault-tolerance to the JavaThread anchor state it also mutates, and adds regression coverage for both:

- `HotspotSupport::getJavaTraceAsync()` mutates the real signal `ucontext`'s pc/sp/fp in place while it feeds them to HotSpot's own `AsyncGetCallTrace` (via `frame.restore()`/`unwindStub()`/`unwindCompiled()`, and the `PROBE_SP` retry loop), and also patches the live `JavaThread`'s `VMJavaFrameAnchor` (`setLastJavaPC()`) in its `ticks_unknown_not_Java`/`ticks_not_walkable_not_Java` probes. A `SIGSEGV` that strikes mid-mutation is caught by `Profiler::checkFault()` and `siglongjmp`'s straight past any in-function restore — leaving the *actual* ucontext the kernel uses to resume the sampled thread (and/or the live JVM's anchor state) corrupted.
- Adds `HotspotSupport::withUcontextFaultRecovery()`, a reusable sigsetjmp/siglongjmp crash-protection wrapper: it snapshots pc/sp/fp (and, if the wrapped work mutates it, the JavaThread anchor) before running the protected work, chains a `JmpCtxScope` on the current `ProfiledThread`, and restores the snapshot in the recovery branch if `checkFault()` recovers a fault. `walkJavaStack()` now dispatches through it instead of hand-rolling the same protocol inline.
- Extracts the snapshot/restore boilerplate into `StackFrame::RegisterSnapshot` (pc/sp/fp) and `HotspotStackFrame::RegisterSnapshot` (adds `saveJavaAnchor()`/anchor restore on top), shared by both `getJavaTraceAsync()` and the new recovery wrapper. Both snapshot types now carry a destructor that unconditionally calls `restore()` on scope exit, so the restore happens exactly once per call to `withUcontextFaultRecovery()` regardless of which return path is taken — `getJavaTraceAsync()` itself no longer needs (and no longer has) explicit `ctx_snapshot.restore()` calls at each of its own return points, since the destructor on the caller-owned snapshot already covers every one of them.
- **New: `SafeAccess::store()` / `store32()` / `storePtr()`** — a write-side counterpart to the existing `load()`/`load32()`/`loadPtr()`, added specifically so the JavaThread anchor's `lastJavaPC` can be put back on the recovery path without risking a *second* fault. The anchor restore in `RegisterSnapshot::restore()` runs inside (or past) the signal-recovery landing pad itself, where there's no more outer protection layer left to catch a further fault — so that one write goes through the new `SafeAccess::storePtr()` instead of a raw pointer store. This adds `safestore32_impl`/`safestore64_impl` assembly stubs (x86_64 + aarch64, Linux + Apple) mirroring the existing `safefetch32_impl`/`safefetch64_impl`, extends `SafeAccess::handle_safefetch()` to redirect faults from them, and adds `SAFESTORE_FAILED`/`SAFESTORE_WHILE_PROTECTED` counters alongside the existing `SAFEFETCH_*`/`SAFECOPY_*` ones. `VMJavaFrameAnchor::setLastJavaPC()` is now templated on `SafeStore` (defaulting to the safe path); the initial anchor patch in `getJavaTraceAsync()`'s probe uses the plain/unguarded store (already covered by the outer sigsetjmp there), while the recovery-path restore uses the safe store.
- Threads the already-valid partial frame count through a `partial_result` out-parameter so `withUcontextFaultRecovery()`'s shared recovery branch can still report it as truncated-but-valid — preserving `walkJavaStack()`'s pre-existing behavior for a fault recovered after `getJavaTraceAsync()` has already returned frames (e.g. inside `fillFrameTypes()` or the virtual-thread continuation check), not a new fix.
- Rewrites `hotspot_crash_protection_ut.cpp`'s `WalkJavaStackUcontextRestoreTest` suite to call the real `HotspotSupport::withUcontextFaultRecovery()` directly (driven through the actual `Profiler::checkFault()`, not a hand-simulated `siglongjmp`), and adds an `INJECT_FAULT_ADDRESS_UNLIKELY` site on the unguarded anchor-derived `sp` dereference in `getJavaTraceAsync()` so the recovery path is exercised for real rather than only in the unit test's simulated walk.

**Motivation**:
A profiling signal can interrupt a sampled thread at any point, including mid-way through `getJavaTraceAsync()`'s in-place mutation of that thread's own signal `ucontext` or its `JavaThread`'s anchor. If the resulting `SIGSEGV` is recovered by `checkFault()` without putting pc/sp/fp (and the anchor) back the way they were, the signal handler returns and the kernel resumes the sampled thread with a corrupted register set instead of its real one — a crash indistinguishable from stack corruption in the profiled process itself (see branch name). The anchor restore needed its own fault-tolerant store, rather than a raw write, because that particular restore can itself run on the tail end of a fault-recovery path where nothing is left to catch a second fault.

**Additional Notes**:
- `withUcontextFaultRecovery()`'s recovery branch takes a `truncated` flag and a `partial_result` out-parameter, mirroring how `walkJavaStack()` already reported partial/truncated traces before this refactor — no behavior change there; the out-parameter only exists because the shared recovery branch no longer has direct lexical access to `java_frames`.
- The crash-protection gate in `Profiler::checkFault()` only recovers faults whose PC falls inside this library's own address range; a fault inside `libjvm.so` (e.g. `AsyncGetCallTrace` itself dereferencing a poisoned sp/pc/fp) is deliberately *not* recovered here, and the unit tests exercise both sides of that gate via the `UNIT_TEST`-only `Profiler::setAddressRangeForTest()`. On that un-recovered path, `RegisterSnapshot`'s destructor still restores the ucontext once `withUcontextFaultRecovery()` returns normally — that destructor-driven restore is exactly what `FaultOutsideProfilerRangeIsNotRecoveredButUcontextIsStillRestored` pins.
- `SafeAccess::store()`/`store32()` are currently only exercised by the new unit tests; `storePtr()` is the one production caller, from `VMJavaFrameAnchor::setLastJavaPC<true>()`.

**How to test the change?**:
Covered by `hotspot_crash_protection_ut.cpp`'s `WalkJavaStackUcontextRestoreTest` suite:
- `FaultInsideProfilerRangeRecoversAndRestoresUcontext` — a fault inside this library's range is recovered and the ucontext's pc/sp/fp are restored to their pre-walk values.
- `FaultInsideProfilerRangeRecoversAndPreservesPartialResult` — the same recovered fault returns whatever partial frame count `work()` had already committed, not a hardcoded 0.
- `FaultInsideProfilerRangeRecoversAndRestoresJavaThreadAnchor` — a recovered fault also restores the live `JavaThread`'s anchor `lastJavaPC` to its pre-mutation value.
- `FaultOutsideProfilerRangeIsNotRecoveredButUcontextIsStillRestored` — a fault outside the range (standing in for a fault inside `libjvm.so`) is *not* recovered by `checkFault()`, but `withUcontextFaultRecovery()` still restores the ucontext on its normal-completion return, via `RegisterSnapshot`'s destructor.
- `NullUcontextSkipsRestoreWithoutCrashing` — a null ucontext (e.g. malloc/socket hooks sampled outside any signal context) doesn't crash the recovery branch.

And by new cases in `safefetch_ut.cpp`'s `SafeFetchTest` suite covering the new write-side primitives: valid/invalid-pointer round-trips for `store32`/`storePtr`/`store`, plus real SIGSEGV recovery (via `mprotect(PROT_READ)`, to isolate a write fault specifically) for `readOnlyMemoryStore32`/`readOnlyMemoryStorePtr`.

These drive the real production `withUcontextFaultRecovery()` and `SafeAccess` code through the real `Profiler::checkFault()`/signal handlers, so a regression to the actual recovery branch (e.g. dropping a restore call, or a store silently failing) fails these tests too, not just a hand-rolled replica of the same logic.

**For Datadog employees**:
- [X] JIRA: [PROF-15903](https://datadoghq.atlassian.net/browse/PROF-15903)

[PROF-15903]: https://datadoghq.atlassian.net/browse/PROF-15903?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/main/cpp/hotspot/hotspotStackFrame.h
+++ b/ddprof-lib/src/main/cpp/hotspot/hotspotStackFrame.h
@@ -17,7 +17,7 @@ public:
 
     class RegisterSnapshot : public StackFrame::RegisterSnapshot {
         private:
-            // volatile: storeJavaAnchor() mutates these (called from
+            // volatile: saveJavaAnchor() mutates these (called from
             // getJavaTraceAsync(), reached through
             // HotspotSupport::withUcontextFaultRecovery()'s work(ctx_snapshot))
             // between that function's sigsetjmp() and a possible siglongjmp()
@@ -35,16 +35,22 @@ public:
                 _anchor(nullptr), _anchor_pc(nullptr) {
             }
 
-            void storeJavaAnchor(VMJavaFrameAnchor* anchor, const void* pc) {
+            virtual ~RegisterSnapshot() {
+                restore();
+            }
+
+            void saveJavaAnchor(VMJavaFrameAnchor* anchor, const void* pc) {
                 assert(anchor != nullptr);
                 _anchor = anchor;
                 _anchor_pc = pc;
             }
 
-            virtual void restore() {
+            virtual void restore() override {
                 StackFrame::RegisterSnapshot::restore();
                 if (_anchor != nullptr) {
-                    _anchor->setLastJavaPC(_anchor_pc);
+                    // Safe store, cannot fault
+                    bool ret = _anchor->setLastJavaPC<true /*safe store*/>(_anchor_pc);
+                    assert(ret && "Failed to restore lastJavaPC");
                     _anchor = nullptr;
                 }
         }
--- a/ddprof-lib/src/main/cpp/hotspot/hotspotStackFrame.h
+++ b/ddprof-lib/src/main/cpp/hotspot/hotspotStackFrame.h
@@ -15,6 +15,30 @@ public:
     explicit HotspotStackFrame(void* ucontext): StackFrame(ucontext) {
     }
 
+    class RegisterSnapshot : public StackFrame::RegisterSnapshot {
+        private:
+            VMJavaFrameAnchor* _anchor;
+            const void*        _anchor_pc;
+        public:
+            explicit RegisterSnapshot(void* ucontext) : StackFrame::RegisterSnapshot(ucontext),
+                _anchor(nullptr), _anchor_pc(nullptr) {
+            }
+
+            void storeJavaAnchor(VMJavaFrameAnchor* anchor, const void* pc) {
+                assert(anchor != nullptr);
+                _anchor = anchor;
+                _anchor_pc = pc;
+            }
+
+            void restore() {
+                StackFrame::RegisterSnapshot::restore();
+                if (_anchor != nullptr) {
+                    _anchor->setLastJavaPC(_anchor_pc);
+                    _anchor = nullptr;
+                }
+        }
+    };
+
     bool unwindCompiled(VMNMethod* nm) {
         return unwindCompiled(nm, pc(), sp(), fp());
     }
--- a/ddprof-lib/src/main/cpp/hotspot/hotspotStackFrame.h
+++ b/ddprof-lib/src/main/cpp/hotspot/hotspotStackFrame.h
@@ -17,8 +17,19 @@ public:
 
     class RegisterSnapshot : public StackFrame::RegisterSnapshot {
         private:
-            VMJavaFrameAnchor* _anchor;
-            const void*        _anchor_pc;
+            // volatile: storeJavaAnchor() mutates these (called from
+            // getJavaTraceAsync(), reached through
+            // HotspotSupport::withUcontextFaultRecovery()'s work(ctx_snapshot))
+            // between that function's sigsetjmp() and a possible siglongjmp()
+            // out of a recovered fault; restore() then reads them back at the
+            // landing pad to decide whether/how to restore the JavaThread
+            // anchor. Per the setjmp/longjmp rules (C11 7.13.2.1p3, inherited
+            // by C++), a non-volatile automatic local modified in that window
+            // has an indeterminate value after longjmp -- the same hazard
+            // ResolvedNames::_long_method_name (hotspotSupport.cpp) guards
+            // against for an analogous fault-recovery readback.
+            VMJavaFrameAnchor* volatile _anchor;
+            const void* volatile        _anchor_pc;
         public:
             explicit RegisterSnapshot(void* ucontext) : StackFrame::RegisterSnapshot(ucontext),
                 _anchor(nullptr), _anchor_pc(nullptr) {
@@ -30,7 +41,7 @@ public:
                 _anchor_pc = pc;
             }
 
-            void restore() {
+            virtual void restore() {
                 StackFrame::RegisterSnapshot::restore();
                 if (_anchor != nullptr) {
                     _anchor->setLastJavaPC(_anchor_pc);
--- a/ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
+++ b/ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
@@ -1049,23 +1049,24 @@ int HotspotSupport::getJavaTraceAsync(void *ucontext, ASGCT_CallFrame *frames,
   }
 
   HotspotStackFrame frame(ucontext);
-  uintptr_t saved_pc = 0, saved_sp = 0, saved_fp = 0;
+  // Snapshot pc/sp/fp before this function starts feeding them to HotSpot's
+  // own AsyncGetCallTrace below (it mutates them in place via frame.restore()
+  // / frame.unwindStub() / frame.unwindCompiled() to try alternate frames),
+  // so they can be put back once AGCT is done. Shared with walkJavaStack()'s
+  // fault-recovery restore -- see StackFrame::RegisterSnapshot.
+  StackFrame::RegisterSnapshot ctx_snapshot(ucontext);
   if (ucontext != NULL) {
-    saved_pc = frame.pc();
-    saved_sp = frame.sp();
-    saved_fp = frame.fp();
-
-    if (JitCodeCache::isCallStub((const void *)saved_pc)) {
+    if (JitCodeCache::isCallStub((const void *)ctx_snapshot.pc())) {
        // call_stub is unsafe to walk
       frames->bci = BCI_ERROR;
       frames->method_id = (jmethodID) "call_stub";
       return 1;
     }
 
-    if (!VMStructs::isSafeToWalk(saved_pc)) {
+    if (!VMStructs::isSafeToWalk(ctx_snapshot.pc())) {
       frames->bci = BCI_NATIVE_FRAME;
       CodeBlob *codeBlob =
-          VMStructs::libjvm()->findBlobByAddress((const void *)saved_pc);
+          VMStructs::libjvm()->findBlobByAddress((const void *)ctx_snapshot.pc());
       if (codeBlob) {
         frames->method_id = (jmethodID)codeBlob->_name;
       } else {
@@ -1109,7 +1110,7 @@ int HotspotSupport::getJavaTraceAsync(void *ucontext, ASGCT_CallFrame *frames,
   JVMSupport::jvmAsyncGetCallTrace(&trace, max_depth, ucontext);
 
   if (trace.num_frames > 0) {
-    frame.restore(saved_pc, saved_sp, saved_fp);
+    ctx_snapshot.restore();
     return trace.num_frames;
   }
 
@@ -1220,12 +1221,12 @@ int HotspotSupport::getJavaTraceAsync(void *ucontext, ASGCT_CallFrame *frames,
     if (anchor == NULL || anchor->lastJavaSP() == 0) {
       // Do not add 'GC_active' for threads with no Java frames, e.g. Compiler
       // threads
-      frame.restore(saved_pc, saved_sp, saved_fp);
+      ctx_snapshot.restore();
       return 0;
     }
   }
 
-  frame.restore(saved_pc, saved_sp, saved_fp);
+  ctx_snapshot.restore();
 
   if (trace.num_frames > 0) {
     return trace.num_frames + (trace.frames - frames);
@@ -1253,112 +1254,77 @@ int HotspotSupport::walkJavaStack(StackWalkRequest& request) {
   bool* truncated = request.truncated;
   u32 lock_index = request.lock_index;
 
-  volatile int java_frames = 0;
-  // walkVM() installs its own sigsetjmp/siglongjmp crash protection (chained
-  // with any pre-existing jmp ctx, see the comment in walkVM), but the
-  // getJavaTraceAsync() path below runs without one: it dereferences
-  // VMThread/anchor state directly and calls into HotSpot's own
-  // AsyncGetCallTrace. Install a jmp ctx here too, so a SIGSEGV anywhere in
-  // walkJavaStack, except HotSpot's AsyncGetCallTrace call, is caught by
-  // Profiler::checkFault() and siglongjmp'd back here instead of crashing the process.
   ProfiledThread* prof_thread = ProfiledThread::acquireCurrent();
   if (prof_thread == nullptr) {
     Counters::increment(SAMPLES_DROPPED_THREAD_LOCAL);
     return 0;
   }
-  const bool prev_unwinding_java = prof_thread->is_unwinding_Java();
 
-  // getJavaTraceAsync() mutates the real ucontext's pc/sp/fp in place (its
-  // pc()/sp()/fp() are references into uc_mcontext) and restores them itself
-  // on every normal exit path. But a SIGSEGV that strikes mid-mutation (e.g.
-  // the PROBE_SP retry loop, or inside unwindStub/unwindCompiled) is caught
-  // by checkFault() and siglongjmp's straight here, skipping those restores.
-  // Since this ucontext is the same one the kernel will use to resume the
-  // sampled thread when this signal handler returns, snapshot it before the
-  // risky work and restore it here too, or the thread resumes with a
-  // corrupted PC/SP/FP.
-  HotspotStackFrame ctx_frame(ucontext);
-  uintptr_t saved_ctx_pc = 0, saved_ctx_sp = 0, saved_ctx_fp = 0;
-  if (ucontext != NULL) {
-    saved_ctx_pc = ctx_frame.pc();
-    saved_ctx_sp = ctx_frame.sp();
-    saved_ctx_fp = ctx_frame.fp();
-  }
-
-  sigjmp_buf crash_protection_ctx;
-  JmpCtxScope jmp_scope(prof_thread);
-
-  if (sigsetjmp(crash_protection_ctx, 1) != 0) {
-    // checkFault() does a siglongjmp from inside segvHandler, bypassing
-    // segvHandler's SignalHandlerScope destructor. Compensate.
-    SIGNAL_HANDLER_UNWIND_AFTER_LONGJMP();
-    jmp_scope.restore();
-    // A recovered siglongjmp bypasses AsyncSampleMutex destructors, so restore
-    // the per-thread guard to its pre-walk value.
-    prof_thread->set_unwinding_Java(prev_unwinding_java);
-    if (ucontext != NULL) {
-      ctx_frame.restore(saved_ctx_pc, saved_ctx_sp, saved_ctx_fp);
-    }
-    if (truncated) {
-      *truncated = true;
+  // walkVM() installs its own sigsetjmp/siglongjmp crash protection (chained
+  // with any pre-existing jmp ctx, see the comment in walkVM), but the
+  // getJavaTraceAsync() path below runs without one: it dereferences
+  // VMThread/anchor state directly, calls into HotSpot's own
+  // AsyncGetCallTrace, and mutates the real ucontext's pc/sp/fp in place
+  // while doing so. withUcontextFaultRecovery() installs a jmp ctx around
+  // both paths, so a SIGSEGV anywhere in this dispatch (except HotSpot's own
+  // AsyncGetCallTrace call) is caught by Profiler::checkFault() and
+  // recovered instead of crashing the process -- see its own comment for why
+  // it also restores the ucontext.
+  return withUcontextFaultRecovery(ucontext, prof_thread, truncated, [&]() -> int {
+    int java_frames = 0;
+    if (features.mixed) {
+      java_frames = walkVM(ucontext, frames, max_depth, features, eventTypeFromBCI(request.event_type), lock_index, truncated);
+    } else if (isHookPrefixedSample(request.event_type)) {
+      if (cstack >= CSTACK_VM) {
+        java_frames = walkVM(ucontext, frames, max_depth, features, eventTypeFromBCI(request.event_type), lock_index, truncated);
+      } else {
+          AsyncSampleMutex mutex(ProfiledThread::current());
+          if (mutex.acquired()) {
+              java_frames = getJavaTraceAsync(ucontext, frames, max_depth, java_ctx, truncated);
+              if (java_frames > 0 && java_ctx->pc != NULL && VMStructs::hasMethodStructs()) {
+                  VMNMethod* nmethod = CodeHeap::findNMethod(java_ctx->pc);
+                  if (nmethod != NULL) {
+                      fillFrameTypes(frames, java_frames, nmethod);
+                  }
+              }
+          }
+          if (java_frames > 0 && VM::hotspot_version() >= 21 && java_frames < max_depth) {
+              VMThread* carrier = VMThread::current();
+              if (carrier != nullptr && carrier->isCarryingVirtualThread()) {
+                  frames[java_frames].bci = BCI_NATIVE_FRAME;
+                  frames[java_frames].method_id = (jmethodID) "JVM Continuation";
+                  LP64_ONLY(frames[java_frames].padding = 0;)
+                  java_frames++;
+              }
+          }
+      }
+    } else if (request.event_type == BCI_CPU || request.event_type == BCI_WALL) {
+      if (cstack >= CSTACK_VM) {
+          java_frames = walkVM(ucontext, frames, max_depth, features, eventTypeFromBCI(request.event_type), lock_index, truncated);
+      } else {
+          AsyncSampleMutex mutex(ProfiledThread::current());
+          if (mutex.acquired()) {
+              java_frames = getJavaTraceAsync(ucontext, frames, max_depth, java_ctx, truncated);
+              if (java_frames > 0 && java_ctx->pc != NULL && VMStructs::hasMethodStructs()) {
+                  VMNMethod* nmethod = CodeHeap::findNMethod(java_ctx->pc);
+                  if (nmethod != NULL) {
+                      fillFrameTypes(frames, java_frames, nmethod);
+                  }
+              }
+          }
+          if (java_frames > 0 && VM::hotspot_version() >= 21 && java_frames < max_depth) {
+              VMThread* carrier = VMThread::current();
+              if (carrier != nullptr && carrier->isCarryingVirtualThread()) {
+                  frames[java_frames].bci = BCI_NATIVE_FRAME;
+                  frames[java_frames].method_id = (jmethodID) "JVM Continuation";
+                  LP64_ONLY(frames[java_frames].padding = 0;)
+                  java_frames++;
+              }
+          }
+      }
     }
     return java_frames;
-  }
-  jmp_scope.install(&crash_protection_ctx);
-
-  if (features.mixed) {
-    java_frames = walkVM(ucontext, frames, max_depth, features, eventTypeFromBCI(request.event_type), lock_index, truncated);
-  } else if (isHookPrefixedSample(request.event_type)) {
-    if (cstack >= CSTACK_VM) {
-      java_frames = walkVM(ucontext, frames, max_depth, features, eventTypeFromBCI(request.event_type), lock_index, truncated);
-    } else {
-        AsyncSampleMutex mutex(ProfiledThread::current());
-        if (mutex.acquired()) {
-            java_frames = getJavaTraceAsync(ucontext, frames, max_depth, java_ctx, truncated);
-            if (java_frames > 0 && java_ctx->pc != NULL && VMStructs::hasMethodStructs()) {
-                VMNMethod* nmethod = CodeHeap::findNMethod(java_ctx->pc);
-                if (nmethod != NULL) {
-                    fillFrameTypes(frames, java_frames, nmethod);
-                }
-            }
-        }
-        if (java_frames > 0 && VM::hotspot_version() >= 21 && java_frames < max_depth) {
-            VMThread* carrier = VMThread::current();
-            if (carrier != nullptr && carrier->isCarryingVirtualThread()) {
-                frames[java_frames].bci = BCI_NATIVE_FRAME;
-                frames[java_frames].method_id = (jmethodID) "JVM Continuation";
-                LP64_ONLY(frames[java_frames].padding = 0;)
-                java_frames++;
-            }
-        }
-    }
-  } else if (request.event_type == BCI_CPU || request.event_type == BCI_WALL) {
-    if (cstack >= CSTACK_VM) {
-        java_frames = walkVM(ucontext, frames, max_depth, features, eventTypeFromBCI(request.event_type), lock_index, truncated);
-    } else {
-        AsyncSampleMutex mutex(ProfiledThread::current());
-        if (mutex.acquired()) {
-            java_frames = getJavaTraceAsync(ucontext, frames, max_depth, java_ctx, truncated);
-            if (java_frames > 0 && java_ctx->pc != NULL && VMStructs::hasMethodStructs()) {
-                VMNMethod* nmethod = CodeHeap::findNMethod(java_ctx->pc);
-                if (nmethod != NULL) {
-                    fillFrameTypes(frames, java_frames, nmethod);
-                }
-            }
-        }
-        if (java_frames > 0 && VM::hotspot_version() >= 21 && java_frames < max_depth) {
-            VMThread* carrier = VMThread::current();
-            if (carrier != nullptr && carrier->isCarryingVirtualThread()) {
-                frames[java_frames].bci = BCI_NATIVE_FRAME;
-                frames[java_frames].method_id = (jmethodID) "JVM Continuation";
-                LP64_ONLY(frames[java_frames].padding = 0;)
-                java_frames++;
-            }
-        }
-    }
-  }
-
-  return java_frames;
+  });
 }
 
 static void patchClassLoaderData(JNIEnv* jni, jclass klass) {
--- a/ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
+++ b/ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
@@ -1085,14 +1085,7 @@ int HotspotSupport::getJavaTraceAsync(void *ucontext, ASGCT_CallFrame *frames,
   bool in_java = (state == _thread_in_Java || state == _thread_in_Java_trans);
   if (in_java && java_ctx->sp != 0) {
     // skip ahead to the Java frames before calling AGCT.
-    // java_ctx was populated by an earlier, separately-protected walk; fault-
-    // inject its values here to exercise walkJavaStack's ucontext-restore-on-
-    // recovered-fault path -- frame.restore() writes straight into the real
-    // ucontext, and this is one of the few sites that can hand it an
-    // outright invalid pc/sp/fp.
-    frame.restore((uintptr_t)INJECT_FAULT_ADDRESS_UNLIKELY(java_ctx->pc),
-                  INJECT_FAULT_ADDRESS_UNLIKELY(java_ctx->sp),
-                  INJECT_FAULT_ADDRESS_UNLIKELY(java_ctx->fp));
+    frame.restore((uintptr_t)java_ctx->pc, java_ctx->sp, java_ctx->fp);
   } else if (state != _thread_uninitialized) {
     VMJavaFrameAnchor* a = vm_thread->anchor();
     if (a == nullptr || a->lastJavaSP() == 0) {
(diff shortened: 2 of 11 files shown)


ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -1172,8 +1173,12 @@ int HotspotSupport::getJavaTraceAsync(void *ucontext, ASGCT_CallFrame *frames,
...
       // We have the last Java frame anchor, but it is not marked as walkable.
-      // Make it walkable here
-      pc = ((const void**)sp)[-1];
+      // Make it walkable here.
+      // sp comes straight from the anchor with no validation; fault-inject it
+      // so the unguarded dereference below exercises the sigsetjmp/siglongjmp
+      // recovery path installed by the caller (walkJavaStack) instead of only
+      // ever running against a known-good sp.
+      pc = ((const void**)INJECT_FAULT_ADDRESS_UNLIKELY(sp))[-1];
       anchor->setLastJavaPC(pc);
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — HIGH]** The new fault-recovery wrapper restores ucontext pc/sp/fp and ProfiledThread::_unwinding_Java, but not the other piece of state getJavaTraceAsync() temporarily mutates: the JVM's own JavaThread frame anchor. Line 1182 writes an unvalidated pc (read from `((const void**)sp)[-1]`) into last_Java_pc and only clears it again at line 1200 on the normal path. A SIGSEGV between those two lines -- e.g. inside CodeHeap::findNMethod(pc), m->frameSize() or m->frameCompleteOffset() with a garbage pc, all of which are inside this library and therefore recovered by checkFault() -- siglongjmps past line 1200 and leaves the sampled thread's anchor permanently marked walkable with a bogus pc. The PR even adds an INJECT_FAULT_ADDRESS_UNLIKELY site at line 1181 that deliberately steers faults into this block.

**Suggestion:** Extend the recovery protocol to cover the anchor mutation as well: either capture/restore last_Java_pc alongside pc/sp/fp (an anchor snapshot analogous to StackFrame::RegisterSnapshot, restored in withUcontextFaultRecovery()'s recovery branch), or validate `pc` before writing it into the anchor and clear it before any call that can fault, so a recovered fault can never leave the JVM anchor mutated.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.h
```diff
@@ -38,7 +39,59 @@ class HotspotSupport {
...
+            SIGNAL_HANDLER_UNWIND_AFTER_LONGJMP();
+            jmp_scope.restore();
+            // A recovered siglongjmp bypasses AsyncSampleMutex destructors, so
+            // restore the per-thread guard to its pre-walk value.
+            prof_thread->set_unwinding_Java(prev_unwinding_java);
+            ctx_snapshot.restore();
+            if (truncated) {
+                *truncated = true;
+            }
+            return partial_result ? *partial_result : 0;
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** The PR description claims this PR fixes a bug where 'a fault recovered after getJavaTraceAsync() has already returned a valid frame count ... now reports that partial trace as truncated instead of discarding it as zero frames'. The merge base (95ac99370) already did exactly that: its recovery branch was `... if (truncated) { *truncated = true; } return java_frames;` over the same `volatile int java_frames`. The `partial_result` out-parameter is behaviour-preserving plumbing for the extraction, not a fix, and the description's own Additional Notes section contradicts the bullet.

**Suggestion:** Drop the 'fixes discarding partial traces as zero frames' claim from the PR description (and soften the corresponding paragraph at hotspotSupport.h:55-61) to state that partial_result preserves the pre-existing partial/truncated reporting through the extraction.

ddprof-lib/src/main/cpp/stackFrame.h
```diff
@@ -33,6 +33,42 @@ class StackFrame {
...
+                _fp = frame.fp();
+            }
+        }
+
+        void restore() const {
+            StackFrame(_ucontext).restore(_pc, _sp, _fp);
+        }
+
+        uintptr_t pc() const { return _pc; }
+        uintptr_t sp() const { return _sp; }
```
rkennke, line 67: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** RegisterSnapshot::sp() and RegisterSnapshot::fp() have no callers anywhere in the repository. Only pc() (three call sites in getJavaTraceAsync) and restore() are used. They are dead code introduced by this diff.

**Suggestion:** Drop the unused sp() and fp() accessors; keep only pc() (the one getJavaTraceAsync's pre-checks need) and restore().

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.h
rkennke: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** The new `partial_result` out-parameter — the only behavioural addition in withUcontextFaultRecovery() beyond the extracted boilerplate — is never exercised by the new tests. All three WalkJavaStackUcontextRestoreTest cases omit the argument, so it is always nullptr and every assertion is EXPECT_EQ(0, result); stubbing this line to `return 0;` would leave the suite passing.

**Suggestion:** Add a test that passes a non-null volatile int* whose value the lambda sets (e.g. to 3) before faulting, and assert the recovered call returns 3 with truncated == true — that is the partial-trace semantics the PR claims to fix.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -1248,91 +1253,77 @@ int HotspotSupport::walkJavaStack(StackWalkRequest& request) {
...
+              VMThread* carrier = VMThread::current();
+              if (carrier != nullptr && carrier->isCarryingVirtualThread()) {
+                  frames[java_frames].bci = BCI_NATIVE_FRAME;
+                  frames[java_frames].method_id = (jmethodID) "JVM Continuation";
+                  LP64_ONLY(frames[java_frames].padding = 0;)
+                  java_frames++;
+              }
+          }
+      }
+    } else if (request.event_type == BCI_CPU || request.event_type == BCI_WALL) {
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** The `isHookPrefixedSample` arm (1276-1299) and the `BCI_CPU || BCI_WALL` arm (1300-1323) are verbatim duplicates of the same 24-line body. The refactor touched and re-indented every line of both copies but left them duplicated, so any future fix to the AsyncSampleMutex / fillFrameTypes / virtual-thread-continuation logic must be applied twice or silently diverge.

**Suggestion:** Collapse the two arms into one: compute `bool use_async = !features.mixed && (isHookPrefixedSample(request.event_type) || request.event_type == BCI_CPU || request.event_type == BCI_WALL);` and extract the shared body into a static helper (e.g. `static int asyncJavaTraceWithPostProcessing(...)`) called from a single branch.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.h
```diff
@@ -7,16 +7,17 @@
 #ifndef _HOTSPOT_HOTSPOTSUPPORT_H
 #define _HOTSPOT_HOTSPOTSUPPORT_H
 
+#include "guards.h"
 #include "hotspot/hotspotStackFrame.h"
 #include "hotspot/jitCodeCache.h"
 #include "frame.h"
 #include "stackFrame.h"
 #include "stackWalker.h"
+#include "threadLocalData.inline.h"
```
rkennke, line 16: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** Replacing the `class ProfiledThread;` forward declaration with a full `#include "threadLocalData.inline.h"` (which itself pulls in guards.h, os.h, threadLocalData.h and threadLocalDataPool.h) widens the dependency surface of a header transitively included by profiler.cpp, jvmSupport.cpp/.inline.h, jvmThread.cpp and vmStructs.inline.h. The only consumer of the extra declarations is the new inline template, which does not belong in this header in the first place.

**Suggestion:** Move `withUcontextFaultRecovery()` out of `hotspotSupport.h` (see the guards.h suggestion) and restore the `class ProfiledThread;` forward declaration here.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.h
```diff
@@ -38,7 +39,59 @@ class HotspotSupport {
...
+    // than being discarded as zero frames.
+    //
+    // Extracted into one place, rather than hand-rolled separately in
+    // walkJavaStack(), so production and its regression test invoke the
+    // identical recovery branch -- see hotspot_crash_protection_ut.cpp's
+    // WalkJavaStackUcontextRestoreTest. A template rather than
+    // std::function<int()> so the hot sample path pays no allocation for
+    // captures.
+    template <typename Fn>
+    static int withUcontextFaultRecovery(void* ucontext, ProfiledThread* prof_thread, bool* truncated, Fn&& work, volatile int* partial_result = nullptr) {
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** `withUcontextFaultRecovery()` contains no HotSpot-specific logic (only ProfiledThread, JmpCtxScope and StackFrame), yet it is declared as a public static member of the HotSpot-only `HotspotSupport` class. The three other sites that hand-roll the identical sigsetjmp/JmpCtxScope/SIGNAL_HANDLER_UNWIND_AFTER_LONGJMP protocol -- stackWalker.cpp:58 (walkFP), stackWalker.cpp:144 (walkDwarf) and flightRecorder.cpp:720 -- live in JVM-agnostic translation units and cannot reuse it without pulling `hotspot/hotspotSupport.h` into shared code, which the project's layering rule prohibits. It also forces the widened includes (`guards.h`, `threadLocalData.inline.h`) into a header transitively reachable from `jvmSupport.inline.h` and `vmStructs.inline.h`.

**Suggestion:** Move the template next to `JmpCtxScope` in `guards.h` (or a small `crashRecovery.h`) as a free function, e.g. `template <typename Fn> int withUcontextFaultRecovery(void* ucontext, ProfiledThread* pt, bool* truncated, Fn&& work, volatile int* partial_result = nullptr)`. That keeps `hotspotSupport.h`'s include surface unchanged, lets walkFP/walkDwarf/flightRecorder migrate onto it later, and removes the need to make a test-only entry point `public` on `HotspotSupport` (which already has the `HotspotSupportTestAccessor` friend for exactly that purpose).

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -1172,8 +1173,12 @@ int HotspotSupport::getJavaTraceAsync(void *ucontext, ASGCT_CallFrame *frames,
...
     if (sp != 0 && pc == NULL) {
       // We have the last Java frame anchor, but it is not marked as walkable.
-      // Make it walkable here
-      pc = ((const void**)sp)[-1];
+      // Make it walkable here.
+      // sp comes straight from the anchor with no validation; fault-inject it
+      // so the unguarded dereference below exercises the sigsetjmp/siglongjmp
+      // recovery path installed by the caller (walkJavaStack) instead of only
+      // ever running against a known-good sp.
+      pc = ((const void**)INJECT_FAULT_ADDRESS_UNLIKELY(sp))[-1];
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** The new fault-injection site poisons the base pointer `sp` but the dereference applies a negative index (`[-1]`). faultinj::poisonAddress() can return exactly the first byte of the PROT_NONE guard region, so `[-1]` reads 8 bytes *before* the guard region -- an address that is often mapped and readable. The injection then produces no SIGSEGV and no recovery; instead `pc` is arbitrary garbage that is immediately written into the sampled thread's live JVM JavaFrameAnchor via anchor->setLastJavaPC(pc) and fed to AsyncGetCallTrace, where a resulting fault has its PC inside libjvm.so and is deliberately NOT recovered by Profiler::checkFault.

**Suggestion:** Wrap the fully-offset address so the poison itself is the address that gets dereferenced, keeping the non-injected expansion byte-identical: `pc = *(const void**)INJECT_FAULT_ADDRESS_UNLIKELY((const void**)sp - 1);`

```suggestion
      pc = *(const void**)INJECT_FAULT_ADDRESS_UNLIKELY((const void**)sp - 1);
```

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -1248,91 +1253,77 @@ int HotspotSupport::walkJavaStack(StackWalkRequest& request) {
...
+  // getJavaTraceAsync() path below runs without one: it dereferences
+  // VMThread/anchor state directly, calls into HotSpot's own
+  // AsyncGetCallTrace, and mutates the real ucontext's pc/sp/fp in place
+  // while doing so. withUcontextFaultRecovery() installs a jmp ctx around
+  // both paths, so a SIGSEGV anywhere in this dispatch (except HotSpot's own
+  // AsyncGetCallTrace call) is caught by Profiler::checkFault() and
+  // recovered instead of crashing the process -- see its own comment for why
+  // it also restores the ucontext.
+  volatile int java_frames = 0;
+  return withUcontextFaultRecovery(ucontext, prof_thread, truncated, [&]() -> int {
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** withUcontextFaultRecovery() wraps the whole walkJavaStack dispatch, including the three walkVM branches. walkVM neither mutates the ucontext (it unwinds through local pc/sp/fp references, lines 602/643/694) nor lets faults escape (it installs its own chained JmpCtxScope at line 282, so checkFault siglongjmps into walkVM's recovery, not the outer wrapper's). So for mixed / cstack>=CSTACK_VM samples the RegisterSnapshot is pure overhead: three out-of-line StackFrame::pc()/sp()/fp() calls into uc_mcontext per sample whose result can never be restored.

**Suggestion:** Narrow the wrapper to the branches that actually mutate the ucontext: keep walkJavaStack's plain dispatch and apply withUcontextFaultRecovery() only around the two `AsyncSampleMutex + getJavaTraceAsync + fillFrameTypes + isCarryingVirtualThread` blocks (which is also where the partial_result semantics are meaningful), or make the snapshot lazy/conditional on the branch taken. Alternatively note in the wrapper's doc comment that the snapshot is only live for the AGCT path.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.h
rkennke: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** The wrapper carries the protected work's result through two redundant channels for the same variable: `work()`'s return value on the normal path and `*partial_result` on the recovery path. The only caller wires both to the same `volatile int java_frames`. Worse, `partial_result` defaults to nullptr, so the correct usage is opt-in: any future caller that omits it silently reverts to returning 0 on a recovered fault - the very 'recovered fault discards an already-valid partial trace' bug this PR claims to fix. Nothing in the signature or the type system forces the accumulator and the lambda's return value to be the same object.

**Suggestion:** Collapse to a single channel: make `work` a `void`-returning callable that writes only into a required (non-defaulted) `volatile int& partial_result`, and have the wrapper return `partial_result` on both paths. That makes the accumulator and the result the same object by construction and removes the silently-wrong default.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -1147,7 +1148,7 @@ int HotspotSupport::getJavaTraceAsync(void *ucontext, ASGCT_CallFrame *frames,
               trace.frames--;
             }
             for (int i = 0; trace.num_frames < 0 && i < PROBE_SP_LIMIT; i++) {
-              frame.sp() += sizeof(void*);
+              frame.sp() = frame.sp() + sizeof(void*);
```
rkennke, line 1155: 🗿 🤖 🔴

**[Sphinx Review — LOW]** Rewriting `frame.sp() += sizeof(void*)` as `frame.sp() = frame.sp() + sizeof(void*)` is a semantic no-op (sp() returns uintptr_t&) and adds unexplained noise to a signal-path hunk in an otherwise narrowly scoped fix.

**Suggestion:** Revert to the original compound-assignment form, or state in the PR why the rewrite is needed.

```suggestion
              frame.sp() += sizeof(void*);
```

ddprof-lib/src/test/cpp/hotspot_crash_protection_ut.cpp
```diff
@@ -576,4 +580,226 @@ TEST_F(SafeFetch64TocTouGuardTest, ZeroReturnMeansGiveUp) {
...
+// straight back out, proving checkFault() truly fell through rather than
+// recovering.
+TEST_F(WalkJavaStackUcontextRestoreTest, FaultOutsideProfilerRangeIsNotRecoveredAndLeavesUcontextCorrupted) {
+    StackFrame frame(&_ctx);
+    uintptr_t saved_pc = frame.pc();
+    uintptr_t saved_sp = frame.sp();
+    bool truncated = false;
+    uintptr_t mutated_pc = 0, mutated_sp = 0, mutated_fp = 0;
+
+    int result = HotspotSupport::withUcontextFaultRecovery(&_ctx, _pt, &truncated, [&]() -> int {
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — LOW]** This test exercises Profiler::checkFault()'s address-range gate, not withUcontextFaultRecovery(): every assertion still passes if the wrapper is reduced to `return work();`, because on the non-recovering path none of the wrapper's sigsetjmp / JmpCtxScope / RegisterSnapshot / set_unwinding_Java logic runs. The suite comment above attributes the coverage to the wrapper.

**Suggestion:** Adjust the comment to say this is a negative control for checkFault()'s range gate (making explicit that FaultInsideProfilerRangeRecoversAndRestoresUcontext is the test that actually pins withUcontextFaultRecovery()'s recovery branch), so a future reader does not treat it as regression coverage for the wrapper.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -1248,91 +1253,77 @@ int HotspotSupport::walkJavaStack(StackWalkRequest& request) {
...
-            VMThread* carrier = VMThread::current();
-            if (carrier != nullptr && carrier->isCarryingVirtualThread()) {
-                frames[java_frames].bci = BCI_NATIVE_FRAME;
-                frames[java_frames].method_id = (jmethodID) "JVM Continuation";
-                LP64_ONLY(frames[java_frames].padding = 0;)
-                java_frames++;
-            }
-        }
+      } else {
+          AsyncSampleMutex mutex(ProfiledThread::current());
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — LOW]** The lambda re-resolves the thread-local ProfiledThread via ProfiledThread::current() (a pthread_getspecific) for AsyncSampleMutex, even though walkJavaStack already holds the identical, null-checked pointer in `prof_thread`. This also splits the AsyncSampleMutex's _unwinding_Java owner from the pointer whose _unwinding_Java the recovery branch restores (prev_unwinding_java is read from prof_thread), so the two should be the same expression.

**Suggestion:** Use the already-acquired local at both sites: `AsyncSampleMutex mutex(prof_thread);` (captured by the lambda), removing two pthread_getspecific calls per sample and making the mutex and the wrapper's prev_unwinding_java restore operate on the same ProfiledThread by construction.

```suggestion
          AsyncSampleMutex mutex(prof_thread);
```

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.h
```diff
@@ -38,7 +39,59 @@ class HotspotSupport {
...
+        const bool prev_unwinding_java = prof_thread->is_unwinding_Java();
```
rkennke, line 114: 🗿 🤖 🔴

**[Sphinx Review — LOW]** withUcontextFaultRecovery() is newly-exported public API that dereferences `prof_thread` on its first statement with no null check and no documented non-null precondition, even though the JmpCtxScope it wraps documents '`pt` must be non-null' and the usual sources (ProfiledThread::acquireCurrent()/current()) can return nullptr.

**Suggestion:** State the precondition in the doc comment above (matching JmpCtxScope's '`pt` must be non-null') and add `assert(prof_thread != nullptr);` as the first statement so debug/test builds catch a future caller that forgets the check.

ddprof-lib/src/main/cpp/stackFrame.h
```diff
@@ -33,6 +33,42 @@ class StackFrame {
...
+        explicit RegisterSnapshot(void* ucontext) : _ucontext(ucontext) {
+            if (_ucontext != nullptr) {
+                StackFrame frame(_ucontext);
+                _pc = frame.pc();
+                _sp = frame.sp();
+                _fp = frame.fp();
+            }
+        }
+
+        void restore() const {
```
rkennke: 🗿 🤖 🔴

**[Sphinx Review — LOW]** `RegisterSnapshot::restore()` is marked `const` even though its sole effect is writing pc/sp/fp back into the ucontext -- the most consequential mutation in this change. Separately, `_pc/_sp/_fp` are non-const and assigned in the constructor body, whereas guards.h:208-213 explicitly documents that guard state living in a sigsetjmp frame is made `const` so that reading it from the landing pad is well-defined; `RegisterSnapshot` sits in exactly such a frame (hotspotSupport.h:72) but does not follow that convention.

**Suggestion:** Drop the `const` from `restore()` so the signature reflects the mutation, and initialise `_pc/_sp/_fp` from a delegating constructor / helper so they can be declared `const uintptr_t`, matching the JmpCtxScope convention documented in guards.h.

```suggestion
        void restore() {
```

ddprof-lib/src/main/cpp/stackFrame.h
```diff
@@ -33,6 +33,42 @@ class StackFrame {
...
+    // unwindStub()/unwindCompiled() writing pc()/sp()/fp() by reference).
+    // Capturing is a no-op when ucontext is null (pc()/sp()/fp() themselves
+    // are not null-safe); restore() delegates to StackFrame::restore() above,
+    // which already is.
+    //
+    // Must be restored via an explicit restore() call, never a destructor:
+    // Profiler::checkFault()'s siglongjmp bypasses destructors, so RAII alone
+    // can't reach code here -- the same reason JmpCtxScope::restore() must be
+    // called explicitly (see guards.h).
+    class RegisterSnapshot {
```
rkennke, line 48: 🗿 🤖 🔴

**[Sphinx Review — LOW]** RegisterSnapshot deliberately forgoes RAII (documented at stackFrame.h:44-47: siglongjmp bypasses destructors), which pushes the restore obligation onto every one of getJavaTraceAsync()'s seven exit points. There is no enforcement for that obligation, and the new walkJavaStack wrapper only compensates on the *fault* path (hotspotSupport.h:85), never on a normal early return. The PR introduces the abstraction that would naturally have closed this class of bug but leaves it as unchecked convention, so any future early return added between the ucontext mutation at hotspotSupport.cpp:1089 and the final restore at :1228 corrupts the sampled thread's sigreturn register set with no test or static check to catch it.

**Suggestion:** Give RegisterSnapshot both a restore() (for the siglongjmp landing pad) and a destructor that calls it, matching JmpCtxScope's design in guards.h (destructor + idempotent explicit restore()). The explicit call stays required only at the landing pad; every normal-return path is then covered automatically. Alternatively, restructure getJavaTraceAsync so the ucontext-mutating region is a single inner helper whose sole return point does the restore.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
rkennke, line 1155: 🗿 🤖 🔴

**[Sphinx Review — LOW]** No test in this PR would detect: mutating the arithmetic operator in `frame.sp() = frame.sp() + sizeof(void*);` from + to - (or removing the operand entirely) in the PROBE_SP retry loop of getJavaTraceAsync().

**Suggestion:** Add a test that directly exercises getJavaTraceAsync()'s PROBE_SP retry path with controlled JVM state where AsyncGetCallTrace fails, verifying the stack pointer is correctly incremented.

rkennke (changes requested):
🗿 🤖 🔴

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
zhengyu123: Reverted

ddprof-lib/src/main/cpp/stackFrame.h
zhengyu123: Pre-exist, won't address here.

ddprof-lib/src/main/cpp/stackFrame.h
zhengyu123: Defer to later time.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
zhengyu123: Reverted

zhengyu123:
**Key additions to what was there before:**

- New `SafeAccess::store()/store32()/storePtr()` section explaining why they exist: the JavaThread anchor's lastJavaPC restore in `RegisterSnapshot::restore()` runs on the tail end of a fault-recovery path where a second fault has nowhere left to be caught, so that one write now goes through a fault-tolerant store (new `safestore32_impl/safestore64_impl` asm stubs, `handle_safefetch` extended to catch their faults, new SAFESTORE_FAILED/SAFESTORE_WHILE_PROTECTED counters) instead of a raw pointer write.
- `setLastJavaPC()` now templated on `SafeStore`, and RegisterSnapshot's destructor now unconditionally restores on scope exit — so the explicit ctx_snapshot.restore() calls scattered through getJavaTraceAsync() were removed as redundant.
- `setLastJavaPC()` uses `SafeStore` on recovery paths to avoid faulting.
- Updated How to test with the actual current test names (...RestoresJavaThreadAnchor, ...PreservesPartialResult, the renamed ...ButUcontextIsStillRestored) and the new safefetch_ut.cpp store-path coverage.

ddprof-lib/src/main/cpp/stackFrame.h
```diff
@@ -33,6 +33,45 @@ class StackFrame {
...
 
+    // Captures pc/sp/fp for a later restore() -- the null-safe save/restore
+    // boilerplate needed around code that mutates the real ucontext in place
+    // (e.g. HotspotSupport::getJavaTraceAsync()'s PROBE_SP loop, or
+    // unwindStub()/unwindCompiled() writing pc()/sp()/fp() by reference).
+    // Capturing is a no-op when ucontext is null (pc()/sp()/fp() themselves
+    // are not null-safe); restore() delegates to StackFrame::restore() above,
+    // which already is.
+    //
+    // Must be restored via an explicit restore() call, never a destructor:
```
jbachorik: Why is there desctructor defined and wired than? Should the comment be updated?

ddprof-lib/src/main/cpp/safeAccess.h
```diff
@@ -92,6 +94,32 @@ class SafeAccess {
...
+   * Safely writes a 32-bit value to the given address.
+   *
+   * <p>CRITICAL: This function MUST NOT be inlined. See safeFetch32 for why --
+   * the same handle_safefetch fault-redirect relies on the store happening at
+   * this function's own stable address.
+   *
+   * @param ptr Address to write to (may be invalid)
+   * @param value Value to store at ptr
+   * @return true if the store succeeded, false if the write faulted
+   */
```
jbachorik, line 107: This is javadoc format, right? Probably not good for C/C++ sources.

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
```diff
@@ -1167,14 +1172,21 @@ int HotspotSupport::getJavaTraceAsync(void *ucontext, ASGCT_CallFrame *frames,
...
     const void* pc = anchor->lastJavaPC();
     if (sp != 0 && pc == NULL) {
       // We have the last Java frame anchor, but it is not marked as walkable.
-      // Make it walkable here
-      pc = ((const void**)sp)[-1];
-      anchor->setLastJavaPC(pc);
+      // Make it walkable here.
+      // sp comes straight from the anchor with no validation; fault-inject it
+      // so the unguarded dereference below exercises the sigsetjmp/siglongjmp
+      // recovery path installed by the caller (walkJavaStack) instead of only
```
jbachorik: Isn't this installed by `withUcontextFaultRecovery` instead of `walkJavaStack`?

jbachorik:
A few explanatory comments, but otherwise looking good

ddprof-lib/src/main/cpp/stackFrame.h
zhengyu123: Updated comments - `Must be restored via an explicit resotore() call on recovery path`

ddprof-lib/src/main/cpp/safeAccess.h
zhengyu123: Updated the format

ddprof-lib/src/main/cpp/hotspot/hotspotSupport.cpp
zhengyu123: Right, updated

jbachorik (approved):
Looks good

ddprof-lib/src/main/cpp/hotspot/hotspotStackFrame.h
```diff
@@ -15,6 +15,47 @@ class HotspotStackFrame : public StackFrame {
...
+                assert(anchor != nullptr);
+                _anchor = anchor;
+                _anchor_pc = pc;
+            }
+
+            virtual void restore() override {
+                StackFrame::RegisterSnapshot::restore();
+                if (_anchor != nullptr) {
+                    // Safe store, cannot fault
+                    bool ret = _anchor->setLastJavaPC<true /*safe store*/>(_anchor_pc);
```
rkennke, line 53: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM]** If `setLastJavaPC<true>()` itself faults (e.g. the anchor's backing `JavaThread` memory became invalid between `saveJavaAnchor()` and this `restore()` — exactly the kind of race this fault-recovery machinery exists to guard against), `ret` is `false`. In debug builds the `assert` catches it; in release builds the failure is silently swallowed and `_anchor` is still nulled out as if the restore succeeded, leaving the real anchor's `lastJavaPC` in whatever state it was in when the fault struck.

Since a raw log/warning isn't safe to add here (this can run from the sigsetjmp recovery branch, which inherits the same async-signal-safety constraints as the rest of this fault path — no malloc, no locks), the idiomatic fix already used elsewhere in this PR is a dedicated counter:

```cpp
bool ret = _anchor->setLastJavaPC<true /*safe store*/>(_anchor_pc);
assert(ret && "Failed to restore lastJavaPC");
if (!ret) {
    Counters::increment(ANCHOR_RESTORE_FAILED);  // new entry in counters.h, alongside SAFESTORE_FAILED
}
_anchor = nullptr;
```

That makes a production failure observable via the existing counter-reporting path instead of disappearing silently.

ddprof-lib/src/main/cpp/hotspot/vmStructs.h
```diff
@@ -737,22 +737,28 @@ DECLARE(VMJavaFrameAnchor)
...
 
     NOADDRSANITIZE const void* lastJavaPC() {
         assert(_anchor_pc_offset >= 0);
-        return SafeAccess::loadPtr((void**) at(_anchor_pc_offset), nullptr);
+        return *(void**) at(_anchor_pc_offset);
     }
 
-    void setLastJavaPC(const void* pc) {
+    template <bool SafeStore = true>
+    bool setLastJavaPC(const void* pc) {
```
rkennke, line 754: 🗿 🤖 🔴

**[Sphinx Review — MEDIUM — test-adequacy]** Two mutation-adequacy gaps around the new anchor-restore path, neither exercised by `FaultInsideProfilerRangeRecoversAndRestoresJavaThreadAnchor`:

1. **Here:** flipping the `SafeStore` template argument from `true` to `false` in `HotspotStackFrame::RegisterSnapshot::restore()`'s call to `setLastJavaPC<true>()` would go undetected. The test only checks that `anchor->lastJavaPC()` ends up equal to `original_pc` — it doesn't distinguish the `SafeAccess::storePtr()` branch from the plain-store branch, and both produce the same final value in this scenario.
   - Suggestion: add a test that verifies `SafeStore=true` actually routes through `SafeAccess::storePtr()` (e.g. by pointing the anchor at an unmapped/read-only address and confirming the store fails safely instead of crashing) versus `SafeStore=false` using a direct store.

2. **`hotspotStackFrame.h`'s `restore()`:** removing `_anchor = nullptr;` after the restore would also go undetected. `restore()` runs twice for this test (once explicitly in the sigsetjmp recovery branch, once implicitly via the destructor); without the guard, the second call would just call `setLastJavaPC(_anchor_pc)` again with the same already-correct value, so the final assertion still passes — coincidentally, not because cleanup actually happened.
   - Suggestion: add an assertion or separate test that pins `_anchor` being cleared after the first `restore()` call (e.g. by making a second `restore()` call observable, or exposing `_anchor`'s state to the test).

(High confidence on both — verified independently by tracing the double-`restore()` call sequence and the template dispatch, not just taking the mutation-testing signal at face value.)

ddprof-lib/src/main/cpp/hotspot/hotspotStackFrame.h
zhengyu123: Fixed as suggested

ddprof-lib/src/main/cpp/hotspot/vmStructs.h
zhengyu123: Actually, the scenarios are covered by fault-injection. `anchor` and `VMJavaFrameAnchor::setLastJavaPC()` both go through `VMStructs::at()` call and the faults are injected there.

rkennke (approved):
Looks good! The remaining failure in integration-test-x64-musl seems to be a flaky test.

PR #787, opened 2026-09-11
[PROF-15786] Reproduce post-JNI Method invalidation

r1viollet:
**What does this PR do?**:

Adds a regression test for a raw `Method*` access after `HotspotSupport::resolve()` leaves its crash-protected metadata walk.

The test makes the fake JVMTI `GetClassMethods` callback invalidate the page containing the original `Method*`. The following `vm_method->validatedId()` call otherwise aborts in assertion-enabled builds.

A second commit contains a candidate fix: use the existing safe `load_at_offset` path in `VMMethod::id()` so the read does not pass through `VMStructs::at()` and its readability assertion.

**Motivation**:

PR #743 protects the initial metadata walk, but the `<clinit>` fallback retains and reads `vm_method` after JNI/JVMTI work and after crash protection has been removed. This draft provides a concrete reproducer and a narrow candidate fix so the lifetime assumptions and preferred resolution can be reviewed.

**Additional Notes**:

- The invalidation is synthetic. Whether class unloading, redefinition, or obsolete-method reclamation can produce the same timing in a live JVM still needs investigation.
- Without the candidate fix, the test observes `SIGABRT` from the readability assertion in `VMStructs::at()`.
- The candidate fix is intentionally limited to the first `VMMethod::id()` load. This draft does not assume that it is the preferred final design.
- Related change: https://github.com/DataDog/java-profiler/pull/743

**How to test the change?**:

```bash
./.claude/commands/build-and-summarize :ddprof-lib:gtestDebug_hotspotMethodId
./.claude/commands/build-and-summarize :ddprof-lib:gtestRelease_hotspotMethodId
```

Both targeted suites pass on Linux with the reproducer enabled and the candidate fix applied.

**For Datadog employees**:
- [x] JIRA: [PROF-15786](https://datadoghq.atlassian.net/browse/PROF-15786)

[PROF-15786]: https://datadoghq.atlassian.net/browse/PROF-15786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/ddprof-lib/src/test/cpp/hotspotMethodId_ut.cpp
+++ b/ddprof-lib/src/test/cpp/hotspotMethodId_ut.cpp
@@ -33,6 +33,8 @@ public:
     static void setHotspot(bool value) { VM::_hotspot = value; }
     static JavaVM* getVm() { return VM::_vm; }
     static void setVm(JavaVM* vm) { VM::_vm = vm; }
+    static jvmtiEnv* getJvmti() { return VM::_jvmti; }
+    static void setJvmti(jvmtiEnv* jvmti) { VM::_jvmti = jvmti; }
 };
 
 // Test-only friend accessor for VMStructs protected static offsets.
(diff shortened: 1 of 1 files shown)


zhengyu123:
I am confused - what's the purpose of the test? to verify that without longjmp protection, `vm_method->validatedId()` call can result in crash?

Current `vm_method->validatedId()` implementation should not crash, as all memory accesses via `SafeAccess`, but I intent to change to plain accesses with longjmp protection, then fault-injection should be able to trigger the crash to test longjmp recovery.

r1viollet:
> I am confused - what's the purpose of the test? to verify that without longjmp protection, `vm_method->validatedId()` call can result in crash?
> 
> Current `vm_method->validatedId()` implementation should not crash, as all memory accesses via `SafeAccess`, but I intent to change to plain accesses with longjmp protection, then fault-injection should be able to trigger the crash to test longjmp recovery.

@zhengyu123 yes, precisely.

The aim is to give you test cases for your refactors. There is no fix proposal here.
Enabling the reproducer demonstrates the SIGABRT in VMStructs::at() after JVMTI invalidates the raw Method* page.

zhengyu123:
> > I am confused - what's the purpose of the test? to verify that without longjmp protection, `vm_method->validatedId()` call can result in crash?
> > Current `vm_method->validatedId()` implementation should not crash, as all memory accesses via `SafeAccess`, but I intent to change to plain accesses with longjmp protection, then fault-injection should be able to trigger the crash to test longjmp recovery.
> 
> @zhengyu123 yes, precisely.
> 
> The aim is to give you test cases for your refactors. There is no fix proposal here. Enabling the reproducer demonstrates the SIGABRT in VMStructs::at() after JVMTI invalidates the raw Method* page.

@r1viollet We already inject faults at the callsites (https://github.com/DataDog/java-profiler/blob/main/ddprof-lib/src/main/cpp/hotspot/vmStructs.inline.h#L36 and https://github.com/DataDog/java-profiler/blob/main/ddprof-lib/src/main/cpp/hotspot/vmStructs.inline.h#L42), so that some of the faults should propagate to `vm_method->validatedId()`

r1viollet:
thanks for checking! closing this as the fault injection should already cover this.

PR #794, opened 2026-09-16
refactor(sampling): replace std::random_device with the existing xorshift64

rkennke:
**What does this PR do?**:

Gives the profiler one random source instead of two. `xorshift.h` holds the
generator and the draws built on it; the four places that had their own copy of
the recurrence now call it, and the three that used `std::mt19937` seeded from
`std::random_device` are converted.

| | before | after |
|---|---|---|
| `PoissonSampler` | xorshift64, inline | `xorshift::` |
| `ProfiledThread::_fi_rng` | xorshift64, inline | `xorshift::` |
| `faultInjection` fallback | xorshift64, inline | `xorshift::` |
| `LivenessTracker` | `mt19937` + `uniform_real_distribution`, per thread, heap | `ThreadLocal<u64>` |
| `ReservoirSampler` | `mt19937` + two distributions | `xorshift::` |
| `BaseWallClock` | `mt19937` + `normal_distribution` | `xorshift::nextNormal` |

`<random>` is gone from the library.

**Motivation**:

`poissonSampler.h` already argued the case, and the argument was never specific
to that one sampler:

- **Allocation and exceptions.** `std::random_device` and the distribution
  wrappers may allocate and may throw. `LivenessTracker::track` runs on
  allocation callbacks on arbitrary application threads, where neither is safe.
- **Seeding.** `std::random_device` may block or return low-entropy values
  depending on the host, and *which* it does is decided by the libstdc++ the
  library was compiled against, not by anything in this source. That made a
  sampling-relevant behaviour a property of the build image.
- **Size.** `std::mt19937` carries ~2.5 KB of state. `LivenessTracker` held one
  per thread, on the heap, plus a distribution object — and the lazy-create /
  explicit-free machinery that existed only because they were heap-allocated.

*Side effect that matters elsewhere.* `std::random_device` is also why the
shipped library referenced `getentropy@GLIBC_2.25`. That reference is not in our
code: libstdc++ decides at *its* build time how to seed `std::random_device`, and
from GCC 12 on it calls `getentropy` where the glibc it was configured against
has it — so `-static-libstdc++` copies `random.o` and that undefined reference
into `libjavaProfiler.so`. glibc has only one version of `getentropy`, so there
is no older symbol to pin it to. Removing the three `std::random_device` uses
keeps `random.o` out of the link entirely, which is what lets #790 hold the
glibc 2.17 floor while building on a supported EL8 toolchain — the remaining
above-2.17 reference there is `expf`, which is in our own code and can be
`.symver`-pinned. This PR stands on the reasons above regardless of that
decision; the floor is a beneficiary, not the motivation.

**Additional Notes**:

*Seeding is no longer OS-backed.* A stream is an identity mixed with a sequence
number — an object address with an epoch, a thread id with a tick. That yields
independent streams per thread and per recording without an entropy source.
Sampling needs independence, not unpredictability, so this is a deliberate
trade rather than an oversight: two processes started on the same tick with
ASLR disabled would correlate. Nothing here is used for anything security-
relevant.

*`ReservoirSampler` now takes its stream as a constructor argument.* Seeding
from the instance address alone meant every sampler built at one call site got
the same stream, so consecutive recordings sampled identically. Making it a
required argument means a caller cannot silently inherit a fixed stream; the
wall-clock loop passes `TSC::ticks()`.

*The wall-clock dither keeps its shape.* It was `normal_distribution(interval,
interval/10)` and still is, drawn by Box-Muller. The alternatives (a flat
dither, sum-of-uniforms) would have changed observable behaviour for no reason.

*`LivenessTracker`'s subsample decision is now an integer compare* against a
threshold computed once in `initialize()`, the same device `faultInjection`
already used for its probability tiers.

**How to test the change?**:

`ddprof-lib/src/test/cpp/xorshift_ut.cpp` — 23 assertions, picked up
automatically by the gtest plugin:

- **seed**: never returns 0 (a fixed point that would freeze the stream),
  including when the two inputs cancel exactly; distinct identities give
  distinct streams; small sequence numbers (0, 1, 2 …) are spread rather than
  differing in a few low bits
- **next**: advances, never returns 0, no repeats over a short run
- **toUnitDouble**: strictly inside (0, 1) at both extremes of the input range
  and over 10k draws, with `log()` finite throughout; roughly uniform by decile
  and mean
- **threshold**: saturates at both ends rather than casting an out-of-range
  double; monotonic; fires at the requested rate (0.01 … 0.9)
- **boundedIndex**: stays in range for every n up to 64; handles n == 0; covers
  and balances the range; does not favour low indices for a non-power-of-two n,
  which a modulo mapping would
- **nextNormal**: matches the requested mean and stddev over 100k draws;
  returns the mean for stddev 0
- **ReservoirSampler**: keeps everything when the input fits; fills to capacity
  when it does not; writes every slot; distinct streams give distinct samples
  and the same stream gives the same samples; empty input; reuse across calls

Two defects in the first draft were found by these tests and fixed:
`toUnitDouble(UINT64_MAX)` returned exactly 1.0 (`1 - 2^-65` rounds up), which
would make `log(1 - u)` infinite for a later caller; and every `ReservoirSampler`
at one call site produced byte-identical output. Both have tests pinned to the
specific failure.

Also run: the full `gtestDebug` suite, and `gtestDebug_faultInjection_ut` under
`-PenableFaultInjection` — the three tier-rate tests there exercise the
consolidated generator statistically, and the fault-injection paths are not
compiled by the default build.

**For Datadog employees**:


ddprof-lib/src/main/cpp/xorshift.h
```diff
@@ -0,0 +1,166 @@
...
+    return 0;
+  }
+  // 2^64 is not representable as u64, and casting a double at or above it is
+  // undefined, so saturate before the cast rather than after.
+  const double scale = 18446744073709551616.0; // 2^64
+  double scaled = p * scale;
+  if (scaled >= (double)UINT64_MAX) {
+    return UINT64_MAX;
+  }
+  return (u64)scaled;
```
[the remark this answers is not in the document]
rkennke: Valid, and fixed in 70d5cc2f0 (pushed after this comment was written).

`threshold()`'s guard is now inverted so an unordered value takes the reject
branch instead of reaching the cast:

```cpp
if (!(p > 0.0)) {
  return 0;
}
```

Covered by `Xorshift.ThresholdRejectsNaN` in `xorshift_ut.cpp`.

On the behavioural half of the point: a `nan` ratio still retains every sample,
as before. `LivenessTracker::track()` gates the subsampling draw on
`_subsample_ratio < 1.0`, which is also false for NaN, so the threshold is never
consulted on that path — the fix removes the undefined cast at `initialize()`
without changing what a NaN ratio does.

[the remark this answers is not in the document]
rkennke: Valid, and fixed in 70d5cc2f0 (pushed after this comment was written).

`threshold()`'s guard is now inverted so an unordered value takes the reject
branch instead of reaching the cast:

```cpp
if (!(p > 0.0)) {
  return 0;
}
```

Covered by `Xorshift.ThresholdRejectsNaN` in `xorshift_ut.cpp`.

On the behavioural half of the point: a `nan` ratio still retains every sample,
as before. `LivenessTracker::track()` gates the subsampling draw on
`_subsample_ratio < 1.0`, which is also false for NaN, so the threshold is never
consulted on that path — the fix removes the undefined cast at `initialize()`
without changing what a NaN ratio does.

ddprof-lib/src/main/cpp/xorshift.h
```diff
@@ -0,0 +1,170 @@
...
+  u64 state = identity ^ (sequence * KNUTH);
+  return state != 0 ? state : 1;
+}
+
+/**
+ * One xorshift64 step, in place. Async-signal-safe: no allocation, no locks,
+ * no libc calls. @p state must be non-zero, which seed() guarantees.
+ */
+inline u64 next(u64 &state) {
+  state ^= state << 13;
```
zhengyu123, line 99: Add assertion for `state != 0`, before and after modification.

ddprof-lib/src/main/cpp/livenessTracker.cpp
```diff
@@ -272,6 +272,9 @@ Error LivenessTracker::initialize(Arguments &args) {
   }
 
   _subsample_ratio = args._live_samples_ratio;
+  // Precomputed here so the per-allocation decision in track() is an integer
+  // compare rather than a double multiply.
+  _subsample_threshold = xorshift::threshold(_subsample_ratio);
```
jbachorik: Would it be possible to have some guardrails to make sure `_subsample_ratio` and `_subsample_threshold` can not get out of sync?

jbachorik (approved):

ddprof-lib/src/main/cpp/livenessTracker.cpp
rkennke: Good catch — done in b631951d5.

The two fields are now one type whose only constructor derives the threshold
from the ratio, so there is no way to write one without the other:

```cpp
struct SubsampleRate {
  double ratio;
  u64 threshold;
  explicit SubsampleRate(double subsample_ratio)
      : ratio(subsample_ratio), threshold(xorshift::threshold(subsample_ratio)) {}
};
```

`initialize()` replaces the pair in a single assignment
(`_subsample = SubsampleRate(args._live_samples_ratio)`), and `track()` /
`Recording` read `_subsample.threshold` / `_subsample.ratio`. A partial update
is no longer expressible rather than merely discouraged.

Worth saying why it mattered more than it looks: the failure mode was silent in
a bad direction — allocations sampled at the stale rate while the JFR chunk
reported the new ratio, so the data would have looked self-consistent while
being wrong.

Three tests in `xorshift_ut.cpp`, asserted against observed behaviour rather
than against `xorshift::threshold` itself, so they still fail if the derivation
stops honouring the advertised rate: the keep-rate over 200k draws matches the
ratio at 0.01/0.1/0.5/0.9, reassignment carries both halves, and the degenerate
ratios (1, 0, NaN) all yield defined thresholds.

ddprof-lib/src/main/cpp/xorshift.h
```diff
@@ -0,0 +1,170 @@
...
+ * triple gets an independent stream. Streams are reproducible for a given
+ * address and sequence, which is deliberate: it needs no entropy source, and
+ * sampling decisions require independence rather than unpredictability.
+ */
+namespace xorshift {
+
+// Fibonacci hashing constant. Spreads low-entropy inputs -- a small thread id,
+// an epoch counter that starts at 0 -- across the full 64-bit range before they
+// are mixed with the identity.
+static constexpr u64 KNUTH = (u64)KNUTH_MULTIPLICATIVE_CONSTANT;
```
zhengyu123, line 59: Nit: there is a duplicated `KNUTH` declaration in `faultinjection.cpp`

ddprof-lib/src/main/cpp/threadLocalData.h
zhengyu123: Calling `xorshift::seed(...)` instead

zhengyu123 (approved):
LGTM, just a few nits.

ddprof-lib/src/main/cpp/xorshift.h
rkennke: Added in 7237b42b4, both positions:

```cpp
inline u64 next(u64 &state) {
  assert(state != 0 && "xorshift64 state is the recurrence's fixed point");
  state ^= state << 13;
  state ^= state >> 7;
  state ^= state << 17;
  assert(state != 0 && "xorshift64 stepped to its fixed point");
  return state;
}
```

The second one is the more interesting of the two: the recurrence is a
bijection on the non-zero states, so a valid state can never step to 0 —
asserting it states that property rather than only guarding the input.

Release builds compile with `-DNDEBUG` (ConfigurationPresets.kt), so neither
check reaches the sampling path; they are active in debug, asan, tsan and the
gtest configs.

While here: the non-zero rule had three separate implementations (inside
`seed()`, `setFiRng`'s `seed ? seed : 1`, and `faultinj::nextRandom`'s
`if (x == 0) x = 1`). It is now `xorshift::nonZero()` alone, with `seed()`
expressed in terms of it.

ddprof-lib/src/main/cpp/xorshift.h
rkennke: Fixed in 7237b42b4 — `faultInjection.cpp` now uses `xorshift::KNUTH` at both of
its sites (`g_fallback_rng`'s initialiser and the per-call-site hash in
`shouldFire`), and its local `constexpr` is gone.

My oversight: I pointed `xorshift.h` at `common.h`'s existing
`KNUTH_MULTIPLICATIVE_CONSTANT` precisely to avoid adding another copy, then
left the one in `faultInjection.cpp` in place.

One deliberate exception: `mallocTracer.cpp:183,187` still spell the constant
out. That is splitmix64 seeding a xoroshiro128+ generator, where the same
golden-ratio constant is the standard increment — a different algorithm that
happens to share the number, so routing it through `xorshift::` would suggest a
relationship that is not there. Happy to revisit if you would rather it were
shared.

ddprof-lib/src/main/cpp/threadLocalData.h
rkennke: Good catch, and it was a real gap rather than a style point — fixed in
7237b42b4.

`ProfiledThread` has two `_fi_rng` seeding sites, `claimAcquire()` in
threadLocalData.inline.h and this constructor, and I had only converted the
first. So a hand-rolled copy of the address-XOR-Knuth-hashed-tid seed (plus its
own `if (_fi_rng == 0)` guard) survived here, while the PR described every copy
as going through the shared helper. Both now call
`xorshift::seed((u64)(uintptr_t)this, (u64)tid)`.

Verified under `-PenableFaultInjection`, where these paths actually compile:
the seven FaultInjectionTest cases pass, including the three tier-rate tests
that exercise the reseeded stream statistically.

PR #796, opened 2026-09-17
JFR event types and emission for reference chains

jbachorik:
**What does this PR do?**:
Adds the JFR plumbing for reference-chain events: the `ReferenceChainEvent` / `ReferenceChainAbandonedEvent` payloads (`event.h`), their dynamic JFR metadata and constant-pool handling (`jfrMetadata.*`), the FlightRecorder emission paths that serialize chain events and per-hop edge labels into recording buffers (`flightRecorder.*`), and the `referencechains` `Arguments` block (`arguments.*`) that configures the feature.

**Motivation**:
Part 2 of the stacked series for reference chains (PROF-15341). Emission is pull-style — the later tracker layer snapshots events and hands them to FlightRecorder — so this layer is independently reviewable and compiles/links without the tracker.

**Additional Notes**:
Stacked series for PROF-15341. Merge order: #804 (scan-build crashNow fix) -> this PR -> #797 -> #798 -> #799 -> #800 -> #801 -> #802. #795 (UBSan ELF fix) is already merged. Until #804 lands, this PR's diff additionally shows the small faultInjection.cpp change from #804.

**How to test the change?**:
The JFR round-trip unit tests for chain events land in the gtest PR later in this stack; `buildDebug` compiles and links this layer standalone.

**For Datadog employees**:
- [x] JIRA: [PROF-15341]

[PROF-15341]: https://datadoghq.atlassian.net/browse/PROF-15341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ddprof-lib/src/main/cpp/flightRecorder.h
```diff
@@ -45,6 +45,20 @@ const int JFR_EVENT_FLUSH_THRESHOLD = RECORDING_BUFFER_LIMIT;
...
+// table's own defensive walk bound (FrontierTable::reconstructChain(),
+// referenceChains.h) is maxCapacity(), which can run into the tens of
+// thousands of entries, and neither that cap nor _hop_cap is itself
+// range-validated against a buffer-safe maximum (see arguments.cpp's own
+// sub-option parsing). recordReferenceChain() truncates event->_chain to
+// this many entries before writing, so its own worst-case size never
+// depends on trusting either of those upstream caps to stay small - a chain
+// longer than this is still truncated defense-in-depth even if a caller
+// changes those caps later.
+const int MAX_REFERENCE_CHAIN_EVENT_HOPS = 4096;
```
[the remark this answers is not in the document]
jbachorik: Fixed in `339be6e3d`: `MAX_REFERENCE_CHAIN_EVENT_HOPS` is no longer a fixed 4096 — it is derived from `RECORDING_BUFFER_LIMIT` minus the event's fixed fields (size prefix, type id, timestamps, target tag, depth, totalHops, counts, rootKind string), divided by the per-hop worst case (one var32 chain entry + one putUtf8 edge label at `MAX_REFERENCE_CHAIN_EDGE_LABEL`). A full-cap event now always fits the buffer, so the reservation can never underflow. Pinned by a new byte-level boundary test (`TruncatesOversizeChainAndKeepsEmittedLabels` in the moved `referenceChainJfrRoundtrip_ut.cpp`) that serializes a chain over the cap with 96-byte labels, then walks the buffer and asserts the emitted size stays within `RECORDING_BUFFER_LIMIT`.

ddprof-lib/src/main/cpp/event.h
```diff
@@ -87,9 +89,77 @@ class ObjectLivenessEvent : public Event {
   u64 _skipped;
   u64 _start_time;
   u64 _age;
+  int64_t leak_tag;  // 0 = untagged; leak tag from LivenessTracker pool
```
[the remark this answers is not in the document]
jbachorik: The flush path already assigns this field: `LivenessTracker::flush_table()` builds the `ObjectLivenessEvent` and sets `event.leak_tag = _table[i].leak_tag` (livenessTracker.cpp), and every table entry gets `_table[idx].leak_tag = 0` at insert in `track()` (release paths re-zero it) — so the read was defined, not indeterminate. That said, the default-construction hole is real for any future path that forgets the assignment, so `leak_tag` is now default-initialized to 0 (`int64_t leak_tag = 0;`) — an unset field serializes a defined "untagged" value. Pinned by `DefaultLivenessEventHasZeroLeakTag`.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -2332,11 +2339,154 @@ void Recording::recordHeapLiveObject(Buffer *buf, int tid, u64 call_trace_id,
...
+  // rootKindName() never returns a string longer than
+  // "first_observed_via:stack_local" (31 bytes) - reserve a fixed, generous
+  // 32 bytes for its putUtf8() length prefix + payload rather than computing
+  // strlen() up front.
+  const char *root_kind_name = rootKindName(event->_root_kind);
+  // Per-hop edge labels: same emitted_size as the chain, each bounded by
+  // MAX_REFERENCE_CHAIN_EDGE_LABEL (event.h) + putUtf8's 1-byte encoding
+  // tag and up to 5-byte varint length prefix.
+  const u32 edge_count =
+      event->_edges.size() == emitted_size ? emitted_size : 0;
```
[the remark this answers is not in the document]
jbachorik: Fixed in `339be6e3d`: the label count is now `min(_edges.size(), emitted_size)` instead of the equality gate. Labels align with the chain's leaf-first element order (edges[i] describes the hop into chain[i]), and truncation keeps the FIRST emitted_size chain entries, so the first emitted_size labels are exactly the labels of the emitted range — truncation now loses only the root-side hops and their labels. The empty-count case remains for events whose label collection never ran (empty _edges). Covered by the same byte-walk test, which asserts edge_count == emitted_size for a truncated chain.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -2332,11 +2339,154 @@ void Recording::recordHeapLiveObject(Buffer *buf, int tid, u64 call_trace_id,
...
+  case 26:
+    return "thread";
+  case 27:
+    return "other";
+  default:
+    return "unknown";
+  }
+}
+
+void Recording::recordReferenceChain(Buffer *buf, ReferenceChainEvent *event) {
```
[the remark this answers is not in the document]
jbachorik: The stacked series carries these tests in the gtest and Java-test layers (#799: `referenceChainJfrRoundtrip_ut.cpp`, `arguments_ut.cpp`; #800: `ReferenceChainJfrParserTest`), but per this review the two unit-test files now live in THIS PR — they test exactly this layer's encodings. The round-trip test was rewritten to construct events directly instead of going through the tracker (which does not exist yet at this layer), and three boundary tests were added: oversize-chain truncation with per-byte verification of the size prefix / totalHops / emitted size / edge label count against the buffer walk, the size-prefix invariant, and the default leak tag. Full `gtestDebug` green (625 cases, 0 failures).

ddprof-lib/src/main/cpp/event.h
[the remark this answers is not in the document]
jbachorik: The read was defined, not indeterminate: `LivenessTracker::flush_table()` sets `event.leak_tag = _table[i].leak_tag` when building the event, and `track()` zeroes `_table[idx].leak_tag` at insert (release paths re-zero it), so the field was always initialized before `recordHeapLiveObject()` read it. Nonetheless, `leak_tag` is now default-initialized to 0 in `event.h` (`339be6e3d`), so any construction path that forgets the assignment serializes a defined "untagged" value — pinned by the `DefaultLivenessEventHasZeroLeakTag` unit test in this PR.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -2332,11 +2339,154 @@ void Recording::recordHeapLiveObject(Buffer *buf, int tid, u64 call_trace_id,
...
+  // Per-hop edge labels: same emitted_size as the chain, each bounded by
+  // MAX_REFERENCE_CHAIN_EDGE_LABEL (event.h) + putUtf8's 1-byte encoding
+  // tag and up to 5-byte varint length prefix.
+  const u32 edge_count =
+      event->_edges.size() == emitted_size ? emitted_size : 0;
+  const char *edge_labels[MAX_REFERENCE_CHAIN_EVENT_HOPS];
+  for (u32 i = 0; i < edge_count; i++) {
+    edge_labels[i] = event->_edges[i].c_str();
+  }
+  flushIfNeeded(
```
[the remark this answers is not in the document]
jbachorik: Fixed in `339be6e3d`: `MAX_REFERENCE_CHAIN_EVENT_HOPS` is now derived from `RECORDING_BUFFER_LIMIT` minus the event's fixed serialization overhead, divided by the per-hop worst case (var32 chain entry + full-length putUtf8 edge label), instead of a fixed 4096 that permitted a ~438 KB event. A full-cap event always fits the recording buffer, so the reservation cannot underflow. Verified by `TruncatesOversizeChainAndKeepsEmittedLabels` in `referenceChainJfrRoundtrip_ut.cpp`: a chain over the cap with 96-byte labels is serialized and the event bytes are walked to assert the write stays within `RECORDING_BUFFER_LIMIT`.

ddprof-lib/src/main/cpp/arguments.cpp
```diff
@@ -444,6 +465,98 @@ Error Arguments::parse(const char *args) {
...
+        char *cursor = config;
+        while (cursor != NULL) {
+          char *next = strchr(cursor, ':');
+          if (next) {
+            *(next++) = 0;
+          }
+          char *eq = strchr(cursor, '=');
+          if (eq) {
+            *(eq++) = 0;
+            // Floor every sub-option at the parse boundary rather than
```
rkennke: That comment is *way* to verbose to justify why the values need to be floored and ceiling'ed. Either make them very concise or remove it altogether, it's kinda obvious that it's not a good idea to allow values outside the reasonable bounds here.

ddprof-lib/src/main/cpp/arguments.h
```diff
@@ -23,12 +23,127 @@
...
 
+#include "arch.h"
+
 const long DEFAULT_CPU_INTERVAL = 10 * 1000 * 1000;  // 10 ms
 const long DEFAULT_WALL_INTERVAL = 50 * 1000 * 1000; // 50 ms
 const long DEFAULT_ALLOC_INTERVAL = 524287;          // 512 KiB
 const int DEFAULT_WALL_THREADS_PER_TICK = 16;
 const int DEFAULT_JSTACKDEPTH = 2048;
 
+// Every constant below is a provisional default pending empirical
```
rkennke: Similar problem in this file: way too much comments explaining the same fact over and over again. Make it short and concise once, and drop all the other repetitions.

ddprof-lib/src/main/cpp/arguments.h
```diff
@@ -177,6 +292,25 @@ class Arguments {
   double _live_samples_ratio;
   bool _record_heap_usage;
   bool _gc_generations;
+  // Reference-chain tracking (PROF-15341 - see
+  // doc/architecture/LiveHeapReferenceChains-ImplementationPlan.md and
```
rkennke: I don't think those docs exist in this PR, and they shouldn't, but the reference to them also needs to go. Also, I don't think we want Jira-ticket references in our code.

ddprof-lib/src/main/cpp/event.h
```diff
@@ -87,9 +89,81 @@ class ObjectLivenessEvent : public Event {
...
+  u32 _depth;
+  u8 _root_kind;
+  std::vector<u32> _chain;
+  // Retention-edge label per hop, ALIGNED with _chain's leaf-to-root order
+  // (_edges[i] = the edge by which _chain[i] is retained - the field name of
+  // its parent hop for FIELD/STATIC_FIELD edges, the edge-kind label
+  // otherwise; ReferenceChainTracker::fillHopEdgeLabels). Empty for events
+  // built before the edge-label change or when label resolution is
+  // unavailable (partial mock environments).
+  std::vector<std::string> _edges;
```
rkennke: If _chain and _edges need to go hand-in-hand, then maybe make it a single vector of struct {u32; std::string } instead?

ddprof-lib/src/main/cpp/event.h
```diff
@@ -87,9 +89,81 @@ class ObjectLivenessEvent : public Event {
...
+
+  ReferenceChainEvent()
+      : Event(), _start_time(0), _target_tag(0), _depth(0), _root_kind(0) {}
+};
+
+// Search-level abandonment signal (design doc's Termination section:
+// "explicit reporting of abandoned searches ... no silent truncation").
```
rkennke: again, don't reference a non-existing design doc.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -48,14 +48,18 @@
...
 
-// JVM spec SS4.7.3 caps a method's bytecode (code_length) at 65535 bytes (u2),
-// so a well-formed LineNumberTable can never have more entries than that.
-// Used to sanity-bound line_number_table_size before it drives the byte-count
-// passed to SafeAccess::safeCopy(): if GetLineNumberTable()
-// returns a corrupted pointer for a stale jmethodID (see the TOCTOU race
-// documented in fillJavaMethodInfo below), the paired out-param size is just
-// as likely to be corrupted, and an implausible size should be rejected
-// before it is trusted to compute a byte range.
+// JVM spec SS4.7.3 caps a method's bytecode (code_length) at 65535 bytes (u2).
```
rkennke: This change seems unrelated?

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -2332,11 +2339,160 @@ void Recording::recordHeapLiveObject(Buffer *buf, int tid, u64 call_trace_id,
...
+// label for datadog.ReferenceChain's rootKind field - only the values
+// heapReferenceCallback() can actually produce (a root reference's own
+// jvmtiHeapReferenceKind, or JVMTI_HEAP_REFERENCE_STATIC_FIELD for the
+// "referrer is a pre-tagged class" root-like case, see that method's own
+// comment) have entries; anything else (including 0, root_kind's
+// "not set" default) reports "unknown" rather than crashing on an
+// out-of-range index.
+//
+// STACK_LOCAL (24) and JNI_LOCAL (25) are labeled "first_observed_via:..."
+// rather than plain "stack_local"/"jni_local" (design doc's "Honest labeling
```
rkennke: stale referencs to design doc, again

rkennke (changes requested):
Nice change. I did a manual review, see below some findings.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -2332,11 +2339,160 @@ void Recording::recordHeapLiveObject(Buffer *buf, int tid, u64 call_trace_id,
...
+  }
+  // Edges array, LAST so its metadata position (after "chain", jfrMetadata.cpp)
+  // matches the write order - the leakTag field-order invariant
+  // (find-leaktag-jfr-field-misalignment) generalized. The labels align with
+  // the chain's element order (see edge_count above), so the emitted range
+  // carries its labels even when the chain was truncated at its root side;
+  // a count of 0 means label collection never ran for this event.
+  buf->putVar32(edge_count);
+  for (u32 i = 0; i < edge_count; i++) {
+    buf->putUtf8(edge_labels[i]);
```
rkennke, line 2462: 🗿 🤖 🔴

**[Sphinx Review — HIGH]** recordReferenceChain() reserves MAX_REFERENCE_CHAIN_EDGE_LABEL (96) + 6 bytes per edge label, but writes labels with putUtf8(const char*), which truncates at MAX_STRING_LENGTH (8191). Any label longer than 96 bytes (a real Java field name easily can be) under-reserves by up to ~8 KB per hop and writes past the recording buffer - Buffer::put()'s only guard is an assert() that is compiled out in release builds. This is exactly the 'do not trust an upstream cap' argument the function's own comment makes for the chain length, but it is not applied to the labels.

**Suggestion:** Clamp each label's length to MAX_REFERENCE_CHAIN_EDGE_LABEL at the write site by using the two-argument putUtf8(v, len) overload, so the serializer's worst case matches its own reservation regardless of what the producer hands it.

```suggestion
    buf->putUtf8(edge_labels[i], (u32)(strlen(edge_labels[i]) < MAX_REFERENCE_CHAIN_EDGE_LABEL ? strlen(edge_labels[i]) : MAX_REFERENCE_CHAIN_EDGE_LABEL));
```

rkennke (changes requested):
🗿 🤖 🔴

ddprof-lib/src/main/cpp/arguments.cpp
jbachorik: Compressed to four lines: the hops negative-wrap case, the negative-budget collapse, the unbounded-values-into-loop-bounds case, and the one-boundary rationale. The details live in the constant comments in arguments.h now, stated once.

ddprof-lib/src/main/cpp/arguments.h
jbachorik: Done - the block now states the provisional-default fact once up front, and each constant keeps only a note that says something the name does not.

ddprof-lib/src/main/cpp/arguments.h
jbachorik: Plan-document references are gone, and so are the Jira references in code (also swept the same pattern out of the upper layers of the stack).

ddprof-lib/src/main/cpp/event.h
jbachorik: Done - ReferenceChainEvent now carries a single `std::vector<ReferenceChainHop>` (`{u32 klass_id; std::string edge_label;}`), and the JFR writer derives both the chain array and the labels array from it. Labels are all-or-none per event (canary events carry none), so the writer treats any empty label in the emitted range as "no labels" - same bytes as before.

ddprof-lib/src/main/cpp/event.h
jbachorik: Removed; the abandoned-event comment now states the rationale without citing the document.

ddprof-lib/src/main/cpp/flightRecorder.cpp
jbachorik: Reverted - it was an accuracy nitpick on a pre-existing comment and does not belong in this PR.

ddprof-lib/src/main/cpp/flightRecorder.cpp
jbachorik: Removed the design-doc reference; kept the mechanism explanation (first_observed_via labels overstate what a popped frame proves).

ddprof-lib/src/main/cpp/flightRecorder.cpp
jbachorik: Addressed in 63112c925: the writer now clamps each label with `strnlen(label, MAX_REFERENCE_CHAIN_EDGE_LABEL)` via the two-arg `putUtf8(v, len)` overload, so the serializer's worst case matches its own reservation regardless of what the producer hands it. The chain-length cap on this branch already had that defense; the label cap now does too.

ddprof-lib/src/main/cpp/flightRecorder.h
```diff
@@ -45,6 +45,40 @@ const int JFR_EVENT_FLUSH_THRESHOLD = RECORDING_BUFFER_LIMIT;
...
+// the event's fixed fields cost REFERENCE_CHAIN_EVENT_FIXED_BYTES - so a
+// full-cap event always fits inside RECORDING_BUFFER_LIMIT and the
+// reservation in recordReferenceChain() can never underflow (a fixed 4096
+// allowed a ~438 KB worst case against a 61 KB buffer - a debug assert and
+// a release-mode write past the buffer).
+const int REFERENCE_CHAIN_EVENT_FIXED_BYTES =
+    MAX_VAR32_LENGTH /* multi-byte event size prefix */ +
+    3 * MAX_VAR64_LENGTH /* type id, start_time, target_tag */ +
+    3 * MAX_VAR32_LENGTH /* depth, totalHops, chain count */ +
+    32 /* rootKind string, generous */ +
```
kaahos, line 73: nit: could this value be a constant? Then it could be used in [flightRecorder.cpp:2432](https://github.com/DataDog/java-profiler/pull/796/changes#diff-9a6a575bc5c4e9a0980683bb7bb698b45a23bdb480b6a250ae1a2a8eb9374b01R2432) to avoid potential future drifting.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -2332,11 +2335,161 @@ void Recording::recordHeapLiveObject(Buffer *buf, int tid, u64 call_trace_id,
...
+  const u32 edge_count = labeled == emitted_size ? emitted_size : 0u;
+  // Clamped to MAX_REFERENCE_CHAIN_EDGE_LABEL at the write site below (via
+  // strnlen) rather than trusted from the producer - the same "do not trust
+  // an upstream cap" defense this function applies to the chain length, so
+  // the reservation above stays an upper bound regardless of what the
+  // producer hands it.
+  const char *edge_labels[MAX_REFERENCE_CHAIN_EVENT_HOPS];
+  for (u32 i = 0; i < edge_count; i++) {
+    edge_labels[i] = event->_hops[i].edge_label.c_str();
+  }
```
kaahos:
```suggestion
```

could this be removed to avoid a stack allocation of size `MAX_REFERENCE_CHAIN_EVENT_HOPS` every time? The transformation could be performed in the next loop later on.

ddprof-lib/src/main/cpp/flightRecorder.cpp
```diff
@@ -2332,11 +2335,161 @@ void Recording::recordHeapLiveObject(Buffer *buf, int tid, u64 call_trace_id,
...
+    buf->putVar32(event->_hops[i].klass_id);
+  }
+  // Edges array, LAST so its metadata position (after "chain", jfrMetadata.cpp)
+  // matches the write order. The labels pair with the chain's element order
+  // (see edge_count above); a count of 0 means label collection never ran
+  // for this event.
+  buf->putVar32(edge_count);
+  for (u32 i = 0; i < edge_count; i++) {
+    buf->putUtf8(edge_labels[i],
+                 (u32)strnlen(edge_labels[i], MAX_REFERENCE_CHAIN_EDGE_LABEL));
```
kaahos:
```suggestion
    const std::string &label = event->_hops[i].edge_label;
    buf->putUtf8(label.c_str(),
                 (u32)strnlen(label.c_str(), MAX_REFERENCE_CHAIN_EDGE_LABEL));
```

kaahos (approved):
Nice work, thanks! I've suggested 1 potential optimization and left 1 minor comment but nothing blocking for me.

jbachorik:
> Nice work, thanks! I've suggested 1 potential optimization and left 1 minor comment but nothing blocking for me.

@kaahos Thanks! I applied the suggestions - they were all valuable!

rkennke:
> Thanks! I applied the suggestions - they were all valuable!

Are you sure you pushed those fixes? I don't see them.

rkennke (approved):
Looks good. You might have forgotten to push some fixes (see other comment).

jbachorik:
@rkennke Yep. The signature window timed out and the push was just abandoned. Repushed. Thanks for pointing that out!

PR #805, opened 2026-09-18
ci: restore EL7 stresstest coverage, add nightly EL7 functional tests

rkennke:
## Summary

Follow-up to #790's "One coverage regression this introduces, and how to undo it" section.

Moving `build:x64` to AlmaLinux 8 (#790) took `stresstest:x64` with it, since it ran in the same image — runtime coverage at the shipped binary's actual glibc floor (2.17) went from thin to zero. This implements the two follow-ups #790 proposed:

1. **`.stresstest_job` gets its own runtime image**, independent of the image the artifact was built in. `RUNTIME_IMAGE` replaces the reused `BUILD_IMAGE` variable in the job template; `stresstest:x64` now points it at a new Oracle Linux 7 image (glibc 2.17), while `build:x64` stays on AlmaLinux 8 — matching #790's own note that this "is *better* than the old arrangement, because it would exercise the el8-built artifact on 2.17 rather than a 2.17-built one." Every other `stresstest:*` job keeps its existing image, just renamed to the new variable.
2. **A functional job matrix**, `functional:x64-el7-jdk{8,17,21}`, runs the `ddprof-test` debug suite against the shipped x64 artifact on the same EL7 image — the "libc-sensitive suites" coverage #790 called out. `jdk17` and `jdk21` are nightly-only (`CI_PIPELINE_SOURCE == "schedule"`, same pattern as the existing `chaos:build` job), so the full 3-JDK sweep doesn't multiply the per-PR matrix. `jdk8` additionally runs on every pipeline — it's the oldest/most different from the JDK 21 that runs Gradle, so it's the one job worth the added per-PR cost, and it keeps EL7 functional coverage from being gated on the nightly schedule entirely.

## Details

- `.gitlab/base/el7/Dockerfile` (new): Oracle Linux 7, runtime-only — no compiler. Both jobs consume the AlmaLinux-8-built `.so` via `-Pskip-native -Pwith-libs`, so nothing here needs to compile anything. Built and smoke-tested locally with Docker (JDK install via SDKMAN, `java -version` on glibc 2.17), then **built and pushed for real** in pipeline 138399627 (job [2056785023](https://gitlab.ddbuild.io/DataDog/java-profiler/-/jobs/2056785023)): `registry.ddbuild.io/ci/async-profiler-build:v138399627-x64-el7-base@sha256:d324b9b6c3fc48c552185e8969d88c836fb1e17a0e5dc483ba04bbd4f4ed2f38`, now pinned as `BUILD_IMAGE_X64_EL7`.
- `.gitlab/scripts/functional-tests.sh` (new): fetches the JDK under test from Adoptium into `JAVA_TEST_HOME`, so Gradle itself keeps running on JDK 21 (Gradle 9's minimum) regardless of which JDK the suite targets. Verified the Adoptium downloads and resulting `java -version` for 8/11/17/21 from inside the EL7 image.
- `.gitlab/build-deploy/images.yml`, `rebuild-images.sh`, `check-image-updates.sh`: wired `BASE_IMAGE_EL7` / `x64-el7` through the same image-build, repin and staleness-check machinery as the existing images.

## Not done here

That same `rebuild-images` run also failed to rebuild `x64-glibc` and `arm64` (both pre-existing images, unrelated to this change) — SDKMAN's CDN flaked mid-download on both (`Stop! The archive was corrupt and has been removed!`). Their pins are untouched here; a retry of `rebuild-images` should fix those independently.

## Test plan

- [x] `.gitlab/base/el7/Dockerfile` builds locally (`docker build`) and the resulting image runs `java -version` against glibc 2.17
- [x] Adoptium JDK fetch (8/11/17/21) verified to return 200 and produce a runnable `java` inside the EL7 image
- [x] `.gitlab/build-deploy/.gitlab-ci.yml` and `images.yml` YAML-validated
- [x] `bash -n` + `shellcheck` on `functional-tests.sh`
- [x] `x64-el7` image built and pushed via a real pipeline, digest pinned
- [ ] `stresstest:x64` and `functional:x64-el7-jdk*` green in a real GitLab pipeline (both nightly and per-PR paths)


rkennke:
Checked all 6 review threads — all point at the very first commit (daedd84e8), before the follow-up fixes landed. Each is already addressed:

- **Placeholder EL7 digest** (2 threads): pinned to the real built-and-pushed image, then re-pinned again after the `gcc-c++` fix.
- **`functional-tests.sh` not executable** (2 threads): `chmod +x`'d.
- **`-Pskip-native` also breaking `ddproftest`** (2 threads): `PlatformUtils.isNativeSkipped()` now supports a comma-separated project list, so `functional-tests.sh` skips only `ddprof-lib`'s own build while still compiling the JNI test helper for real.

Resolving all six.

.gitlab/build-deploy/.gitlab-ci.yml
```diff
@@ -164,9 +172,62 @@ stresstest:arm64-musl:
...
+# Runs the functional (ddprof-test) suite on Oracle Linux 7 against the
+# shipped x64 artifact, so the libc-sensitive suites get real coverage at the
+# glibc 2.17 floor (see PR #790's follow-up). Schedule-only by default: a
+# 2-3 JDK subset is enough to catch a libc regression, and Temurin 8/17/21
+# all start on glibc 2.17. jdk8 additionally runs on every PR (below), so
+# EL7 functional coverage isn't gated on the nightly schedule entirely.
+.functional_job:
+  extends:
+    - .retry-config
+    - .cache-config-pull
```
kaahos: this doesn't cache `.jdk-cache/` so it would be pulled everytime.

build-logic/conventions/src/main/kotlin/com/datadoghq/native/SimpleNativeLibPlugin.kt
```diff
@@ -56,7 +56,7 @@ class SimpleNativeLibPlugin : Plugin<Project> {
 
             // Compile task
             val compileTask = project.tasks.register("compileLib", NativeCompileTask::class.java) {
-                onlyIf { extension.enabled.get() && !project.hasProperty("skip-native") }
+                onlyIf { extension.enabled.get() && !PlatformUtils.isNativeSkipped(project) }
```
kaahos, line 59: shouldn't this be used widely in all the `*.kt` files?

kaahos (approved):
I've left 2 comments, but nothing blocking. Thanks!

.gitlab/build-deploy/.gitlab-ci.yml
zhengyu123: .gitlab/scripts/functional-tests.sh:42 — the JDK download cache directory (.jdk-cache/${TEST_JDK}) isn't listed in .gitlab/common.yml's cache paths, so it's never persisted across CI runs. Every functional-test job (including the one that runs on every PR) re-downloads the JDK tarball from Adoptium each time — the caching code is a no-op.

rkennke: Good catch — .jdk-cache/ was never in any cache paths (`.cache-config`/`.cache-config-pull` only cover .gradle/caches/, .gradle/wrapper/, .m2/repository/), so every run re-downloaded the JDK tarball from Adoptium. Gave `.functional_job` its own two-entry cache instead of extending `.cache-config-pull`: the JDK tarball's natural key is the major version (`TEST_JDK`), not the branch/commit the Gradle cache is keyed on, so it gets a static, ref-independent `pull-push` entry while the Gradle entry stays `pull`-only exactly as before.

build-logic/conventions/src/main/kotlin/com/datadoghq/native/SimpleNativeLibPlugin.kt
rkennke: Yes — fixed. GtestPlugin.kt, GtestTaskBuilder.kt, FuzzTargetsPlugin.kt and ddprof-lib/benchmarks/build.gradle.kts still used the bare hasProperty("skip-native") check.

GtestPlugin.kt/GtestTaskBuilder.kt are a direct swap (applied directly to ddprof-lib, so project.name is already "ddprof-lib" there). FuzzTargetsPlugin.kt and benchmarks/build.gradle.kts needed more care: they're applied to the :ddprof-lib:fuzz and :ddprof-lib:benchmarks *subprojects* (project.name == "fuzz"/"benchmarks"), so a naive swap to isNativeSkipped(project) would have been a real regression — -Pskip-native=ddprof-lib,malloc-shim would no longer have skipped them, since neither "fuzz" nor "benchmarks" is in that list. Fuzzing/benchmarking are part of ddprof-lib's native surface, not something a caller names on its own, so those go through project.parent's identity instead.

Verified locally: -Pskip-native=ddprof-lib,malloc-shim still skips :ddprof-lib:benchmarks:compileBenchmark (SKIPPED, not executed), and it still compiles normally without the flag.

zhengyu123 (approved):
LGTM

